PR #1, opened 2022-09-28
Add a collector Trait and a way to register and manipulate them

atenart:
This adds a collector Trait, which defines the collector interface; and an implementation of a collection, allowing to register and manipulate collector in batches.

More info in the commit logs and especially in the code itself. The current Trait is basic and is expected to evolve and should be a good starting point to work on specific topics (logging, cmdline argument parsing, event retrieving, etc).

The way collectors are registered and manipulated is by using a CollectorCollection type and registering collectors to it. This type will probably be a singleton, initialized once in the program lifetime. Another solution would have been to use a static variable as a singleton and to provide macros to register the collectors at the program startup, but after some investigations this did not seemed a good solution. Part of the reason is a combination of lazy_static and startup should have been used and the latter has been yanked. This also looked too complex for no real benefit (I initially thought otherwise).

### DIFF
--- a/src/collector/mod.rs
+++ b/src/collector/mod.rs
@@ -19,6 +19,9 @@ use std::collections::HashMap;
 use anyhow::{bail, Result};
 use log::{error, warn};
 
+mod skb;
+use skb::SkbCollector;
+
 /// Generic trait representing a collector. All collectors are required to
 /// implement this, as they'll be manipulated through this trait.
 trait Collector {
@@ -116,6 +119,8 @@ pub(crate) fn get_collectors() -> Result<CollectorsCollection> {
     let mut collection = CollectorsCollection::new();
 
     // Register all collectors here.
+    collection
+        .register(Box::new(SkbCollector::new()?))?;
 
     Ok(collection)
 }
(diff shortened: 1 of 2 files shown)


src/collector/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// Initialize all collectors by calling their `init()` function. Collectors
+    /// failing to initialize will be removed from the collection.
+    pub(crate) fn init(&mut self) -> Result<()> {
+        let mut to_remove = Vec::new();
+
+        // Try initializing all collectors in the collection. Failing ones are
+        // put on a list for future removal.
+        for (_, c) in self.list.iter_mut() {
+            if let Err(_) = c.init() {
+                to_remove.push(c.name());
```
amorenoz, line 88: Maybe move the above `error!` up here to log the initialization error alongside what c.init() returns?

src/collector/mod.rs
```diff
@@ -19,6 +19,9 @@ use std::collections::HashMap;
 use anyhow::{bail, Result};
 use log::{error, warn};
 
+mod skb;
+use skb::SkbCollector;
```
amorenoz, line 23: Shouldn't modules be organized the other way around?
i.e: modules create Collectors that `use` the `collector`module

PR #3, opened 2022-10-03
core: add a kernel_symbol module

atenart:
Simple interface to get kernel symbols addresses from their name and the opposite. This will be useful for probing the kernel, from configuring kprobes to understanding where an event is coming from.

### DIFF
--- a/src/core/kernel_symbols.rs
+++ b/src/core/kernel_symbols.rs
@@ -1,0 +1,143 @@
+//! # Kernel symbols
+//!
+//! Interface to query the kernel symbol addresses / name relationship, both
+//! ways. It does so by parsing /proc/kallsyms and by using a singleton
+//! initialized on-demand.
+
+// Internal library, some helpers might not be used, that's fine.
+#![allow(dead_code)]
+
+use std::fs;
+
+use anyhow::{anyhow, bail, Result};
+use bimap::BiHashMap;
+use once_cell::sync::OnceCell;
+
+/// Path to the kernel symbols file. It should be under /proc/kallsyms (and this
+/// is the default, see below). To allow unit tests to run, as /proc/kallsyms
+/// might vary and requires privileged access, a OnceCell is used to allow unit
+/// tests to override it.
+static KALLSYMS: OnceCell<String> = OnceCell::new();
+
+/// Kernel symbols bidirectional map. To retrieve it, please use:
+/// ```
+/// let symbols = get_symbols!().unwrap();
+/// ```
+static SYMBOLS: OnceCell<BiHashMap<u64, String>> = OnceCell::new();
+
+/// Return a reference to the symbol map and initialize it on first access. To
+/// set the initial values, KALLSYMS is parsed as it contains the kernel symbol
+/// addr<>name relationships.
+macro_rules! get_symbols {
+    () => {
+        SYMBOLS.get_or_try_init(|| {
+            let kallsyms_file = KALLSYMS.get_or_init(|| String::from("/proc/kallsyms"));
+            let file = fs::read_to_string(kallsyms_file)?;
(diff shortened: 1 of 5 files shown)


src/core/kernel_symbols.rs
```diff
@@ -0,0 +1,132 @@
...
+        // be considered near it (and part of it).
+        if target < *addr {
+            continue
+        }
+
+        let score = target.abs_diff(*addr);
+        if score < best_score {
+            nearest = *addr;
+            best_score = score;
+        }
```
amorenoz, line 82: nit: Maybe if score == 0 {break;}?

amorenoz (commented):
Besides a small nit, the code looks good to me.
Tested on my system and works well.

amorenoz:
I was going to add it to #4 but I run rustfmt here first:
```
rustfmt --check src/main.rs
Diff in /home/amorenoz/code/packet-tracer/src/core/kernel_symbols.rs at line 7:
 // Internal library, some helpers might not be used, that's fine.
 #![allow(dead_code)]
 
-use std::fs;
 use anyhow::{anyhow, bail, Result};
 use bimap::BiHashMap;
 use once_cell::sync::OnceCell;
Diff in /home/amorenoz/code/packet-tracer/src/core/kernel_symbols.rs at line 14:
+use std::fs;
 
 /// Path to the kernel symbols file. It should be under /proc/kallsyms (and this
 /// is the default, see below). To allow unit tests to run, as /proc/kallsyms
Diff in /home/amorenoz/code/packet-tracer/src/core/kernel_symbols.rs at line 30:
 macro_rules! get_symbols {
     () => {
         SYMBOLS.get_or_try_init(|| {
-            let kallsyms_file = KALLSYMS.get_or_init(|| { String::from("/proc/kallsyms") });
+            let kallsyms_file = KALLSYMS.get_or_init(|| String::from("/proc/kallsyms"));
             let file = fs::read_to_string(kallsyms_file)?;
             let mut map = BiHashMap::new();
 
Diff in /home/amorenoz/code/packet-tracer/src/core/kernel_symbols.rs at line 40:
                     bail!("Invalid kallsyms line: {}", line);
                 }
 
-                let symbol: &str = data[2].split('\t').next()
+                let symbol: &str = data[2]
+                    .split('\t')
+                    .next()
                     .ok_or(anyhow!("Couldn't get symbol name for {}", data[0]))?;
 
                 map.insert(u64::from_str_radix(data[0], 16)?, String::from(symbol));
Diff in /home/amorenoz/code/packet-tracer/src/core/kernel_symbols.rs at line 53:
 
 /// Return a symbol name given its address, if a relationship is found.
 pub(crate) fn get_symbol_name(addr: u64) -> Result<String> {
-    Ok(get_symbols!()?.get_by_left(&addr)
+    Ok(get_symbols!()?
+        .get_by_left(&addr)
         .ok_or(anyhow!("Can't get symbol name for {}", addr))?
         .clone())
 }
Diff in /home/amorenoz/code/packet-tracer/src/core/kernel_symbols.rs at line 60:
 
 /// Return a symbol address given its name, if a relationship is found.
 pub(crate) fn get_symbol_addr(name: &str) -> Result<u64> {
-    Ok(*get_symbols!()?.get_by_right(name)
+    Ok(*get_symbols!()?
+        .get_by_right(name)
         .ok_or(anyhow!("Can't get symbol address for {}", name))?)
 }
 
Diff in /home/amorenoz/code/packet-tracer/src/core/kernel_symbols.rs at line 72:
         // The target address has to be greater or equal to a symbol address to
         // be considered near it (and part of it).
         if target < *addr {
-            continue
+            continue;
         }
 
         let score = target.abs_diff(*addr);

```

amorenoz (approved):

PR #20, opened 2022-10-19
Initial collector & kernel probes support

atenart:
First PR adding a collector trait (& a minimal dummy collector), alongside with a kernel module providing an API to add probes (kprobes & raw tracepoints). So far the probes aren't doing anything, this will come in a follow-up PR.

More details in the commit logs and in the code comments.

Note: to please Rust check & clippy some modules define an 'allow(dead_code)' directive. This will be removed in the future wherever possible and this should be easy to track as each of these directives comes with a 'FIXME' comment.


src/core/probe/kernel/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#![allow(dead_code)] // FIXME
+
+use anyhow::Result;
+
+/// Main object representing the kernel probes and providing an API for
+/// consumers to register probes, hooks, maps, etc.
+pub(crate) struct Kernel {}
+
+impl Kernel {
```
amorenoz, line 51: General question. What should be our policy with regards to `mod.rs`?

I see that [recent changes to rust-lang have made the use of `mod.rs` not needed](https://doc.rust-lang.org/beta/edition-guide/rust-2018/path-changes.html#no-more-modrs).

Particularly, I find it annoying to have actual code (not just relatively specific things like re-exports and includes) in a file that's named the same as many others. I end up having a bunch of `mod.rs` files open in my editor and it's confusing. It feels to me like putting important code into `__init__.py`. Maybe it's just me and I just have to get used to it, although it does seem like[ one (small) reason behind the 2018 shift](https://github.com/rust-lang/rfcs/blob/master/text/2126-path-clarity.md#the-modrs-file), so I'm not alone, phew :-)

What do you guys think?
/cc @vlrpl

Some additional reading:
https://blog.devgenius.io/how-to-mod-in-rust-b2da3e1ff0f2

A nice image:
https://i.redd.it/1yy98srxyvx81.png

amorenoz: So, are you planning on updating the PR or should we switch to mod.rs re-exporting after?

src/core/probe/kernel/mod.rs
```diff
@@ -109,6 +134,18 @@ trait ProbeBuilder {
     fn attach(&mut self, target: &str) -> Result<()>;
 }
 
+fn reuse_map_fds(open_obj: &libbpf_rs::OpenObject, map_fds: &Vec<(String, i32)>) -> Result<()> {
+    if !map_fds.is_empty() {
```
vlrpl, line 138: I noticed in more than one place some logic including is_empty() before iterating.
Is the purpose of this to avoid some hidden costs, e.g. into_iter() or something?

src/core/probe/kernel/mod.rs
```diff
@@ -109,6 +134,18 @@ trait ProbeBuilder {
...
+        for map in map_fds {
```
vlrpl, line 139: another nit based on locality :D 
What about using explicit .iter() here?
I may be wrong, but I find implicit conversion a bit error prone as it may vary depending on the fact the collection is a [mut] ref or not

amorenoz (approved):
/lgtm

PR #22, opened 2022-10-19
Initial hooks support

atenart:
Initial support to build, register and load hooks into kernel probes. The API allows to register hooks running on all probes and hooks running on targeted ones. For now if a hook is loaded in a probe it will run unconditionally; this will change later by checking probe capabilities at runtime.

This implementation attach all generic hooks to all probes, including ones added for a specific hook whether or not if the probe was registered through `add_probe`. We could change this later and only attach generic hooks to probes explicitly added. WDYT?

More details in the commit logs and in the code itself. An example of this being used can be found in branch at/next.

This is based on #20


src/core/probe/kernel/bpf/include/common.h
```diff
@@ -0,0 +1,45 @@
...
+ *            helper.
+ */
+struct trace_context {
+	u64 timestamp;
+	u64 ksym;
+	struct regs regs;
+};
+
+/* Helper to retrieve a function parameter argument using the common context */
+#define get_param(ctx, offset, type)	\
```
amorenoz, line 42: Wondering if we should prefix helper names. The project name is not written in stone but we can start with "trace_"? or "trace_context", "trace_ctx"

src/core/probe/kernel/mod.rs
```diff
@@ -137,30 +168,111 @@ impl Kernel {
...
+
+        self.targeted_probes.push(set);
+        Ok(())
+    }
+
     /// Attach all probes.
     pub(crate) fn attach(&mut self) -> Result<()> {
+        // Take care of target probes first, they are probably higher prio.
+        for set in self.targeted_probes.iter_mut() {
+            let hooks = [set.hooks.clone(), self.hooks.clone()].concat();
```
amorenoz, line 247: I don't know if it will be the case but I'm thinking that it's more likely that targeted probes make use of information that come from generic probes than the opposite. If so, maybe we should run generic probes first. WDYT?

amorenoz: Sorry I didn't respond to this:

> This raises lots of questions:
> 
>     * How do we share info between hooks? Event struct? Custom one? Array of pointers to stack variables w/ an unique id?
> 

I was thinking maybe if a hook may depend on another hook populating some info on the event struct itself.

>     * Is the dependency generic enough to be targeted probes -> generic ones? Or do we need a 1:1 mapping?
> 
>     * If a hook we depend on fail to process, what do to? Do we (and how) propagate the error? The proposal here discard any hook error to let others try gathering info.
> 

No idea. I haven't thought it through too much.

> 
> Do you see not supporting this as a blocking point for your use cases? Is this something we'll have to support, or requesting hooks to be independent would still work?
> 

No, I don't think so.

> IMHO until the above is defined we can ignore the hook calling order. Such information sharing / hook dependency support can come in its own proposal and rework the above part.
> 
> Nit: turns out the comment in the code snippet above is wrong as the hook calling order do not really matter for now...

amorenoz (approved):

PR #24, opened 2022-10-26
Extend probe runtime inspection and provide parameter offsets

atenart:
The two first commits are small fixes.

The last one extends the runtime inspection of symbols (functions/tracepoints) to dynamically check if given parameters are available and if so to provide their offsets to the BPF probes.

The extra information is then used in BPF to add helpers to easily retrieve known parameters, e.g.

```
struct sk_buff *skb = trace_get_sk_buff(ctx);
if (!skb)
        return 0;
[...]
```

Current known parameters: `struct sk_buff *`, `enum skb_drop_reason`, `struct net_device *`, `struct net *`.

An helper is also added for collectors to inspect function & tracepoints at runtime, e.g.

```
let skb = kernel.function_parameter_offset(ProbeType::Kprobe, "skb_free_reason", "struct sk_buff *")?.is_some();
```

Based on #22 .

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,7 +38,7 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 [[package]]
 name = "btf-rs"
 version = "0.1.0"
-source = "git+https://github.com/net-trace/btf-rs#a8ffd76a4b21c1aaa735af380241bb0300fc8cfb"
+source = "git+https://github.com/net-trace/btf-rs#15aea1f65fed82e6a50e0ff68be98cbf8c9fde20"
 dependencies = [
  "anyhow",
  "byteorder",
@@ -331,9 +331,11 @@ dependencies = [
  "btf-rs",
  "libbpf-cargo",
  "libbpf-rs",
+ "libbpf-sys",
  "log",
  "memmap2",
  "once_cell",
+ "plain",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


src/collector/mod.rs
```diff
@@ -194,6 +194,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg_attr(not(feature = "test_cap_bpf"), ignore)]
```
amorenoz, line 197: Can we move all ebpf handling to init() phase?
Why I'm asking this:
In my current working version of cli management (#46) I need to register cli arguments _before_ actually parsing them, hence, before knowing what command we'll be running. So, the attempt to load any ebpf should be postponed to some later stage (init() or start()) so it does not interfere with other commands that might not require ebpf at all (e.g. postprocessing). Does it make sense?

amorenoz: > > Why I'm asking this: In my current working version of cli management (#46) I need to register cli arguments _before_ actually parsing them, hence, before knowing what command we'll be running. So, the attempt to load any ebpf should be postponed to some later stage (init() or start()) so it does not interfere with other commands that might not require ebpf at all (e.g. postprocessing). Does it make sense?
> 
> It would be great to check the subcommand first w/o having to register all arguments; this way we can avoid to register/initialize collectors when doing post-processing, and the opposite as well. Maybe not using clap for this would work, but idk if that would be easy to handle.

It's definitely not going to be very clean but I can try to implement it.
Or maybe after PoC?

PR #25, opened 2022-10-26
Rename modules mod.rs -> <module>.rs + re-export in mod.rs

atenart:
As discussed, renaming modules from:

```
module/
  mod.rs
```

to:

```
module/
  mod.rs   // only re-exporting module.rs
  module.rs
```

This comes with two small differences with the previous code:
- Objects, interfaces, etc shared in the module have to use `pub(super)` otherwise they won't be shared with the sub-modules.
- Sub-modules have to be imported using `super::` in the now main module file.

IMHO this is OK and even provide more separation of code between a module and its sub-modules.

Based on #24


vlrpl (approved):
lgtm

PR #50, opened 2022-11-15
Allow hooks to reuse their own maps

atenart:
With this collectors can register maps and let their hooks reuse the maps fds. This will help in having per-collector/hook maps for retrieving specific information or retrieving configuration in the BPF part.


amorenoz (commented):
In `core: probe: kernel: add an helper to retrieve the ksym of probes`
s/from where there are being called/from where they are being called/

vlrpl (approved):
lgtm, thanks!

PR #51, opened 2022-11-15
Track skbs using a dedicated collector

atenart:
This adds a collector which will track skbs and generate a unique id so we can recognize a collector later in the stack and reconstruct its life at post-processing time (or manually looking at the events; once available).

Some extra commits to make the tracking collector easier to implement. More info in the commit themselves.

Based on #50.

### DIFF
--- a/src/collector/collector.rs
+++ b/src/collector/collector.rs
@@ -89,7 +89,7 @@ impl Group {
             let c = self
                 .list
                 .get_mut(name)
-                .ok_or_else(|| anyhow!(format!("unknown collector: {}", &name)))?;
+                .ok_or_else(|| anyhow!("unknown collector: {}", &name))?;
             if let Err(e) = c.init(cli, &mut self.kernel) {
                 to_remove.push(c.name());
                 error!(
--- a/src/collector/collector.rs
+++ b/src/collector/collector.rs
@@ -5,6 +5,7 @@ use log::{error, warn};
 
 use super::ovs::OvsCollector;
 use super::skb::SkbCollector;
+use super::skb_tracking::SkbTrackingCollector;
 use crate::cli::{cmd::collect::Collect, dynamic::DynamicCommand, CliConfig};
 use crate::core::probe;
 
@@ -141,6 +142,7 @@ pub(crate) fn get_collectors() -> Result<Group> {
 
     // Register all collectors here.
     group
+        .register(Box::new(SkbTrackingCollector::new()?))?
         .register(Box::new(SkbCollector::new()?))?
         .register(Box::new(OvsCollector::new()?))?;
 
--- a/src/collector/skb_tracking/bpf/tracking_hook.bpf.c
+++ b/src/collector/skb_tracking/bpf/tracking_hook.bpf.c
@@ -1,0 +1,139 @@
+#include <vmlinux.h>
+#include <bpf/bpf_core_read.h>
+
+#include <common.h>
+
+/* Tracking configuration to provide hints about what the probed function does
+ * for some special handling scenarios.
+ *
+ * Indexed in the tracking_config_map by the function ksym address.
+ *
+ * Please keep in sync with its Rust counterpart in collector::skb_tracking.
+ */
+struct tracking_config {
+	/* Function is freeing skbs */
+	u8 free;
+	/* Function is invalidating the head of skbs */
+	u8 inv_head;
+};
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, PROBE_MAX);
+	__type(key, u64);
+	__type(value, struct tracking_config);
+} tracking_config_map SEC(".maps");
+
+/* The tracking_info structure stores information on known skbs. It is indexed
+ * in the tracking_map by the skb data address (and in some temporary cases by
+ * the skb address directly).
+ *
+ * In order to uniquely identify skbs, the tuple (addr, timestamp) is used and
+ * must be reported as part of all events (TODO).
+ *
+ * Please keep in sync with its Rust counterpart in collector::skb_tracking.
+ */
+struct tracking_info {
+	/* When the skb was first seen */
+	u64 timestamp;
+	/* When the skb was last seen */
+	u64 last_seen;
+	/* Original head address; useful when the head is invalidated */
+	u64 orig_head;
+} __attribute__((packed));
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 8192);
+	__type(key, u64);
+	__type(value, struct tracking_info);
+} tracking_map SEC(".maps");
+
+/* Must be called with a valid skb pointer */
+static __always_inline int track_skb(struct trace_context *ctx,
+				     struct sk_buff *skb)
+{
+	enum skb_drop_reason drop_reason = 0;
+	struct tracking_info *ti = NULL, new;
+	bool free = false, inv_head = false;
+	struct tracking_config *cfg;
+	u64 head, ksym = ctx->ksym;
+
+	/* Try to retrieve the tracking configuration for this symbol. Only
+	 * specific ones will be found while we want to track skb in all
+	 * functions taking an skb as a parameter. When no tracking
+	 * configuration is found, the function being probed is just quite
+	 * generic.
+	 */
+	cfg = bpf_map_lookup_elem(&tracking_config_map, &ksym);
+	if (cfg) {
+		free = cfg->free;
+		inv_head = cfg->inv_head;
+	}
+
+	head = (u64)BPF_CORE_READ(skb, head);
+	if (!head)
+		return 0;
+
+	ti = bpf_map_lookup_elem(&tracking_map, &head);
+
+	/* No tracking info was found for this skb. */
+	if (!ti) {
+		/* It might be temporarily stored it using its skb address. */
+		ti = bpf_map_lookup_elem(&tracking_map, (u64 *)&skb);
+		if (ti) {
+			/* If found, index it by its data address from now on,
+			 * as others.
+			 */
+			bpf_map_delete_elem(&tracking_map, (u64 *)&skb);
+			bpf_map_update_elem(&tracking_map, &head, ti,
+					    BPF_NOEXIST);
+		}
+	}
+
+	/* Still NULL, this is the first time we see this skb. Create a new
+	 * tracking info.
+	 */
+	if (!ti) {
+		ti = &new;
+		ti->timestamp = ctx->timestamp;
+		ti->last_seen = ctx->timestamp;
+		ti->orig_head = head;
+
+		/* No need to globally track it if the first time we see this
+		 * skb is when it is freed.
+		 */
+		if (!free)
+			bpf_map_update_elem(&tracking_map, &head, &new,
+					    BPF_NOEXIST);
+	}
+
+	/* Track when we last saw this skb, as it'll be useful to garbage
+	 * collect tracking map entries if we miss some events.
+	 */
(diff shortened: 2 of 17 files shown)


src/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -0,0 +1,144 @@
...
+	/* Still NULL, this is the first time we see this skb. Create a new
+	 * tracking info.
+	 */
+	if (!ti) {
+		struct tracking_info new = {
+			.timestamp = ctx->timestamp,
+			.last_seen = ctx->timestamp,
+			.orig_head = head,
+		};
+		ti = &new;
```
amorenoz, line 101: I have no idea how stack unwinding works on ebpf  but this would typically raise a stack-use-after-scope error on clang ASAN. Are we ok doing this?

src/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -0,0 +1,144 @@
...
+					    BPF_NOEXIST);
+	}
+
+	/* Track when we last saw this skb, as it'll be useful to garbage
+	 * collect tracking map entries if we miss some events.
+	 */
+	ti->last_seen = ctx->timestamp;
+
+	/* If the function invalidates the skb head, we can't know what will be
+	 * the new head value. Temparorily track the skb using its skb address.
```
amorenoz, line 117: Temporarily

src/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -0,0 +1,144 @@
...
+		bpf_map_update_elem(&tracking_map, (u64 *)&skb, ti, BPF_NOEXIST);
+	/* If the skb is freed, remove it from our tracking list. */
+	else if (free)
+		bpf_map_delete_elem(&tracking_map, &head);
+
+	if (trace_arg_valid(ctx, skb_drop_reason))
+		drop_reason = trace_get_skb_drop_reason(ctx);
+
+	// TODO: do something with the tracking id
+	// (ti->orig_head << 64 | ti->timestamp) and the drop reason.
```
amorenoz, line 129: Are you planning to do this in this PR?

src/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -0,0 +1,144 @@
...
+	/* Track when we last saw this skb, as it'll be useful to garbage
+	 * collect tracking map entries if we miss some events.
+	 */
+	ti->last_seen = ctx->timestamp;
+
+	/* If the function invalidates the skb head, we can't know what will be
+	 * the new head value. Temparorily track the skb using its skb address.
+	 */
+	if (inv_head)
+		bpf_map_update_elem(&tracking_map, (u64 *)&skb, ti, BPF_NOEXIST);
```
amorenoz, line 120: If the head is no longer a valid key, and we're using the skb as temporary key to find the uid, shouldn't we remove the old (no-longer-valid) head from the map?

src/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -0,0 +1,144 @@
...
+			.orig_head = head,
+		};
+		ti = &new;
+
+		/* No need to globally track it if the first time we see this
+		 * skb is when it is freed.
+		 */
+		if (!free)
+			bpf_map_update_elem(&tracking_map, &head, &new,
+					    BPF_NOEXIST);
```
amorenoz, line 108: And if (inv_head), we should use skb instead of head right? I know we're doing it later on anyway but why adding it twice?

src/collector/skb_tracking/skb_tracking.rs
```diff
@@ -0,0 +1,195 @@
...
+    garbage_collector: Option<thread::JoinHandle<()>>,
+}
+
+impl Collector for SkbTrackingCollector {
+    fn new() -> Result<SkbTrackingCollector> {
+        Ok(SkbTrackingCollector::default())
+    }
+
+    fn name(&self) -> &'static str {
+        "skb-tracking"
```
amorenoz, line 32: We'll probably need to make it a const if we're registering arguments.

src/collector/skb_tracking/skb_tracking.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    fn init(&mut self, kernel: &mut kernel::Kernel) -> Result<()> {
+        self.init_tracking(kernel)?;
+
+        // We'd like to track free reasons as well. As this is not available on
+        // older kernels, it is allowed to fail.
+        let _ = kernel.add_probe(ProbeType::Kprobe, "kfree_skb_reason");
```
amorenoz, line 40: Maybe warn or log something if it fails?

src/collector/skb_tracking/skb_tracking.rs
```diff
@@ -0,0 +1,195 @@
...
+        // Take care of gargabe collection of tracking info. This should be done
+        // in the BPF part for most if not all skbs but we might lose some
+        // information (and tracked functions might fail resulting in incorrect
+        // information).
+        self.garbage_collector = Some(thread::spawn(move || {
+            let tracking_map = tracking_map.get_mut();
+
+            loop {
+                // Let's run every 5 seconds.
+                thread::sleep(Duration::from_secs(5));
```
amorenoz, line 138: Make it a const at the top of the file?

src/core/workaround.rs
```diff
@@ -0,0 +1,103 @@
+/// # Workaround
```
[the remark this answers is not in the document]
amorenoz: Can we avoid exporting  `SendWrapper` to avoid any implementation outside of this module?

src/core/workaround.rs
```diff
@@ -0,0 +1,103 @@
...
+/// For now we support SendWrapperMap.
+
+/// SendWrapper<T> should not be used directly but only specific typedefs ones.
+/// When adding a new SendWrapper type research should be done to ensure this is
+/// actually safe!
+///
+/// SendWrapper<T> only implements the Send trait, if Sync is needed and more
+/// than one user wants access use the common Arc<Mutex<SendWrapper<T>>>
+/// construction (see core::workaround::tests::sync).
+pub(crate) struct SendWrapper<T>(T)
```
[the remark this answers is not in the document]
amorenoz: How about a sealed trait? I saw this somewhere and found it an interesting solution, basically the idea is to create an internal non-public module:

```rust
mod sealed {
    pub(super) trait Hacked {}
    /* Do not add anything else here unless it's absolutely necessary!!!! */
    impl Hacked for libbpf-rs::Map {}
    impl Hacked for libbpf-rs::RingBuf {}
}

/* Only a handful of hacked structs can implement this trait. And for someone to do it, they have to touch a very specific part of the code that raises the reviewer's alarms. */
pub(crate) struct SendWrapper<T: mod::Hacked>(T)
```
I don't know if this would work in this case but it might be worth a try.

src/core/probe/kernel/inspect.rs
```diff
@@ -16,16 +19,36 @@ pub(super) struct TargetDesc {
...
             btf: Btf::from_file("/sys/kernel/btf/vmlinux")?,
             #[cfg(test)]
             btf: Btf::from_file("test_data/vmlinux")?,
-        })
+            tracable_funcs: None,
+        };
+
+        // Not all functions we'll get from BTF/kallsyms are traceable. Use the
+        // following, when available, to narrow down our checks.
+        if let Ok(file) = fs::read_to_string("/sys/kernel/debug/tracing/available_filter_functions")
```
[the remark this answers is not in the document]
amorenoz: Agree. Either that or put `_kprobe` in the function name so it's clear it's a kprobe-only thing because it being part of generic `kernel` could lead to confusion.

src/collector/collector.rs
```diff
@@ -4,6 +4,7 @@ use anyhow::{bail, Result};
 use log::{error, warn};
 
 use super::skb::SkbCollector;
+use super::skb_tracking::SkbTrackingCollector;
 use crate::core::probe;
```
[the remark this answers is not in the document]
amorenoz: Interesting, even rustfmt's example puts `super` before `crate`: https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#StdExternalCrate%5C%3A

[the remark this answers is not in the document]
amorenoz: I guess we can keep it as it is then.

src/core/workaround.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Get a mutable reference to the inner object.
+    #[allow(dead_code)]
+    pub(crate) fn get_mut(&mut self) -> &mut T {
+        &mut self.0
+    }
+}
+
+unsafe impl<T> Send for SendWrapper<T> {}
+
+// Define below types where it is actually safe to send them to another thread.
```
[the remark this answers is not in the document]
amorenoz: Created #56  to remind ourselves to do this.

src/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -0,0 +1,144 @@
...
+	 */
+	cfg = bpf_map_lookup_elem(&tracking_config_map, &ksym);
+	if (cfg) {
+		free = cfg->free;
+		inv_head = cfg->inv_head;
+	}
+
+	head = (u64)BPF_CORE_READ(skb, head);
+	if (!head)
+		return 0;
```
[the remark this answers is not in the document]
amorenoz: Fair enough

src/collector/skb_tracking/bpf/tracking_hook.bpf.c
amorenoz: 2. could lead to a key leak right?

amorenoz: I mean, that's OK since we have the garbage collector but it'll be nice to have those cases documented.

src/core/probe/kernel/inspect.rs
```diff
@@ -0,0 +1,272 @@
+#![allow(dead_code)] // FIXME
+
+use anyhow::{bail, Result};
+use btf_rs::{Btf, Type};
+
+use super::{config::ProbeConfig, ProbeType};
+use crate::core::kernel_symbols;
+
+#[derive(Default)]
+pub(super) struct TargetDesc {
```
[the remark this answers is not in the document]
amorenoz: I think public structs should be documented with document comments (///). It really makes reading the doc much easier (as opposed to reading the names of the members and figuring it out)

src/collector/skb_tracking/bpf/tracking_hook.bpf.c
amorenoz: If the first event we get is an `inv_head` we add 2 keys: `old_head` and `skb`.
The next event will fail to find a key on `new_head` but it'll find it on `skb`, it'll remove `skb` add `new_head`. Then eventually it'll free `new_head`, but no one removed `old_head`. Maybe I'm missing something?

amorenoz: In my original comment what I meant is:
I understand that this mechanism is needed because we need to contemplate situation 2:  we don't know if this packet was a clone and there's actually an skb somewhere with `old_head` so we need to keep it in the map.

amorenoz: I was missing the fact that we're tracing head frees not skb ones.

amorenoz (approved):

PR #53, opened 2022-11-21
BPF<>Rust event reporting and events internal representation

atenart:
Following our discussion, I think we kind of agreed on at least the BPF <> Rust event logic (dynamic, variable length, tlv-ish). I implemented a first version to at least validate this would work, and it surprisingly does :)

Things we hadn't settled on and can change:

- I went for using per-collector structures to format the event sections instead of only using generic types. Changing this would be quite easy.
- Raw events are put in an `HashMap<String, String>` when unmarshaling.

Events are also currently not propagated after unmarshaling. Also One thing I'm not happy with for now is the naming in the event module.

Based on #51.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,6 +20,10 @@ fn main() -> Result<()> {
             let config = cli.run()?;
             collectors.init(&config)?;
             collectors.start(&config)?;
+            loop {
+                let event = collectors.poll_event()?;
+                println!("{}", event.to_json());
+            }
         }
         _ => {
             error!("not implemented");


amorenoz (commented):
Did a first quick round and apart from a couple of nits I think it's in a very good shape. The design is what we discussed and implementation seems to me very good.
Nice job!
PS: I'll give it another round with more coffee next week :-D

src/core/events/events.rs
```diff
@@ -0,0 +1,173 @@
...
+    };
+}
+
+/// Implementation of an event field type, used to hold the actual data and
+/// provide helpers to serialize/deserialize it.
+pub(crate) trait EventFieldType {
+    fn name(&self) -> &'static str;
+    fn as_any(&self) -> &dyn Any;
+    fn to_json(&self) -> serde_json::Value;
+    fn from_json(from: serde_json::Value) -> Result<Self>
```
amorenoz, line 139: I'm thinking about how the `from_json` side will work.
Although we have `from_json` in the `EventFieldType` trait, we will need to call it on the specific type. So we need a way to get the type from the owner and field_name.
A possible way would be to convert the `EventUnmarshaler` into a trait, something like:
```rust
trait EventUnmarshaler {
   type Field: EventFieldType;
   fn from_raw(&Vec[u8]) -> Result<Field>;
   fn from_json(serde_json::Value]) -> Result<Field>;
}
```
WDYT?

amorenoz: Sorry I think I wrote some nonsense which made it more confusing.
I didn't mean to tie everything at the `EventFieldType` level but at the section level.

> I haven't found yet a perfect representation for the json to event translation. The issue we have is there is no serde_json::Value to EventFieldType relationship (same with raw sections). 

There is: `EventFieldType::from_json(from: serde_json::Value) -> Result<Self>` . Am I missing something?


>  What we have is raw section 1:1 internal collector section representation 1:n section n:1 event. 

I don't understand what's the differnt between "section" and "internal collector section representation".


What I see is that: for each `BpfEventOwner` we have an `Unmarshaler` capable of generating a `Vec[EventField]` which is then converted into an internal representation of an event section such as:
```rust
HashMap::from([
    ("owner", HashMap::from([
        ("field0_name", field0_val),
        ("field1_name", field1_val)
    ]),
])
``` 
Which is trivially converted into json:
```
"owner": {
    "field0_name": field0_val,
    "field1_name": field1_val,
}
```
Mainly because fieldN_val implement `to_json()`

To keep things simple, I would just mimic this design on the json side and have:
```rust
trait EventUnmarshaler {
   fn from_raw_section(&BpfRawSection, &mut Vec<EventField>) -> Result<()>
   fn from_json_section(&json::Map, &mut Vec<EventField>) -> Result<()>
}
```
Following the example of `skb_tracing` we would have something like:

```rust
impl Collector for SkbTrackingCollector {  
    fn init(...events: &mut BpfEvents) {
        events.register_unmarshaller(SkbTrackingUnmarshaler{})
    }  
}

struct SkbTrackingUnmarshaler{}:

impl EventUnmarshaler  for SkbTrackingUnmarshaler {
fn from_raw_section(section: &BpfRawSection, fields: &mut Vec<EventField>) -> Result<()> {
    // some integrity checks
    // use an internal struct that happens to have the same binary layout as the RawSection (nice implementation detail but not important from an API PoV)
    let field0_val = [...]; // Implements EventFieldType
    let field1_val = [...]; // Implements EventFieldType
    fields.push(event_field!("field0_name", field0_val));
    fields.push(event_field!("field1_name", field1_val));
}

fn from_json_section(map: &json::Map, fields: &mut Vec<EventField>) -> Result<()> {
    // some integrity checks
    // Maybe perform some manual complex parsing.
    let field0_val = manually_parse_field0(map.get("field0_name"));  // Implements EventFieldType
    let field1_val = manually_parse_field1(map.get("field1_name"));  // Implements EventFieldType
    
    // Maybe create an intermediate object that uses serde::Deserialize to perform some semi-automatic deserialization. Maybe it's even the same object that we use for RawSection deserialization, who knows.
    // But ultimately...
    fields.push(event_field!("field0_name", field0_val));
    fields.push(event_field!("field1_name", field1_val));
} 
}
```

Each module defines how it's section is unmarshaled from `BpfRawSection -> Vec<EventField> ` and from `json::Map -> Vec<EventField>`

This would allow collectors to implement arbitrarily complex json_to_vector_of_EventFields unmarshalers.

One step further would be to implement serde:Deserialize for our binary TLV format so a single struct can, indeed, be used to deserialize both from binary and from json.

amorenoz: BTW, I'm not arguing for doing all this in this series but I do think we should at least agree on how we are going to do it so that we minimize the API changes in the future.

Essentially the problem is that currently the serialization is done at the `EventFieldType` level but deserialization cannot be done at that level unless we map names to types somehow. Having serialization and deserialization at different levels is a bit ugly IMO but doable.

A possible way to move serialization one level "up", i.e: from `EventField` to `Event` (which is a small variation of the above snippet) would be to have something like:



```rust
trait Event {
   fn from_raw_section(&BpfRawSection) -> Result<Self>
   fn from_json(&serde::Value) -> Result<Self>
   fn to_json(&self) -> Result<serde::Value>
   fn to_fields(&self, mut &Vec[EventFields]) -> Result<()>
}

// in mycollector.rs
// We could use the same object for json and binary or not.
#[derive(Default, Deserialize, Serialize)]
#[repr(C, packed)]
struct MyEventData {
    data1: u8,
    data2: u32,
}

// Not sure if we need the wrapper or not
struct MyEvent {
    data: MyEventData,
}

impl Event for MyEvent {
   // Pretty much what we have now...
   fn from_raw_section(raw_section: &BpfRawSection) -> Result<Self> {
       let mut data = MyEventData::default();
       plain::copy_from_bytes(&mut data, &raw_section.data);
       Ok(MyEvent{data})
   }
   
    // ... but split in two parts
   fn to_fields(&self, fields mut &Vec[EventFields]) -> Result<()> {
        fields.push(event_field!("data1", self.data.data1));
        fields.push(event_field!("data2", self.data.data1));
   }
   
   // to_json is optional!! The default implementation calls fields() and perform field-level marshalling just as we do now into a json map. That way we don't _need_ to implement Serialize.
   // fn to_json(&self) -> Result<serde::Value>{...} 
   
    // However, from_json needs to be implemented. There is no way we can write a default implementation becasuse we don't know the types.  
   fn from_json(json_data: &serde::Value) -> Result<Self> {
       let mut data = MyEventData::default();
       serde_json::from_json(&mut data, json_data);
       Ok(MyEvent{data})
   }
}

```

amorenoz: Wondering if @vlrpl has any thoughts.

src/core/events/bpf/include/events.h
```diff
@@ -0,0 +1,82 @@
...
+
+struct trace_raw_event {
+	u16 size;
+	u8 data[RAW_EVENT_DATA_SIZE];
+} __attribute__((packed));
+
+/* Please keep synced with its Rust counterpart. */
+struct trace_raw_event_section_header {
+	u8 owner;
+	u8 data_type;
```
amorenoz, line 24: I don't see an  enum declared for this. The value `1` hardcoded both in the bfpf and the rust side. I assume `1` means "custom data type" but I think we should have it declared and not hardcoded.

amorenoz: Oh, ok. Then I miss-interpreted it. I thought it was some kind of reference to the actual type of data that it contained, the use of `DATA_TYPE_U128` in the unit tests got me confused. It's more an "id" than a "type".

amorenoz:
Having concluded that regardless of the decision we make on https://github.com/net-trace/packet-tracer/pull/53#discussion_r1048290385 it won't change the infrastructure significantly, I think the PR is in a good shape.

amorenoz (approved):

vlrpl:
> Having concluded that regardless of the decision we make on [#53 (comment)](https://github.com/net-trace/packet-tracer/pull/53#discussion_r1048290385) it won't change the infrastructure significantly, I think the PR is in a good shape.

Agreed. The discussion was quite interesting, but I have the feeling that a preference will stand out based on the level of flexibility/consistency we'd need. Looked at the PR from a formal point of view as well and LGTM too. Any potential overlook (if any) may perfectly be addressed in subsequent patches, considering this series is needed for multiple reasons. Going to merge it.

vlrpl (approved):

PR #55, opened 2022-12-02
Allow cli users to add kernel probes & rework the internal symbol representation

atenart:
Using cli parameters, allow users to add kernel probes on what we currently support: kprobe and raw_tracepoints. This is implemented a level before the actual collectors as kernel probes is really at the core of our tool and isn't linked to any collector being used in particular.

This PR comes with lots of internal code move, to prepare for coming additions (and some were actually required to implement the user defined probes support).

Based on #53.

### DIFF
--- a/src/core/probe/probe.rs
+++ b/src/core/probe/probe.rs
@@ -1,5 +1,3 @@
-#![allow(dead_code)] // FIXME
-
 use std::fmt;
 
 use crate::core::kernel;
--- a/build.rs
+++ b/build.rs
@@ -1,5 +1,3 @@
-#![allow(dead_code)] // FIXME
-
 use std::{
     fs::{create_dir_all, File},
     io::Write,
--- a/src/collect/cli.rs
+++ b/src/collect/cli.rs
@@ -1,7 +1,6 @@
 //! # Collect
 //!
 //! Collect is a dynamic CLI subcommand that allows collectors to register their arguments.
-#![allow(dead_code)] // FIXME
 
 use anyhow::Result;
 use std::any::Any;
--- a/src/core/events/bpf.rs
+++ b/src/core/events/bpf.rs
@@ -1,6 +1,6 @@
 //! Handles the BPF to Rust event retrieval and the unmarshaling process.
 
-#![allow(dead_code)] // FIXME
+#![cfg_attr(test, allow(dead_code))]
 #![cfg_attr(test, allow(unused_imports))]
 
 use std::{
--- a/src/core/kernel/symbol.rs
+++ b/src/core/kernel/symbol.rs
@@ -1,5 +1,3 @@
-#![allow(dead_code)] // FIXME
-
 use std::fmt;
 
 use anyhow::{bail, Result};
--- a/src/core/probe/kernel/config.rs
+++ b/src/core/probe/kernel/config.rs
@@ -34,7 +34,8 @@ pub(super) struct ProbeConfig {
 
 unsafe impl plain::Plain for ProbeConfig {}
 
-#[allow(dead_code)] // When testing this isn't used as the config map is hidden.
+// When testing this isn't used as the config map is hidden.
+#[cfg_attr(test, allow(dead_code))]
 pub(super) fn init_config_map() -> Result<libbpf_rs::Map> {
     let opts = libbpf_sys::bpf_map_create_opts {
         sz: mem::size_of::<libbpf_sys::bpf_map_create_opts>() as libbpf_sys::size_t,
--- a/src/core/probe/kernel/inspect.rs
+++ b/src/core/probe/kernel/inspect.rs
@@ -1,5 +1,3 @@
-#![allow(dead_code)] // FIXME
-
 use anyhow::Result;
 
 use super::config::ProbeConfig;


src/core/kernel/kernel.rs
```diff
@@ -0,0 +1,239 @@
...
+use anyhow::{bail, Result};
+
+use super::{inspect, symbols};
+
+/// Kernel symbol representation. Only supports traceable symbols: events and
+/// functions.
+pub(crate) struct Symbol {
+    r#type: SymbolType,
+    name: String,
+}
```
amorenoz: Looking at how it's used and what I'm working on for usdt/uprobe, what do you think about this:

```rust
struct TracePointSymbol {
   provider: String,
   name: String,
}
enum Symbol {
    Kprobe(String),
    Tracepoint(TracePointSymbol)
    // Later
    USDT(USDTSymbol)
} 
// Later
struct USDTSymbol {
    pid: pid_t,
    path: PathBuf,
    provider: String
    name: String
}
```
Having userspace probes share the Symbol representation will make the APIs much more unified.

src/core/probe/kernel/kernel.rs
```diff
@@ -253,6 +265,27 @@ impl Kernel {
...
+    /// ```
+    /// mod hook {
+    ///     include!("bpf/.out/hook.rs");
+    /// }
+    ///
+    /// [...]
+    ///
+    /// kernel.register_hook_to(hook::DATA, ProbeType::Kprobe, "kfree_skb_reason")?;
+    /// ```
+    pub(crate) fn register_hook_to(
```
amorenoz, line 289: Why keep this API? The new one is much better :-) and completely supersedes this one.

src/collector/cli.rs
```diff
@@ -18,6 +18,22 @@ pub(crate) struct CollectArgs {
...
+    #[arg(
+        id = "probe",
+        short,
+        long,
+        help = "Add a probe on the given target. Can be used multiple times. Probes should
+follow the TYPE:TARGET pattern.
+
+Valid TYPEs:
+- kprobe: kernel probes.
+- tp: tracepoints.
```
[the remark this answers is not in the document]
amorenoz: OK, I see your point.

>     Tool level: we do target the kernel, whatever the collector & config combinations. We might not actually install a probe in certain cases but doing so would still make sense.
>     Collect command level: here, and I'll develop later, we do target a given area. For now, networking. This includes the kernel, and potentially userspace components (which is quite important, don't get me wrong).
>     Collector level: here we do target a specific use case or data type. And this includes userspace components.

Does that mean that a collector cannot add a probe that was not added in the `collect` command? So, if you enable the `ovs` collector, or the `conntrack` collector, or the `ipsec` collector but you don't explicitly know the names of the probes and add them to the cli, they won't work?
If the answer is yes, then it's quite a burden for the user and they need to know where to look (as opposed to the collectors to encoding the knowledge of where to look for certain data).
If the answer is no, then it's not clear what collectors are affected by a global `--probe`.

Also, I'm concerned about the clarity of the UX when some probes just don't apply to some collectors. For instance, what does `packet-tracer collect --collectors=conntrack,ovs --probe kfree_skb` mean? Only the conntrack and ovs collectors are enabled and none of them collect any information from the `kfree_skb`. Will I see an empty event with no data at all on each `kfree_skb`? Will I get events from other conntrack-related or ovs-related kprobes/tracepoints?

src/core/kernel/kernel.rs
amorenoz: I think my problem is that `Symbol` and `Target` are, as you argue, two conceptually different things but they seem interleaved.

collectors can attach to `Symbol`s or `Target`s
```rust
impl Kernel {
[...]
pub(crate) fn register_hook_to_symbol(
        &mut self,
        hook: Hook,
        r#type: ProbeType,
        symbol: Symbol,
    ) -> Result<()>
    
  pub(crate) fn register_hook_to(&mut self,
        hook: Hook,
        r#type: ProbeType,
        target: &str,
    ) -> Result<()> {
        let symbol = Symbol::from_name(target)?;
        self.register_hook_to_symbol(hook, r#type, symbol)
    } 
}
```

When `Kernel` sends the request down to probes they receive a `Symbol`.
```rust
impl ProbeBuilder for KprobeBuilder {
fn attach(&mut self, symbol: &Symbol, _: &TargetDesc) -> Result<()>
}
impl ProbeBuilder for RawTracepointBuilde {
fn attach(&mut self, symbol: &Symbol, _: &TargetDesc) -> Result<()>
}
```

But it's possible that a probe type cannot be added to a symbol, so `Symbol` has a `r#type` (which for me is a strong indicator that can be made into a rust variant) and a `can_probe` that yields a 1:1 relationship between `Symbol.type` and `TargetType`. Also, the `Symbol.attach_name()` depends on the `Symbol.type`, ultimately on the `TargetType`.

So for me a more accurate data model would be:
```rust
enum Target {
    Kprobe(Kprobe)
    Tracepoint(Tracepoint)
    USDT(USDT)
    Uprobe(Uprobe)
}

impl Target {
   /// From strings such as: "kprobe::func_name" (or just "func_name"), "tp::name", "usdt:provider:name", "uprobe::func_name".
   fn from_string(str) -> Result<Self> 
   /// We may want to get the symbol from the target. Some targets (e.g: USDT) might not have symbols associated!
   fn get_symbol(&self) -> Result<Option<Symbol>>
}

enum Symbol {
   Kernel(KernelSymbol)
   User(UserSymbol)
}

```

Collectors can still use `Symbol` to inspect features on kernel or userspace programs. We can even build _concrete_ `Target`s from `Symbol`s if we want (Kprobe.from_symbol(KernelSymbol)). However, the probe attachment API (a.k.a `ProbeBuilder` would just receive a `Target` which has all the information we need to attach to it.

I'm not saying this should be done in this PR but I may need to do this in order to add USDT support. WDYT?

amorenoz: We can make `Target` a trait instead but ultimately `ProbeBuilder`s will have to `downcast` them (or `match` them)

src/collector/cli.rs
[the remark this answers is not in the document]
amorenoz: Could we
- Issue a warning if a probe added by the user is not used by any enabled collector
- When this happens, have a "default hook" that copies (as best as it can) the probe arguments? With the TLV event format I think it should be possible. 
?

.mailmap
```diff
@@ -0,0 +1,3 @@
+Adrian Moreno <amorenoz@redhat.com> <52785490+amorenoz@users.noreply.github.com>
```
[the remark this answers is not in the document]
amorenoz: And why do we need the map?

src/core/kernel/kernel.rs
amorenoz: I think we're getting somewhere :)

> What would be the advantage of this? I believe giving a (ProbeType, Symbol) tuple or a Target is exactly the same, as a Target is a Symbol associated with a ProbeType. Is there a specific blocking point for USDT support here? Is it we can attach per-target data using this approach?

I guess my initial inclination towards Symbol -> Target rather than the opposite is that I didn't see USDT as a symbol, i.e: it's not a function. But if I mentally relax the meaning of Symbol then I think it's OK.


> 
> enum Symbol {
>     Kernel(KernelSymbol),
>     Userspace(UserspaceSymbol),
> }

Is `KernelSymbol` itself an enum with `Function` and `Tracepoint` variants?

> trait SymbolTrait {
>     fn name(&self) -> &str;
>     fn attach_name(&self) -> &str;
>     fn addr_name(&self) -> &str;
>     fn typedef_name(&self) -> &str;
>     fn addr(&self) -> Result<u64>;
>     fn nargs(&self) -> Result<u32>;
>     fn parameter_offset(&self, parameter_type: &str) -> Result<Option<u32>>;
>     [...]
>     // With the following we can remove `can_probe()`
>     fn to_probe(&self, r#type: ProbeType) -> Result<Probe>;
> }

If all variants need to implement all of these we might as well use a `dyn Symbol` here but I think User and Kernel symbols might be different enough to fit into a clean trait. E.g: `typedef_name` does not make sense for USDT, all userspace symbols need `pid` and `path` which doesn't make sense for kernel, etc.

Alternatively, if things like `name`, `attach_name`, `addr_name`, etc are going to be used inside each `ProbeBuilder` they could be implemented only if needed on each variant since each `ProbeBuilder` would have to `match` the concrete `Probe` type variant type anyhow.

Finally, are you planning to do this in this PR or should I do this rework as part of the USDT work?

src/collector/cli.rs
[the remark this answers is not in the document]
amorenoz: > 
>     * When this happens, have a "default hook" that copies (as best as it can) the probe arguments? With the TLV event format I think it should be possible.
> 
> We'll end up with addresses of struct and some values. Not sure how useful would that be, our first goal is to retrieve and make sense out of the data. Idk, I would say we don't need this, but this can still be added at some point if people ask for it.

I'm trying to make sense out of `packet-tracer collect --collectors=conntrack,ovs --probe kfree_skb` as in: "Please load the ovs and conntrack collectors because I want full info on those modules but also record kfree_skb events (just copy the args) because for some reason I find it relevant and I don't specifically need the skb module, maybe I just want to correlate them by hand or something".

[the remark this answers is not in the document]
amorenoz: If we don't load anything and capture those events, we are left with 2 options:
- print a warning and ignore the argument
- print an error an exit
In both cases it's not a great UX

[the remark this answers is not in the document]
amorenoz: Also, what does `packet-tracer collect --colectors=skb --probe kfree_skb --probe some_other_skb_func` mean? Skb collector will _only_ probe those targets? or it'll probe all of the targets defined by skb collector _plus_ those ones?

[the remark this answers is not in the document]
amorenoz: > Also, what does `packet-tracer collect --colectors=skb --probe kfree_skb --probe some_other_skb_func` mean? Skb collector will _only_ probe those targets? or it'll probe all of the targets defined by skb collector _plus_ those ones?

You already answered :-) I guess those will be additional? i.e: if not already covered by the collector itself really. Correct?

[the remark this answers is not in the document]
amorenoz: > Yep. In general I don't believe collectors will add lots of kernel probes, except for special cases like tracking. Automatically adding probes would be more the job of the profile.

That would put a burden to the user who needs to know all the names of the functions and tracepoints they want, right? I mean, not all of us are kernel developers :-)

I would expect collectors to have sane defaults, i.e: I would expect `packet-tracer --collectors=socket,skb` to generate some events on well-known socket and skb places without me having to pass a huge list of function names.


> I guess with `packet-tracer collect --collectors=conntrack,ovs --probe kfree_skb`, we are left with:
> 
>     1. Ignoring the fact that the above does not make sense.
> 
>     2. Reporting at least the parameter values. I'm not convinced this really improved the UX as the data will probably be hard to use (skb won't be tracked, other pointers won't make sense, actual values are not that common).
> 
>     3. Issue an error if no data will be reported; for this we need to validate the user defined probes with the collectors in some way.
> 
> 
> To me, option 2 would be nice to have but as a collector itself, because while it looks like something that might be useful I don't think it would improve the UX much as a fallback. It would still be confusing. This leaves us with option 3.
> 
> Having said that I also think option 1 is OKish. We can't expect magic to happen when the configuration does not. The only thing I would be worried is if this "nonsense" stays invisible. But with the common event part we would see events coming from those probes, with just the symbol name + timestamp (which isn't useful as-is, but well).
> 
> WDYT?

As a design principle I always try to create APIs that are difficult - if not impossible - to use wrongly.

Let's take 2 with the behavior implemented as a default collector. Now, let's say this collector picks up generic data (cpu_id, thread_id, com, args, timestamp, etc). This is pretty much what perf would do which is pretty cool for our baseline :-D. Besides, this probes might also be affected by filters. E.g: ` --cpu` or `--pid` filter.  In fact, let's add this collector and have it attach to anything!

So, my preference would be 2 + default collector as end goal, 1 or 3 for now.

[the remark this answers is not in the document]
amorenoz: > Agreed, the default settings should probe a set of generic functions. In my mind this is however the job of the profiles. We can have a default one, providing what you would expect, and then targeted ones for specific use cases. If the user doesn't want that, we should also provide an option to not load a profile.

Another idea: make the probes have sane defaults so the CLI can work without profile and have an option "--no-defaults" that disables these default probes (and forces their explicit configuration). All profiles define this option. This option could even be internal-only and tied to the loading of the profile itself.

> Well, after all, we already support 2 (kind of). The common part of the event will always be reported, and it contains generic data. It's just a matter of adding a few more like cpu id, thread id, com, etc. The issue with using a dedicated collector for this is, by design, it will be possible to disable it; or we would need to have two kinds of collectors. (Supporting args retrieval might be a little too complex for this common event part thought).

Just not register it with the CLI?

Another thought: I think we already have two kinds of collectors (in mind):

- "Generic" collectors: those that collect data on any probe based on automagically detected args.
  - They accept "--probes"
  - Examples: skb, socket?, etc
- "Targeted" ones that probe a very specific set of well-known probes to extract information from small areas
  - "--probes" means nothing to these
  - Examples: : ovs, conntrack, nftables, ipsec, tcp, upd, bridge, etc.

Maybe making this difference explicit might help with our UX concerns.

> Another solution is to always have the skb module loaded, which would retrieve only basic info by default.

I don't think that's a good idea. Even if it affects a limited set of probes, the number of events could be huge on production setups and if it's not what the user is looking for, it could be a huge amount of noise.

src/core/kernel/kernel.rs
amorenoz: > I feel there's something not quite right about this. It looks to me the `Symbol` now would have two variant not sharing much, and not used in the same code paths. To split shared helpers from specific ones we can have a trait + variant specific implementations, but that comes with a cost when using those.
> 
> Or I might be a bit confused with all the changes :) I have a few questions:
> 
>     * Where do you see the `Symbol` obj being used, where both variants makes sense?
> 

In all the `ProbeSet` / `ProbeBuilder` APIs. I'm trying to avoid having an almost identical copy of `core/probe/kernel/kernel.rs`.

>     * What code could be reused between the two `Symbol` variants?
> 

The way I see it, `enums` are not about sharing code but about sharing APIs. Conceptually, user probes also have `Target`s or `Probe`s, which also have `Symbols`, and also need `ProbeBuilder`s to attach `Hooks` to them. The alternative to create uprobe/usdt support is either duplicating those structs, or duplicating the functions on those structs.

amorenoz: Maybe I should just start by duplicating everything and we can later see if we can refactor something in some common part.

src/core/events/bpf/include/events.h
```diff
@@ -0,0 +1,82 @@
...
+
+struct trace_raw_event {
+	u16 size;
+	u8 data[RAW_EVENT_DATA_SIZE];
+} __attribute__((packed));
+
+/* Please keep synced with its Rust counterpart. */
+struct trace_raw_event_section_header {
+	u8 owner;
+	u8 data_type;
```
amorenoz, line 24: I don't see an  enum declared for this. The value `1` hardcoded both in the bfpf and the rust side. I assume `1` means "custom data type" but I think we should have it declared and not hardcoded.

src/core/probe/probe.rs
```diff
@@ -0,0 +1,39 @@
...
+            Probe::RawTracepoint(_) => 1,
+        }
+    }
+}
+
+/// Allow nice log messages.
+impl fmt::Display for Probe {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let name = match self {
+            Probe::Kprobe(_) => "kprobe",
```
amorenoz, line 32: You might want to add the symbol name?

src/core/probe/probe.rs
```diff
@@ -0,0 +1,39 @@
...
+    RawTracepoint(kernel::Symbol),
+}
+
+// Use mem::variant_count::<Probe>() when available in stable.
+pub(crate) const PROBE_VARIANTS: usize = 2;
+
+impl Probe {
+    /// We do use probe types as indexes, the following makes it easy.
+    ///
+    /// TODO: use #[repr(usize)] when Rust 1.66.0 is generally available.
```
amorenoz, line 19: Arg! I don't think think `#[repr(usize)]` is going to help us here: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=829cf1e3779502552116a987a9feff8b

It seems that if we use a Variant with actual types (in this case `kernel::Symbol`s), the `repr(usize)` does not work as expected.
Am I missing something?

amorenoz: Also, looking at `Kernel` I see two options for the implementation of Userspace probes:
A) `struct Kernel` is reused (and renamed into something like `ProbeMgr`, and distinguish between user and kernel probes internally (e.g: userspace probes don't accept dynamic probes, nor more than one hook per probe, nor config map, etc)
B) We add `struct User` with a very similar API nicely based on `Probe` object. Now, if that's the case `Probe` will include USDT but `Kernel` will not need to increase the size of its vectors...

Thoughts?

src/core/probe/probe.rs
amorenoz: I think it'll be useful. I'm planning to a global log level switch so we can start adding `debug` traces.

src/core/probe/probe.rs
amorenoz: > To be explicit, I'm suggesting to move the `Kernel` implementation one module up and keep the kernel-specific implementations inside `core::probe::kernel`.

See #62 as a proposed implementation of this

amorenoz:
This PR does, indeed, serve as foundation for a more compact and generic API. So I think it's OK to merge it.

vlrpl:
Besides a couple of nits that definitely don't worth a respin, the series LGTM.
I have my series on top of this, so it's been tested to some extent, and I couldn't spot any problem.
Had some questions, but the answers were already in the previous discussions.
Thanks!

vlrpl (approved):

PR #64, opened 2023-01-19
Initial support for skb data retrieval (aka skb module events)

atenart:
Let's put an emphasis on *initial*:

- Logic to retrieve more data, and/or to get some data from other sources can and will be added later.
- Event field names might change.
- There are probably some corner cases not supported, e.g. probing certain functions means parts of the skb wasn't processed yet and accessing this data might not be possible (w/o implementing a full packet parser) or not set yet fields resulting in invalid data[1].

However I believe this PR is in a good shape to provide an example of what data collection from a module should look like and might trigger some discussions; and good enough to be used already for retrieving common fields in well known L2 to L4 protocols as well as from the skb metadata.

[1] So far I put measures in place to fix the one I know and found.

### DIFF
--- a/src/core/events/bpf/include/events.h
+++ b/src/core/events/bpf/include/events.h
@@ -13,6 +13,7 @@ enum trace_event_owners {
 	KERNEL = 2,
 	USERSPACE = 3,
 	COLLECTOR_SKB_TRACKING = 4,
+	COLLECTOR_SKB = 5,
 };
 
 struct trace_raw_event {
--- a/src/core/events/bpf/include/events.h
+++ b/src/core/events/bpf/include/events.h
@@ -76,6 +76,19 @@ static __always_inline void *get_event_section(struct trace_raw_event *event,
 	return section;
 }
 
+/* Similar to get_event_section but initialize the section data to 0s. */
+static __always_inline void *get_event_zsection(struct trace_raw_event *event,
+						u8 owner, u8 data_type, const u16 size)
+{
+	void *section = get_event_section(event, owner, data_type, size);
+
+	if (!section)
+		return NULL;
+
+	__builtin_memset(section, 0, size);
+	return section;
+}
+
 struct common_event {
 	u64 timestamp;
 } __attribute__((packed));
--- a/src/module/skb/bpf.rs
+++ b/src/module/skb/bpf.rs
@@ -22,6 +22,8 @@ use crate::{
 pub(super) const SECTION_L2: u64 = 0;
 pub(super) const SECTION_IPV4: u64 = 1;
 pub(super) const SECTION_IPV6: u64 = 2;
+pub(super) const SECTION_TCP: u64 = 3;
+pub(super) const SECTION_UDP: u64 = 4;
 
 /// Global configuration passed down the BPF part.
 #[repr(C, packed)]
@@ -131,6 +133,69 @@ pub(super) fn unmarshal_ipv6(
     fields.push(event_field!("ip_version", 6));
     fields.push(event_field!("l3_len", u16::from_be(event.len)));
     fields.push(event_field!("protocol", event.protocol));
+
+    Ok(())
+}
+
+/// TCP data retrieved from skbs.
+#[derive(Default)]
+#[repr(C, packed)]
+struct SkbTcpEvent {
+    /// Source port. Stored in network order.
+    sport: u16,
+    /// Destination port. Stored in network order.
+    dport: u16,
+    /// Sequence number. Stored in network order.
+    seq: u32,
+    /// Ack sequence number. Stored in network order.
+    ack_seq: u32,
+    /// TCP window. Stored in network order.
+    window: u16,
+    /// TCP flags (from low to high): fin, syn, rst, psh, ack, urg, ece, cwr.
+    flags: u8,
+    /// TCP data offset: size of the TCP header in 32-bit words.
+    doff: u8,
+}
+unsafe impl Plain for SkbTcpEvent {}
+
+pub(super) fn unmarshal_tcp(
+    raw_section: &BpfRawSection,
+    fields: &mut Vec<EventField>,
+) -> Result<()> {
+    let event = parse_raw_section::<SkbTcpEvent>(raw_section)?;
+
+    fields.push(event_field!("sport", u16::from_be(event.sport)));
+    fields.push(event_field!("dport", u16::from_be(event.dport)));
+    fields.push(event_field!("tcp_seq", u32::from_be(event.seq)));
+    fields.push(event_field!("tcp_ack_seq", u32::from_be(event.ack_seq)));
+    fields.push(event_field!("tcp_window", u16::from_be(event.window)));
+    fields.push(event_field!("tcp_flags", event.flags));
+
+    Ok(())
+}
+
+/// UDP data retrieved from skbs.
+#[derive(Default)]
+#[repr(C, packed)]
+struct SkbUdpEvent {
+    /// Source port. Stored in network order.
+    sport: u16,
+    /// Destination port. Stored in network order.
+    dport: u16,
+    /// Lenght: length in bytes of the UDP header and UDP data. Stored in network order.
+    len: u16,
+}
+unsafe impl Plain for SkbUdpEvent {}
+
+pub(super) fn unmarshal_udp(
+    raw_section: &BpfRawSection,
+    fields: &mut Vec<EventField>,
+) -> Result<()> {
+    let event = parse_raw_section::<SkbUdpEvent>(raw_section)?;
+
+    fields.push(event_field!("sport", u16::from_be(event.sport)));
+    fields.push(event_field!("dport", u16::from_be(event.dport)));
+    fields.push(event_field!("udp_len", u16::from_be(event.len)));
 
     Ok(())
 }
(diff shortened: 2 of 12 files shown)


src/module/skb/skb.rs
```diff
@@ -1,17 +1,30 @@
...
+#[derive(Args, Default)]
 pub(crate) struct SkbCollectorArgs {
+    #[arg(
+        long,
+        help = "Comma separated list of data to collect from SKBs.
+
+Possible values: all, l2.
+Default value:"
+    )]
+    skb_sections: Option<String>,
```
amorenoz, line 27: clap supports arguments with type `Vec<String>` which can hold multiple values and support default values and "possible values" (the `--collectors` is done this way).

I believe this could be implemented this way making it much simpler since clap would autogenerate the help.

I tried a little experiment to confirm it works with the Derive interface:

```rust
use clap::{arg, builder::PossibleValuesParser, Parser};

#[derive(Parser)]
#[command(about = "I am a program", long_about = None)]
struct Cli {
    #[arg(short,
          long,
          value_parser=PossibleValuesParser::new(["foo", "bar", "baz", "all"]),
          value_delimiter=',',
          default_value="foo,bar")]
    opt: Vec<String>,
}

fn main() {
    let cli = Cli::parse();
    println!("{:?}", cli.opt);
}
```
```shell
> ./target/debug/clappossible -h
I am a program and I work, just pass `-h`

Usage: clappossible [OPTIONS]

Options:
  -o, --opt <OPT>  [default: foo,bar] [possible values: foo, bar, baz, all]
  -h, --help       Print help

> ./target/debug/clappossible 
["foo", "bar"]
> /target/debug/clappossible --opt bar,baz
["bar", "baz"]
```

src/module/skb/bpf.rs
```diff
@@ -0,0 +1,81 @@
...
+    raw_section: &BpfRawSection,
+    fields: &mut Vec<EventField>,
+) -> Result<()> {
+    let event = parse_event::<SkbL2Event>(raw_section)?;
+
+    fields.push(event_field!("etype", u16::from_be(event.etype)));
+    fields.push(event_field!(
+        "src",
+        format!(
+            "{:02x}:{:02x}:{:02x}:{:02x}:{:02x}:{:02x}",
```
amorenoz, line 80: Should we use a proper struct instead of a string? I guess things like offline filtering and comparison would be easier

src/module/skb/bpf.rs
```diff
@@ -150,3 +152,101 @@ pub(super) fn unmarshal_ipv6(
 
     Ok(())
 }
+
+/// TCP data retrieved from SKBs.
```
amorenoz, line 161: NIT: There seems to be a bit of inconsistency in the use of "SKB" or "skb"

src/module/skb/bpf.rs
```diff
@@ -310,3 +311,21 @@ pub(super) fn unmarshal_dev(
...
+    netns: u32,
+}
+unsafe impl Plain for SkbNsEvent {}
+
+pub(super) fn unmarshal_ns(
+    raw_section: &BpfRawSection,
+    fields: &mut Vec<EventField>,
+) -> Result<()> {
+    let event = parse_event::<SkbNsEvent>(raw_section)?;
+    fields.push(event_field!("netns", event.netns));
```
amorenoz, line 330: Should we have a netns num-to-name resolution in userspace? The number will probably little to the user

src/module/skb/bpf.rs
```diff
@@ -0,0 +1,81 @@
...
+use anyhow::{bail, Result};
+use plain::Plain;
+
+use crate::{
+    core::events::{bpf::BpfRawSection, EventField},
+    event_field,
+};
+
+/// Helper to check a raw section validity and parse it into a structured type.
+fn parse_event<T>(raw_section: &BpfRawSection) -> Result<T>
```
amorenoz, line 22: This helper is very convenient! Maybe move it to some common place so it can be used by other modules?

src/module/skb/bpf/skb_hook.bpf.c
```diff
@@ -0,0 +1,322 @@
...
+		bpf_probe_read_kernel(&e->type, sizeof(e->type), &icmp->type);
+		bpf_probe_read_kernel(&e->code, sizeof(e->code), &icmp->code);
+	}
+
+	return 0;
+}
+
+/* Must be called with a valid skb pointer */
+static __always_inline int process_skb(struct trace_context *ctx,
+				       struct event *event, struct sk_buff *skb)
```
amorenoz, line 240: It's actually `struct trace_raw_event`.

amorenoz: I was quite shocked to see this error fall through so I created #77

src/core/events/bpf/include/events.h
```diff
@@ -76,6 +77,19 @@ static __always_inline void *get_event_section(struct trace_raw_event *event,
...
+/* Similar to get_event_section but initialize the section data to 0s. */
+static __always_inline void *get_event_zsection(struct trace_raw_event *event,
+						u8 owner, u8 data_type, u16 size)
+{
+	void *section = get_event_section(event, owner, data_type, size);
+
+	if (!section)
+		return NULL;
+
+	memset(section, 0, size);
```
amorenoz, line 89: This raises a warning in clang, we could either  macro it or use  `__builtin_memset` or even `__builtin_memset_inline`

src/module/skb/bpf.rs
```diff
@@ -0,0 +1,361 @@
...
+unsafe impl Plain for SkbIpv6Event {}
+
+pub(super) fn unmarshal_ipv6(
+    raw_section: &BpfRawSection,
+    fields: &mut Vec<EventField>,
+) -> Result<()> {
+    let event = parse_event::<SkbIpv6Event>(raw_section)?;
+
+    let src = Ipv6Addr::from(u128::from_be(event.src));
+    fields.push(event_field!("saddr", format!("{}", src)));
```
amorenoz, line 150: All these `format!()` calls raise a warning on rust v1.69.0:

```
warning: variables can be used directly in the `format!` string                                                                                                                                                                                                                   
   --> src/module/skb/bpf.rs:136:39                                                                                                                                                                                                                                               
    |                                                                                                                                                                                                                                                                             
136 |     fields.push(event_field!("saddr", format!("{}", src)));                                                                                                                                                                                                                 
    |                                       ^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                    
    |                                                                                                                                                                                                                                                                             
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
help: change this to
    |
136 -     fields.push(event_field!("saddr", format!("{}", src)));
136 +     fields.push(event_field!("saddr", format!("{src}")));
```

src/module/skb/bpf.rs
```diff
@@ -150,3 +152,101 @@ pub(super) fn unmarshal_ipv6(
...
+    fields: &mut Vec<EventField>,
+) -> Result<()> {
+    let event = parse_event::<SkbTcpEvent>(raw_section)?;
+
+    fields.push(event_field!("sport", u16::from_be(event.sport)));
+    fields.push(event_field!("dport", u16::from_be(event.dport)));
+    fields.push(event_field!("tcp_seq", u32::from_be(event.seq)));
+    fields.push(event_field!("tcp_ack_seq", u32::from_be(event.ack_seq)));
+    fields.push(event_field!("tcp_window", u16::from_be(event.window)));
+    fields.push(event_field!("tcp_fin", event.fin()));
```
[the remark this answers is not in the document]
vlrpl: No strong opinion, just a personal preference: I tend to prefer the single field mostly because as you said it doesn't bloat the output.

src/module/skb/skb.rs
```diff
@@ -1,17 +1,30 @@
...
 const SKB_COLLECTOR: &str = "skb";
 
-#[derive(Args)]
+#[derive(Args, Default)]
 pub(crate) struct SkbCollectorArgs {
+    #[arg(
+        long,
+        help = "Comma separated list of data to collect from SKBs.
+
+Possible values: all, l2.
```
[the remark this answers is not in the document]
vlrpl: makes sense. It was just a remark based on the hands-on-without-reading-the-doc expectation.
In any case, I'm ok to keep it as it is.

src/module/skb/bpf.rs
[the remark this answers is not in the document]
vlrpl: that would be really nice. Having something like `tcp_flags: fin|ack` would be really readable and at the same time will not bloat the output

src/module/skb/skb.rs
```diff
@@ -25,20 +43,78 @@ impl Collector for SkbCollector {
...
         &mut self,
-        _: &CliConfig,
+        cli: &CliConfig,
         probes: &mut ProbeManager,
-        _events: &mut BpfEvents,
+        events: &mut BpfEvents,
     ) -> Result<()> {
-        probes.register_kernel_hook(Hook::from(skb_hook::DATA))?;
+        // First, get the cli parameters.
+        let args = cli.get_section::<SkbCollectorArgs>("skb")?;
```
vlrpl, line 56: can't we use SKB_COLLECTOR here?

amorenoz:
I'm happy with the shame of this PR.

amorenoz (approved):

amorenoz:
Let's wait for @vlrpl 's ack and we can merge

PR #66, opened 2023-01-27
Fix the shared library detection logic to avoid being killed by the oom

atenart:
When parsing a process we're looking for all shared libraries mentioned in /proc/<pid>/maps. The issue was we were only rejecting paths matching the process one, and letting various files which where not actually a shared library. This become an issue when entries like /dev/zero where found as we ended up endlessly reading those until we were killed by the oom. This was the case when running our runtime unit tests as when a program loads an eBPF object, /dev/zero is found in its maps file.

Fix this by reworking the shared library detection.

### DIFF
--- a/src/core/user/proc.rs
+++ b/src/core/user/proc.rs
@@ -17,7 +17,7 @@ use byteorder::BigEndian as Endian;
 #[cfg(target_endian = "little")]
 use byteorder::LittleEndian as Endian;
 use byteorder::ReadBytesExt;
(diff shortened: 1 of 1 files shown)


vlrpl (approved):
The series LGTM, and works as expected.
Will just approve for the time being as we have some failures in our ci (fix pending #69).
Once ci passes, we can proceed to merge this.
Thanks!

vlrpl (approved):
thanks for the fix!

PR #67, opened 2023-01-27
Support for writing events to a file (& some related changes)

atenart:


src/cli/cli.rs
```diff
@@ -74,8 +74,22 @@ impl Debug for dyn SubCommand {
...
 ///
 /// packet-tracer is a tool for capturing networking-related events from the system using ebpf and analyzing them.
-#[derive(Args, Default, Debug)]
-pub(crate) struct MainConfig {}
+#[derive(Args, Debug, Default)]
+pub(crate) struct MainConfig {
+    #[arg(
+        long,
+        default_value = "warn",
+        help = "Log level: error, warn, info, debug. Default: warn."
```
amorenoz, line 82: Default warn makes the tool really silent (hopefully we don't have warnings :-D).
Not even "hey we are capturing" or the like.
How about default=info?

src/main.rs
```diff
@@ -10,21 +10,37 @@ use cli::get_cli;
...
+            "debug" => LevelFilter::Debug,
+            x => bail!("Invalid log_level: {}", x),
+        };
+        TermLogger::init(
+            log_level,
+            Config::default(),
+            TerminalMode::Stderr, // Use stderr so logs do not conflict w/ other output.
+            ColorChoice::Auto,
+        )?;
+    }
```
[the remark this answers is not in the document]
amorenoz: How about a third option? Inspiration: tcpdump.
`--log-level` controls the logs (we can allow `--log-level off` to disable them but that's probably not super useful)
`--out` (-w in tcpdump) writes data to a file. By default this disables stdout printing.
`--print` when used in combination with `--out` re-enables stdout printing

src/main.rs
```diff
@@ -10,21 +10,37 @@ use cli::get_cli;
...
+            "error" => LevelFilter::Error,
+            "warn" => LevelFilter::Warn,
+            "info" => LevelFilter::Info,
+            "debug" => LevelFilter::Debug,
+            x => bail!("Invalid log_level: {}", x),
+        };
+        TermLogger::init(
+            log_level,
+            Config::default(),
+            TerminalMode::Stderr, // Use stderr so logs do not conflict w/ other output.
```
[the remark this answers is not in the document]
amorenoz: I love tools that do this. +1

src/cli/cli.rs
amorenoz: Also, consider using clap's PosibleValueParser

src/cli/cli.rs
```diff
@@ -72,8 +74,22 @@ impl Debug for dyn SubCommand {
...
+    )]
+    pub(crate) log_level: String,
+    #[arg(
+        short,
+        long,
+        default_value = "false",
+        help = "Suppress stdout messages."
+    )]
+    pub(crate) quiet: bool,
```
amorenoz, line 91: Does it make sense a global --quite flag? what does `retis --quiet analyze`  mean? I'm thinking maybe this is more a `collect` option linked to `--out` as already exposed in another comment.

src/collect/cli.rs
```diff
@@ -34,6 +36,8 @@ Valid TYPEs:
 Example: --probe tp:skb:kfree_skb --probe kprobe:consume_skb"
     )]
     pub(super) probes: Vec<String>,
+    #[arg(short, long, help = "Write the events to a file for later processing.")]
+    pub(super) out: Option<String>,
```
amorenoz, line 40: clap supports options being `PathBuf`s directly which does some initial input validation when converting input.

src/collect/output.rs
```diff
@@ -0,0 +1,87 @@
...
+pub(super) struct Output<'a> {
+    /// Reference to the BPF events, mainly used to retrieve events.
+    bpf_events: &'a BpfEvents,
+    /// Should we output messages to stdout?
+    quiet: bool,
+    /// Writer to a file to store events for later processing, if any.
+    writer: Option<BufWriter<File>>,
+    /// Output file used in the above writer.
+    output_file: Option<File>,
+}
```
amorenoz, line 22: This output struct assumes the only way of printing events is in json (which is not very readable tbh).
If we aim at supporting various ways of printing data even on collection phase, how about we define a simple trait such as:
```rust
trait Processor {
    process_one(e: Event) -> Result<()>,
    close() -> Result<()>
}
```
Then this `FileOutput` would implement the trait (and only output to a file) but allow us to combine more than one simultaneous processor easily, some that come to mind at first `JsonPrint` (existing functionality), `PrettierThanJsonPrint`, `StatsPrint`,  `Spinner`, `SendToSock`, etc.

amorenoz: It's a good idea.
My only concern is that there might be `Processors` and `Write` that don't make much sense together, like a textual-like table and a UDP socket. On the other hand, could there be `Processor` implementation that need a bit more than just a `Write`? e.g: a TUI interface?

If so, we could have  something like:
```rust
trait Processor {
     process_one(e: Event) -> Result<()>,
}

trait Formatter {...} 
trait Write {...}

struct FormatAndWrite<F: Formatter,  W: Write>  {
     formatter: &F,
     writers: Vec<W>,
}
impl<F, W>Processor for  FormatAndWrite<F, W> {
    process_one(e: Event) -> Result<()> {
            let bytes = self.formatter.format(e);
            for w in self.writers { 
                   w.write(bytes);
            }
    }
}
```

That would leave space for all kind of processing. But maybe this is over-engineering something that's way simpler right now.

src/collect/output.rs
```diff
@@ -0,0 +1,103 @@
...
+    let mut writers: Vec<Box<dyn Write>> = Vec::new();
+
+    // Write the events to a file if asked to.
+    if let Some(out) = args.out.as_ref() {
+        writers.push(Box::new(BufWriter::new(
+            OpenOptions::new()
+                .create(true)
+                .write(true)
+                .truncate(true)
+                .open(out)?,
```
amorenoz, line 28: Maybe we could wrap the errors that can happen here and print the file that we're trying to open.
RIght now this is the first file we handle but when we have profiles, only printing "Not such file or directory" will be confusing.

PR #79, opened 2023-02-15
build.rs: do not always rebuild on non debian based systems.

vlrpl:
on Fedora this could stretch build time:

```
$ touch src/core/events/events.rs && /bin/time -p cargo build
   Compiling packet-tracer v0.1.0
    Finished dev [unoptimized + debuginfo] target(s) in 8.96s
real 9.17
user 5.99
sys 3.02
```

with this applied:

```
$ touch src/core/events/events.rs && /bin/time -p cargo build
   Compiling packet-tracer v0.1.0
    Finished dev [unoptimized + debuginfo] target(s) in 2.28s
real 2.31
user 1.85
sys 0.44
```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -2,6 +2,7 @@ use std::{
     env,
     fs::{create_dir_all, File},
     io::Write,
+    path::Path,
     process::Command,
 };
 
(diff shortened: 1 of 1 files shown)


vlrpl:
> If I understand this correctly, this is a fixup for the `/usr/include/x86_64-linux-gnu` header path, which is used to look for `errno.h`.
> 
> If the path doesn't exist (e.g. on Fedora), then do we use the right `errno.h` file?

Yes, the problem is that directory.
AFAICT, the problem is that for debian system that include path is needed as <asm/errno.h> fails to be included otherwise:

 ```
        stderr=
         In file included from src/core/probe/user/bpf/usdt.bpf.c:3:
        In file included from /tmp/.tmpOVQRLT/bpf/src/bpf/usdt.bpf.h:6:
        /usr/include/linux/errno.h:1:10: fatal error: 'asm/errno.h' file not found
        #include <asm/errno.h>
                 ^~~~~~~~~~~~~
        1 error generated.
```

in my Fedora <asm/errno.h> is under /usr/include, so we should still get it from the standard include paths.

PR #80, opened 2023-02-20
module: skb: bpf: fix event struct type

atenart:
Thanks to commit 09ddb6e9f1c8 spotted a type issue in the skb module hook.


amorenoz (approved):
LGTM

PR #75, opened 2023-02-03
Cli: Add version

vlrpl:
this small series includes the new -V option which shows the version in Cargo.tml and also adds the completion script for bash

edit: dropped the bash script. As discussed during the mtg we'll look into Clap's facilities first.

PR #87, opened 2023-02-23
Allow reporting the version in the cli

atenart:
Based on #75 with an extra commit.

Instead of only reporting the package version, also report the git HEAD information. This is useful as the Rust package version isn't directly linked to git information (tag, commit, etc) and the same version could be reported for different states of the project. This will help in supporting users to ensure the version points to a real state of the project.

The drawback is this information is retrieved in `build.rs` and the logic re-triggers the full `build.rs` for every `.git/HEAD` change. IMHO this is fine as we'll just force a rebuild when switching branches and/or adding new commits, but not for every development iterations.

Thoughts?

### DIFF
--- a/src/cli/cli.rs
+++ b/src/cli/cli.rs
@@ -231,8 +231,16 @@ impl ThinCli {
         I: IntoIterator<Item = T>,
         T: Into<OsString> + Clone,
     {
+        let version = if cfg!(debug_assertions) {
+            format!("{}-dbg", env!("CARGO_PKG_VERSION"))
+        } else {
+            env!("CARGO_PKG_VERSION").to_string()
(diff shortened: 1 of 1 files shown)


vlrpl:
> Retrieving the commit id doesn't work when packaging the application (it's not a git env), it turns out we might not be able to do this after all. Two options here:
> 
>     * Only include the first commit.
> 
>     * Or, the above and add a small indication if a build is a debug one in the reported version. (That's the current content of this PR).
> 
> 
> @vlrpl thoughts?

My preference is toward adding the small indication.

vlrpl (approved):
/lgtm
Thanks!

PR #102, opened 2023-03-22
Events rework

atenart:
Please have a look at this WIP & test; for initial comments and checking we're going in the right direction. Only the 3 last commits are part of this PR.

Main questions would be:
- Are the event traits defined properly?
- I changed a bit how collectors and unmarshalers are registered. My plan after the current state of this PR is to move the collector::Group to module and do the registration there for more clarity (but still keep the collector specific in `collector`).
- One thing to note is with the current implementation `None` values are converted to `null` in JSON (e.g. `"netns":null`). So all `None` values are present in the output. However I don't think we'll use JSON for stdout so we should have more control, and if needed we can use https://docs.rs/serde_with/latest/serde_with/attr.skip_serializing_none.html

Remarks:
- While both implementation are already possible using the current PR, I chose to use the same object for both the `EventSection` and `EventSectionFactory` traits in all modules (& core) in addition to  `#[serde(skip)]` for internal members. But we could use separate objects.
- I'm using derive to implement common traits for events and their factories. This is nice because we can add new traits with a default implementation w/o any module update. It's also required for `as_any` to work properly (it needs to be defined in the proper obj implementation).

Based on #67.

### DIFF
--- a/src/collect/collector.rs
+++ b/src/collect/collector.rs
@@ -13,7 +13,7 @@ use crate::core::probe::kernel::{config::init_stack_map, kernel::KernelEventFact
 use crate::{
     cli::{dynamic::DynamicCommand, CliConfig, FullCli},
     core::{
-        events::bpf::BpfEventsFactory,
+        events::{bpf::BpfEventsFactory, EventFactory},
         filters::{
             filters::{BpfFilter, Filter},
             packets::filter::FilterPacket,
@@ -63,12 +63,13 @@ pub(crate) struct Collectors {
     modules: Modules,
     probes: probe::ProbeManager,
     cli: CliConfig,
+    factory: Box<dyn EventFactory>,
     known_kernel_types: HashSet<String>,
 }
 
 impl Collectors {
     #[allow(unused_mut)] // For tests.
-    fn new(mut modules: Modules, mut cli: FullCli) -> Result<Self> {
+    fn new(mut modules: Modules, mut cli: FullCli, factory: Box<dyn EventFactory>) -> Result<Self> {
         // Register all collectors' command line arguments. Cli registration
         // errors are fatal.
         let cmd = cli.get_subcommand_mut()?.dynamic_mut().unwrap();
@@ -101,6 +102,7 @@ impl Collectors {
             probes,
             known_kernel_types: HashSet::new(),
             cli,
+            factory,
         })
     }
 
@@ -170,7 +172,12 @@ impl Collectors {
     /// Start the event retrieval for all collectors by calling
     /// their `start()` function.
     pub(crate) fn start(&mut self) -> Result<()> {
-        self.modules.start_factory()?;
+        let section_factories = match self.modules.section_factories.take() {
+            Some(factories) => factories,
+            None => bail!("No section factory found, aborting"),
+        };
+        self.factory.start(section_factories)?;
+
         self.probes.attach()?;
 
         self.modules.collectors().iter_mut().for_each(|(id, c)| {
@@ -215,11 +222,7 @@ impl Collectors {
         });
 
         loop {
-            match self
-                .modules
-                .factory
-                .next_event(Some(Duration::from_secs(1)))?
-            {
+            match self.factory.next_event(Some(Duration::from_secs(1)))? {
                 Some(event) => processors
                     .iter_mut()
                     .try_for_each(|p| p.process_one(&event))?,
@@ -282,7 +285,7 @@ pub(crate) fn get_collectors(cli: FullCli) -> Result<Collectors> {
     let factory = BpfEventsFactory::new()?;
     let event_map_fd = factory.map_fd();
 
-    let mut collectors = Collectors::new(get_modules(Box::new(factory))?, cli)?;
+    let mut collectors = Collectors::new(get_modules()?, cli, Box::new(factory))?;
     collectors.probes.reuse_map("events_map", event_map_fd)?;
 
     Ok(collectors)
@@ -358,12 +361,12 @@ mod tests {
     }
 
     fn new_collectors(modules: Modules) -> Result<Collectors> {
-        Collectors::new(modules, get_cli()?)
+        Collectors::new(modules, get_cli()?, Box::new(BpfEventsFactory::new()?))
     }
 
     #[test]
     fn register_collectors() -> Result<()> {
-        let mut group = Modules::new(Box::new(BpfEventsFactory::new()?))?;
+        let mut group = Modules::new()?;
         assert!(group
             .register(
                 ModuleId::Skb,
@@ -383,7 +386,7 @@ mod tests {
 
     #[test]
     fn register_uniqueness() -> Result<()> {
-        let mut group = Modules::new(Box::new(BpfEventsFactory::new()?))?;
+        let mut group = Modules::new()?;
         assert!(group
             .register(
                 ModuleId::Skb,
@@ -409,7 +412,7 @@ mod tests {
 
     #[test]
     fn init_collectors() -> Result<()> {
-        let mut group = Modules::new(Box::new(BpfEventsFactory::new()?))?;
+        let mut group = Modules::new()?;
         let mut dummy_a = Box::new(DummyCollectorA::new()?);
         let mut dummy_b = Box::new(DummyCollectorB::new()?);
 
@@ -437,7 +440,7 @@ mod tests {
 
     #[test]
     fn start_collectors() -> Result<()> {
-        let mut group = Modules::new(Box::new(BpfEventsFactory::new()?))?;
+        let mut group = Modules::new()?;
         let mut dummy_a = Box::new(DummyCollectorA::new()?);
         let mut dummy_b = Box::new(DummyCollectorB::new()?);
 
@@ -462,7 +465,7 @@ mod tests {
 
     #[test]
     fn parse_probe() -> Result<()> {
-        let mut group = Modules::new(Box::new(BpfEventsFactory::new()?))?;
+        let mut group = Modules::new()?;
         group.register(
             ModuleId::Skb,
             Box::new(DummyCollectorA::new()?),
--- a/src/module/module.rs
+++ b/src/module/module.rs
@@ -116,8 +116,6 @@ where
 pub(crate) struct Modules {
     /// Set of registered modules we can use.
     modules: HashMap<ModuleId, Box<dyn Module>>,
-    /// Factory used to retrieve events.
-    pub(crate) factory: Box<dyn EventFactory>,
     /// Event section factories to parse sections into an event. Section
     /// factories come from modules. They are under an Option to allow ownership
     /// change so they can be consumed by the event factory (and moved to a
@@ -126,7 +124,7 @@ pub(crate) struct Modules {
 }
 
 impl Modules {
-    pub(crate) fn new(factory: Box<dyn EventFactory>) -> Result<Modules> {
+    pub(crate) fn new() -> Result<Modules> {
         let mut section_factories: HashMap<ModuleId, Box<dyn EventSectionFactory>> = HashMap::new();
 
         // Register core event sections.
@@ -136,7 +134,6 @@ impl Modules {
 
         Ok(Modules {
             modules: HashMap::new(),
-            factory,
             section_factories: Some(section_factories),
         })
     }
@@ -175,16 +172,6 @@ impl Modules {
         Ok(self)
     }
 
-    /// Start the event retrieval in the factory.
-    pub(crate) fn start_factory(&mut self) -> Result<()> {
-        let section_factories = match self.section_factories.take() {
-            Some(factories) => factories,
-            None => bail!("No section factory found, aborting"),
-        };
-
-        self.factory.start(section_factories)
-    }
-
     /// Get an hashmap of all the collectors available in the registered
     /// modules.
     pub(crate) fn collectors(&mut self) -> HashMap<&ModuleId, &mut dyn Collector> {
@@ -219,8 +206,8 @@ impl Modules {
     }
 }
 
-pub(crate) fn get_modules(factory: Box<dyn EventFactory>) -> Result<Modules> {
-    let mut group = Modules::new(factory)?;
+pub(crate) fn get_modules() -> Result<Modules> {
+    let mut group = Modules::new()?;
 
     // Register all collectors here.
     group
@@ -250,11 +237,8 @@ pub(crate) fn get_modules(factory: Box<dyn EventFactory>) -> Result<Modules> {
 
 #[cfg(test)]
 mod tests {
-    use crate::core::events::bpf::BpfEventsFactory;
-
     #[test]
     fn get_modules() {
-        let factory = BpfEventsFactory::new().unwrap();
-        assert!(super::get_modules(Box::new(factory)).is_ok());
+        assert!(super::get_modules().is_ok());
     }
 }


src/module/group.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// factories come from modules.
+    pub(crate) section_factories: Option<HashMap<ModuleId, Box<dyn EventSectionFactory>>>,
+}
+
+impl Group {
+    pub(crate) fn new(factory: Box<dyn EventFactory>) -> Result<Group> {
+        let mut section_factories: HashMap<ModuleId, Box<dyn EventSectionFactory>> = HashMap::new();
+
+        section_factories.insert(ModuleId::Common, Box::<CommonEvent>::default());
+        // FIXME: provide stack map in a way
```
amorenoz, line 36: This is already done in `collectors.rs`

amorenoz (commented):
I've just started going through this PR which is big so it'll take me some time.
In general I love the direction this is going. Thanks for this work which is not easy at all :-D

src/process/process.rs
```diff
@@ -0,0 +1,52 @@
...
+
+        let factory = FileEventsFactory::new(&config.file);
+        let mut modules = get_modules(Box::new(factory))?;
+
+        modules.start_factory()?;
+
+        // Read all events from the factory.
+        let mut events = Vec::new();
+        while let Some(event) = modules.factory.next_event(None)? {
+            events.push(event);
```
amorenoz, line 33: I don't think putting all the events in memory is a good idea. It's likely that, for many usecases the size of this will be very big.

amorenoz (commented):
Having read the rest of the series I'll try to summarize what feels to me a bit wrong about `Group`.
- Firstly, the name suggests that it implements some kind of Structural pattern but it really contains heterogeneous things and seems intended to provide at least 3 heterogeneous functionalities:
    -  `Collector`  registration
    -  Grouping the `EventSectionFactories` that are incrementally added while dynamic registration of modules and finally combine them and feed them to the `EventFactory`.
    - Wrap `EventFactory`'s next_event() (although not currently done using a public func) 
- The temptation of using it both in the event collection and postprocessing tells me we see some parallelism between both behaviors that we want to refactor out of  `collectors.rs`  and `process.rs`.

So for me it makes sense to think about what we want `Group` for:

A ) An homogeneous group of modules. If so, I'd consider removing the factory and defining a wider registration API that contains both collector and pos-processing functionality:

```rust
/// Encapsulate all the functionality provided by a module (more extensible)
struct Module {
    collector: Option<Box<dyn Collector>>,
    factory: Option<Box<dyn EventSectionFactory>>,
    pos-process: Option<Box<dyn PostProcessor>>, // TBD
}
```
This way the registration API is (and will always be) like this:
```rust
impl Group  {
   pub(crate) fn  (id: ModuleID, module: Module) {}
}
```
(not sure this is very useful)

B) A wrapper around EventFactory that helps with the dynamic registration of `EventSectionFactories`. In such case, I'd remove the `module` i.e: the `Collector`s (`Collector` registration stays in `collectors.rs` same as `PostProcessor` registration stays in `process.rs`). Call it something like `DynamicFactory` (not a very good name either I admit) and define an API such as:
```rust
impl DynamicFactory {
    /// Creates a new DyamicFactory 
    pub fn new(factory: Box<dyn EventFactory>) -> DynamicFactory {}
    /// Add a new EventFactory
    pub fn add_factory(id: ModuleId, Box<dyn EventSectionFactory>) -> Result() {}
    /// When event registration has ended, start the factory
    pub fn start() -> Result<()> {}
    /// Get an event from the factory (it must have been started before)
    pub fn next_event() -> Result<Event> {}
}
```

C) (Extension / variation of B): Define a common way of processing events. In which case, again remove `Collector`s, and move the ` loop { factory.next_event() } `  inside allowing for external processing functions:
```rust
impl EventLoop  {
    // Could contain all of the above funcs from DynamicFactory or a DynamicFactory as member.
    fn loop(proc: F, exit: mpsc::Receiver) -> Result<Event()> 
       where F: Fn(&Event) -> Result<()>
```
This might seem as adding a little but I foresee the event loop might get more complicated as we try to increase performance (e.g: by adding multi-threaded processing)

Notes:
- In option C) I tried to move away from naming it something`Processor` because we already have a nomenclature difficulty to distinguish between 2 existing "processors": The output processor and post-processing, so we might want to look for better names.

amorenoz:
> Having read the rest of the series I'll try to summarize what feels to me a bit wrong about `Group`.

Clearly failed to summarize :-D.

src/module/skb/skb.rs
```diff
@@ -37,26 +30,17 @@ impl Collector for SkbCollector {
...
-        SKB_COLLECTOR
-    }
-
     fn known_kernel_types(&self) -> Option<Vec<&'static str>> {
         Some(vec!["struct sk_buff *"])
     }
 
     fn register_cli(&self, cmd: &mut DynamicCommand) -> Result<()> {
-        cmd.register_module::<SkbCollectorArgs>(SKB_COLLECTOR)
+        cmd.register_module::<SkbCollectorArgs>(ModuleId::Skb.to_str())
```
amorenoz, line 38: I think we can make `DynamicCommand` accept a `ModuleId` directly instead of a str

src/module/group.rs
```diff
@@ -0,0 +1,146 @@
...
+        )?
+        .register(
+            ModuleId::Skb,
+            Box::new(SkbCollector::new()?),
+            Box::<SkbEvent>::default(),
+        )?
+        .register(
+            ModuleId::Ovs,
+            Box::new(OvsCollector::new()?),
+            Box::<OvsEvent>::default(),
```
amorenoz, line 131: Just double-checking here: We are not forcing the same struct to implement `EventType` and `EventSectionFactory` right?

src/core/events/events.rs
```diff
@@ -32,156 +32,134 @@
...
+pub(crate) type SectionFactories = HashMap<ModuleId, Box<dyn EventSectionFactory>>;
+
+/// Implemented by objects generating events from a given source (BPF, file,
+/// etc).
+pub(crate) trait EventFactory {
+    /// Starts the factory events collection.
+    fn start(
+        &mut self,
+        section_factories: HashMap<ModuleId, Box<dyn EventSectionFactory>>,
+    ) -> Result<()>;
```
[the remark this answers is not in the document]
amorenoz: I didn't notice many of our traits are like this. I guess it's ok although not the easiest format for reading.

src/core/probe/kernel/kernel.rs
```diff
@@ -62,20 +56,30 @@ impl fmt::Display for KernelProbe {
...
+    /// Kernel symbol name associated with the event (i.e. which probe generated
+    /// the event).
+    pub(crate) symbol: String,
+    /// Probe type: one of "kprobe", "kretprobe" or "raw_tracepoint".
+    pub(crate) probe_type: String,
+    pub(crate) stack_trace: Option<Vec<String>>,
     #[cfg(not(test))]
-    stack_map: libbpf_rs::Map,
+    #[serde(skip)]
+    pub(crate) stack_map: Option<libbpf_rs::Map>,
```
amorenoz, line 69: Having looked briefly at script pos-processing I'm pretty sure we will end up having to split the `EventSection` and  `EventSectionFactory` objects. Even more if they contain non-clonable objects such as libbpf_rs::Maps.

Do you think we should start by splitting them so we set the "preferred way" of doing things from the very beginning?

src/core/events/bpf.rs
```diff
@@ -271,31 +202,19 @@ fn parse_raw_event(data: &[u8], unmarshalers: &mut Unmarshalers) -> Result<Event
...
+            Some(factory) => factory,
+            None => bail!("Unknown factory for event section owner {}", owner),
+        };
 
-        // Fill the event with unmarshaled sections. Unwrap as we know
-        // it's a valid owner.
-        for field in fields {
-            event.insert(owner, field);
-        }
+        event.insert_section(*owner, factory.from_raw(sections.clone())?)?;
```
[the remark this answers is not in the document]
amorenoz: It seems this hasn't changed. The idea was to avoid the `clone()`. Either `pop` or `drain` would to the job

src/module/ovs/event.rs
```diff
@@ -0,0 +1,68 @@
...
+use serde::{Deserialize, Serialize};
+
+use super::bpf::*;
+use crate::{
+    core::events::{bpf::BpfRawSection, *},
+    EventSection, EventSectionFactory,
+};
+
+#[derive(Default, Deserialize, Serialize, EventSection, EventSectionFactory)]
+pub(crate) struct OvsEvent {
```
amorenoz, line 11: Flattening all of the sub-events into one big one makes it quite difficult to read and maintain. Even leading to repeated fields. This event is really semantically an enum, so I believe we can implement it as one using enum tagging, eg: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e1fd08fd86e7db9004630942c4b84a14

I can give it a try and send a patch that implements this proposal. It can then be squashed into this one, or just added at the end. WDYT?

amorenoz: The rust struct that best represents the OvsEvent is:

```rust

#[derive(Deserialize, Serialize)]
#[serde(tag = "action")]
enum ActionData {
   Output(u32),
   Drop,
   Recirc(u32),
...
}

#[derive(Deserialize, Serialize)]
struct Action {
   action: ActionData,
   recirc_id: u32
}

#[derive(Deserialize, Serialize)]
#[serde(tag = "event_type")]
enum OvsEventData {
    #[serde(rename = "upcall")]
    UpcallEvent(UpcallData),
    #[serde(rename = "action_execute")]
    ActionEvent(Action),
    #[serde(rename = "recv_upcall")]
    RecvUpcallEvent(RecvUpcallData),
    #[serde(rename = "flow_operation")]
    OvsOperationEvent(OperationData),
    #[serde(rename = "upcall_enqueue")]
    UpcallEnqueueEvent(UpcallEnqueueData),
    #[serde(rename = "event_type")]
    UpcallReturnEvent(UpcallReturnData),
}

```
In fact, this probably allows us to reuse most of the structs in `bpf.rs`.

amorenoz: OK. Will send a follow up patch.

src/collect/collector.rs
```diff
@@ -66,19 +71,39 @@ pub(crate) trait Collector {
...
+        probes.reuse_map("stack_map", sm.fd())?;
+
+        let kernel_event = KernelEvent {
+            #[cfg(not(test))]
+            stack_map: Some(sm),
+            ..Default::default()
+        };
+
+        section_factories.insert(ModuleId::Common, Box::<CommonEvent>::default());
+        section_factories.insert(ModuleId::Kernel, Box::new(kernel_event));
```
amorenoz: Missing UserEvent?

src/module/skb/event.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub(crate) etype: Option<u16>,
+    /// Source MAC address.
+    pub(crate) src: Option<String>,
+    /// Destination MAC address.
+    pub(crate) dst: Option<String>,
+    // IPv4 & IPv6 fields
+    /// Source IP address.
+    pub(crate) saddr: Option<String>,
+    /// Destination IP address.
+    pub(crate) daddr: Option<String>,
```
amorenoz, line 23: Looking at events generated by this, it seems we need to add `#[serde(skip_serializing_if = "Option::is_none")]` or else the json event has a lot of `"whatever": null`

amorenoz: Another idea to catch these ones (not necessarily to be added to this PR) is to have some unit tests that test event <-> json conversion.

amorenoz: > > One thing to note is with the current implementation None values are converted to null in JSON (e.g. "netns":null). So all None values are present in the output. However I don't think we'll use JSON for stdout so we should have more control, and if needed we can use https://docs.rs/serde_with/latest/serde_with/attr.skip_serializing_none.html
> 
> With this we only should have to define a single attribute for the whole section. However I'm wondering if we need this right now, as I don't know if we'll use json for stdout. If json is only used to store data then I guess it's acceptable to have null values. WDYT?

I think it'll increase the size of the file unnecessarily.

src/module/ovs/event.rs
amorenoz: I've sent #112 based on this. Do you want to merge both PRs into one?

src/core/probe/kernel/kernel.rs
amorenoz: I have been playing with this and have another alternative:

It's true that associated types cannot be boxed.... but we can can create a binding trait to circumvent that:

Summary:
```rust
pub(crate) trait EventSectionFactory: RawEventSectionFactory, SerdeEventSectionFactory {} // None of these must have associated types

// As promised: no associated types
pub(crate) trait SerdeEventSectionFactory {
    fn from_json(&self, val: serde_json::Value) -> Result<Box<dyn EventSection>>;
}

// This trait does have associated types!
pub(crate) trait EventSectionBinding {
    // The associated section type.
    type Event: for<'a> serde::Deserialize<'a> + EventSection + 'static;

   // A default implementation of the from_json function which uses the fact that Event implements Deserialize.
    fn from_json(&self, val: serde_json::Value) -> Result<Box<dyn EventSection>> {
        Ok(Box::new(serde_json::from_value::<Self::Event>(val)?))
    }
}

// Implement SerdeEventSectionFactory for every type that implements EventSectionBinding
impl<A, B: EventSectionBinding<Event = A>> SerdeEventSectionFactory for B {
    fn from_json(&self, val: serde_json::Value) -> Result<Box<dyn EventSection>> {
        // here Self is treated as  EventSectionBinding
        self.from_json(val)
    }
}
```

Full implementation: https://github.com/amorenoz/retis/commit/39b747ace740d76111378f48844af751141fb6b7

WDYT?

src/module/skb/skb.rs
```diff
@@ -68,6 +73,23 @@ impl Collector for SkbCollector {
...
 
+        // Special case the "drop_reason" section retrieval and automatically
+        // add a kprobe to the right function as it's the ~only way to get it.
+        if sections & 1 << SECTION_DROP_REASON != 0 {
+            let symbol = Symbol::from_name("kfree_skb_reason");
+            // Did the probe failed because of an error or because it wasn't
+            // available? In case we can't know, do not issue an error.
+            match symbol.is_ok() {
+                true => {
+                    // Unwrap as we just checked the value is valid.
```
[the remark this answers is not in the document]
amorenoz: What I mean is the following feels more "rust-onic":
```rust
match Symbol::from_name("kfree_skb_reason") {
    Ok(symbol) => {
       if let Err(e) = probes.add_probe(Probe::kprobe(symbol)?) {
           error!("Could not attach to kfree_skb_reason: {}", e);
       }
    }
    Err(_) => info!("Skb drop reasons are not retrievable on this kernel"),
}
```
It does not need an `unwrap()` (and the comment warning about it).

src/module/skb/skb.rs
```diff
@@ -68,6 +73,23 @@ impl Collector for SkbCollector {
...
+                    if let Err(e) = probes.add_probe(Probe::kprobe(symbol.unwrap())?) {
```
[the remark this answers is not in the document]
amorenoz: I get your point.
My thinking is: "concrete" collectors (e.g: ovs) add their own probes. "generic" collectors (skb) allow users to provide the probes they want. So it's either all or nothing: Either I (user) decide what to probe, or I select a collector that does it for me.

So for me it makes more sense for a "concrete" collector (e.g "skb-drop") to add this probe.

src/module/skb_tracking/event.rs
```diff
@@ -29,13 +32,9 @@ pub(crate) struct SkbTrackingEvent {
...
 
 impl RawEventSectionFactory for SkbTrackingEvent {
-    fn from_raw(&mut self, mut raw_sections: Vec<BpfRawSection>) -> Result<Box<dyn EventSection>> {
-        if raw_sections.len() != 1 {
-            bail!("Skb tracking event from BPF must be a single section");
-        }
-
-        // Unwrap as we just checked the vector contains 1 element.
-        let raw = parse_raw_section::<BpfTrackingEvent>(&raw_sections.pop().unwrap())?;
+    fn from_raw(&mut self, raw_sections: Vec<BpfRawSection>) -> Result<Box<dyn EventSection>> {
```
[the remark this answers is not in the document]
amorenoz: OK, not for `KernelEvent` and `UserspaceEvent`, although I'd argue that we should define a struct and use Plain even if it's a temporary one and we then need to do some transformations instead of accessing the fields by offset which I consider really ugly and hard to maintain. But CommonEvent can benefit from this directly

src/module/skb/skb.rs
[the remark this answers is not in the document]
amorenoz: > The above would work the same in the two solutions we're discussing, as the skb module alone does not insert any extra probe (except if drop_reason is asked to be retrieved).

The skb module alone (with default values) does insert drop_reason right?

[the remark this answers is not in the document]
amorenoz: > I'm worrying about the user experience and this being easy to use. If one wants to trace skbs it makes sense to use the skb module. On the other hand the drop reason isn't really an skb attribute, so a dedicated module could make sense :) That might even leave the door open to track other drop reasons (not provided by the current mechanism).

Let's explore the user-experience:
A) drop reason is selected by skb-section
B) drop reason is moved to an independent collector

Cases:
1) I want just drop reasons (dropwatch-mode):
    A) retis collect -c skb-drop
    B) retis collect -c skb --skb-sections=drop_reason

2) I want to attach a bunch of probes in the tcp stack and I also want drops. I exactly all the skb sections I want
    A) retis collect -c skb --skb-sections={SECTIONS},drop_reason -p probe1,probe2...
    B) retis collect -c skb,skb-drop --skb-sections={SECTIONS} -p probe1, probe2...

3) I want to attach a bunch of probes in the tcp stack and I also want drops. I don't really know what skb sections are (I hope default values are sane enough)
    A) retis collect -c skb -p probe1,probe2...
    B) retis collect -c skb,skb-drop -p probe1, probe2...
    
4) I want to attach a bunch of probes in the tcp stack and I _dont_ want drops. I don't really know what skb sections are (I hope default values are sane enough)
    A) retis collect -c skb -skb-sections={SECTIONS} -p probe1,probe2... # I cannot use the defaults so I have to take all the defaults and remove drop-reason
    B) retis collect -c skb,skb-drop -p probe1, probe2...

5) I want to inspect the ovs datapath (with some reasonable skb data). I don't want all the drops in the TCP stack!
   A) retis collect -c ovs,skb --skb-sections={SECTIONS} --ovs-track. # I cannot use the defaults so I have to take all the defaults and remove drop-reason
    B) retis collect -c ovs,skb-drop --ovs-track
    
5) I want to inspect the ovs datapath (with some reasonable skb data). I also want all the drops all over the kernel
   A) retis collect -c ovs,skb --ovs-track.
    B) retis collect -c ovs,skb-drop --ovs-track

So, what I see here is that this forces the user to know about the skb defaults in some cases while not really hindering user experience in the rest.
WDYT?

src/module/skb_tracking/event.rs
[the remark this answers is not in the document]
amorenoz: 100% we should not do that in this PR :-D

src/module/skb/skb.rs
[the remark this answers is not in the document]
amorenoz: BTW, I'm not saying we should move drop_reason to their own module in this PR, I'm just thinking where it makes more sense for it to go

src/core/probe/kernel/kernel.rs
amorenoz: Nice!

src/collect/collector.rs
```diff
@@ -137,22 +150,22 @@ impl Collectors {
...
-
-        self.setup_filters(collect)?;
+        let mut probes = Vec::new();
+        collect.args()?.probes.iter().try_for_each(|p| {
+            probes.push(self.parse_probe(p)?);
+            Ok::<(), anyhow::Error>(())
+        })?;
+        probes
+            .drain(..)
+            .try_for_each(|p| self.probes.add_probe(p))?;
```
amorenoz, line 162: Adding cli as member of `Collectors` forces us to do this two-step probe adding.

Considering my previous experiment, would you consider this other approach?
https://github.com/amorenoz/retis/commit/5cd386e8c0a8fdd5da2490f0e63535ff353b2e95

TL;DR: Instead of passing cli in `Collectors::new()`, pass it in a new `Collectors::run()` that does all the processing. That way, we still own the cli but it's not a member of `Collectors`.

Note: I'm not suggesting adding it to this PR, I'm just considering this direction. If you think it's ok I'll fix the patch and send it.

amorenoz: > I'm wondering if we could avoid using the cli obj as an argument there though, it seems quite complex to take a subcommand out of cli and then use cli as an argument for all the runner helpers. There might be a way to make all this more self-contained, but honestly haven't looked at this so this might be OK this way. (The above makes sense regardless of this).

That's precisely why I ended up having to split the Runner from the SubCommand itself.

amorenoz (approved):
From my point of view, this PR is in a good shape

PR #108, opened 2023-03-30
[RFC] Unified api

atenart:
Following our discussion on how to unify and consolidate the API, here are 3 patches based on top of #102 to start the discussion. The design for this initial work is as follow:

- `core::*` APIs were unified into a `core::Retis` object.
- We keep 3 high level objects: `core::Retis`, `cli::*` and `module::Modules`. It makes sense as those represent the 3 main elements we are dealing with and their separation helps in multiple ways:
  - They can't be grouped in Rust: we need to iterate over modules or over cli arguments while using the main API object (`Retis`).
  - Depending on the consumer, only a combination of those 3 objects is required.
- The end consumer is responsible for bundling the 3 above main objects. E.g. the collector command.
- About the collector, one possibility is to provide a command at some point to collect and post-process at the same time. For this to work we'll need to split the "collector" command part from the "collector" obj using the trait. I don't see this as a blocking point and would prefer to do this only once needed, but feel free to comment.

### DIFF
--- a/src/collect/collector.rs
+++ b/src/collect/collector.rs
@@ -65,12 +65,12 @@ pub(crate) struct Collectors {
 }
 
 impl Collectors {
-    pub(crate) fn new(modules: Modules, retis: Retis, mut cli: FullCli) -> Result<Self> {
+    pub(crate) fn new(mut modules: Modules, retis: Retis, mut cli: FullCli) -> Result<Self> {
         // Register all collectors' command line arguments. Cli registration
         // errors are fatal.
         let cmd = cli.get_subcommand_mut()?.dynamic_mut().unwrap();
         modules
-            .modules
+            .collectors()
             .iter()
             .try_for_each(|(_, c)| c.register_cli(cmd))?;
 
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -26,19 +35,30 @@ fn main() -> Result<()> {
         ColorChoice::Auto,
     )?;
 
+    // Step 3: get the modules.
+    let mut modules = get_modules()?;
+
+    // Step 4: dispatch the command.
     let command = cli.get_subcommand_mut()?;
     match command.name() {
         "collect" => {
```
amorenoz, line 44: Inspired by the great improvements of this RFC and by the ability to use associated types on Boxed traits that I learnt on https://github.com/retis-org/retis/pull/102#discussion_r1165355400 I have added a potential improvement on top of this:
https://github.com/amorenoz/retis/commit/1b6eca3decb16858a45d944235249d0f0a7608c5

It allows us to move 100% of the command logic inside the command itself (in this case `collect/collect.rs`). I think we will appreciate it once we start adding more and more commands to the cli. WDYT?

amorenoz (commented):
In general, having a unified API seems an interesting idea that we should explore. However, right now `retis::Retis` is "just" encapsulating a `EventFactory` (which actually has to be constructed externally, and a `ProbeManager`, which in fact, is only used on certain cases (only on `retis collect`).
Do you think we'll add more things to it?
If the goal is to unify, why not put `modules` inside as well?

PR #109, opened 2023-03-31
Add a contributing guide

### LINKED ISSUE #48: Add a contributing doc
Could be in `CONTRIBUTING.md` and should contain pointers on how to contribute, what to check before the CI does, etc. Another aspect would be to write a small example (with explanations) on how to write a collector.

Please drop below raw information that should be part of it.

atenart:
Fixes #48.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,124 @@
+# Contribution guidelines
+
+Thank you for investing your time and considering contributing to Retis! :tada:
+
+This guide documents the process to get changes merged into this project, what
+one should expect, as well as prerequisites for pull-requests and patches to be
+considered acceptable.
+
+### Table of contents
+
+* [Reporting bugs](#reporting-bugs)
+* [Suggesting new features](#suggesting-new-features)
+* [Code contribution](#code-contribution)
+
+## Reporting bugs
+
+Following the guidelines will help the community understand bug reports, which
+in turns will help in reproducing, triaging and fixing those bugs.
+
+### Before submitting the report
+
+1. Check for similar opened or closed issues. If a similar issue is still
+   opened, please add a comment there.
+1. Check the version used when the bug was triggered. If it was not the latest,
+   please reproduce with it.
+
+### Submitting a good bug report
+
+1. Use a clear, concise and descriptive title.
+1. Describe how to reproduce the bug, in clear exact steps, with examples. The
+   community should not have to guess extra steps, configuration, prerequisites
+   and spend lot of time gathering this information to work on fixing a bug.
+1. How reliable are those steps to reproduce the issue?
+1. Describe the outcome of those steps.
+1. Then describe what would have been expected and why.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+coding style for the BPF parts.
+
+### Opening a pull-request
+
+1. Before opening the pull-request, test the changes. Yes, even for those simple
+   one-liner last minute changes.
+1. Check the following commands do not return an error:
+   1. `cargo fmt --check`
+   1. `cargo clippy -- -D warnings`
+   1. `cargo test`, or to include runtime tests,
```
vlrpl, line 101: I'm not entirely sure, but 

`cirrus run --container-backend=podman`

can be something to consider to add as an alternative.
`--features=test_cap_bpf` still need to be tested locally

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+### Submitting a good bug report
+
+1. Use a clear, concise and descriptive title.
+1. Describe how to reproduce the bug, in clear exact steps, with examples. The
+   community should not have to guess extra steps, configuration, prerequisites
+   and spend lot of time gathering this information to work on fixing a bug.
+1. How reliable are those steps to reproduce the issue?
+1. Describe the outcome of those steps.
+1. Then describe what would have been expected and why.
+1. Don't make vague reports. If it is performances related, include numbers.
```
vlrpl, line 36: I would split the last item. The "performance numbers" seems important enough to stay on its own.

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+   documentation we expect.
+1. Add unit tests.
+1. Base new work on top of the `main` branch.
+
+### Coding style
+
+We strictly follow the Rust coding style as enforced by `rustfmt` and have a
+strong preference for the
+[Linux kernel](https://www.kernel.org/doc/html/latest/process/coding-style.html)
+[networking](https://www.kernel.org/doc/html/latest/process/maintainer-netdev.html#multi-line-comments)
```
vlrpl, line 91: these would render as two separate links for "Linux kernel" and "networking", and it's a bit hidden because of that. It seems better to explicitly mention the netdev guidelines.
WDYT?

CONTRIBUTING.md
vlrpl: that's why I wasn't really sure about it.
Let's just keep it as it is.

CONTRIBUTING.md
vlrpl: ok, I misinterpreted the intent then :)
What about just going with:
"If it is performances related, include numbers."

CONTRIBUTING.md
vlrpl: works for me

vlrpl (approved):
lgtm

PR #110, opened 2023-03-31
core: user: proc: fix get_threads test

atenart:
The core::user::proc::tests::get_threads test might fail sometime due to timing issues, if the thread isn't created fast enough. Fix this by ensuring the thread was started before running the test statements.

To avoid having 2 channels, use a simple Arc<Mutex<>> construction. This should be OK for test purposes.

### DIFF
--- a/src/core/user/proc.rs
+++ b/src/core/user/proc.rs
@@ -579,7 +579,10 @@ fn get_thread_info(pid: i32) -> Result<Vec<ThreadInfo>> {
 
 #[cfg(test)]
 mod tests {
-    use std::{sync::mpsc, thread, time::Duration};
+    use std::{
+        sync::{Arc, Barrier},
+        thread,
+    };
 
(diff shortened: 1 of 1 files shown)


src/core/user/proc.rs
```diff
@@ -669,16 +673,30 @@ mod tests {
 
     #[test]
     fn get_threads() -> Result<()> {
-        let (tx, rx) = mpsc::channel();
+        let cnt = Arc::new(Mutex::new(0));
```
vlrpl, line 676: I would suggest using std::sync::Barrier (waiting from both).
Optional nit: using join() would be nice as well

vlrpl: no, I don't see any. Feel free to ignore the optional one.

vlrpl (approved):
LGTM, thanks!

PR #118, opened 2023-04-19
Move skb tracking into the core & use it for filtering & reworked the probe manager

atenart:
Move the skb tracking logic into the core and use it for filtering. Not a lot of additions, except for the filtering part. Otherwise it's mostly code moving, which is intended for now. We can add Rust helpers later if needed by some modules.

Using the skb tracking logic in the filtering part has two main advantages:

- We are not flooded anymore by tracking events unrelated to the current filter.
- We can track packets even after they are modified. Example:

```
[...]
# ./retis collect -f 'udp port 53 and src host 10.0.42.2' -c skb,skb-drop --skb-sections all \
	-p kprobe:udp_rcv -p kprobe:ip_rcv -p kprobe:ip_finish_output2 -p kprobe:napi_gro_receive \
	-p kprobe:inet_gro_receive -p kprobe:udp_gro_receive &
# ip netns exec ns1 dig @1.1.1.1 redhat.com
{"common":{"timestamp":610794283209},"kernel":{"probe_type":"kprobe","symbol":"ip_rcv"},"skb":{"cloned":true,"daddr":"1.1.1.1","dataref":1,"dev_name":"veth01","dport":53,"dst":"1e:9c:f0:29:be:69","etype":2048,"fclone":false,"ifindex":19,"ip_version":4,"l3_len":79,"netns":4026531840,"protocol":17,"rx_ifindex":19,"saddr":"10.0.42.2","sport":45070,"src":"62:13:b1:87:e1:b0","udp_len":59,"users":1},"skb-tracking":{"orig_head":18446625427471709696,"skb":18446625427490994944,"timestamp":610794283209}}
{"common":{"timestamp":610794304627},"kernel":{"probe_type":"kprobe","symbol":"pskb_expand_head"},"skb":{"cloned":true,"daddr":"1.1.1.1","dataref":1,"dev_name":"veth01","dport":53,"dst":"1e:9c:f0:29:be:69","etype":2048,"fclone":false,"ifindex":19,"ip_version":4,"l3_len":79,"netns":4026531840,"protocol":17,"rx_ifindex":19,"saddr":"10.0.42.2","sport":45070,"src":"62:13:b1:87:e1:b0","udp_len":59,"users":1},"skb-tracking":{"orig_head":18446625427471709696,"skb":18446625427490994944,"timestamp":610794283209}}
{"common":{"timestamp":610794309230},"kernel":{"probe_type":"kprobe","symbol":"skb_free_head"},"skb":{"cloned":true,"daddr":"1.1.1.1","dataref":1,"dev_name":"veth01","dport":53,"dst":"1e:9c:f0:29:be:69","etype":2048,"fclone":false,"ifindex":19,"ip_version":4,"l3_len":79,"netns":4026531840,"protocol":17,"rx_ifindex":19,"saddr":"10.0.42.2","sport":45070,"src":"62:13:b1:87:e1:b0","udp_len":59,"users":1},"skb-tracking":{"orig_head":18446625427471709696,"skb":18446625427490994944,"timestamp":610794283209}}
{"common":{"timestamp":610794328206},"kernel":{"probe_type":"kprobe","symbol":"ip_finish_output2"},"skb":{"cloned":false,"daddr":"1.1.1.1","dataref":1,"dev_name":"eth0","dport":53,"dst":"1e:9c:f0:29:be:69","etype":2048,"fclone":false,"ifindex":2,"ip_version":4,"l3_len":79,"netns":4026531840,"protocol":17,"rx_ifindex":19,"saddr":"172.16.42.90","sport":45070,"src":"62:13:b1:87:e1:b0","udp_len":59,"users":1},"skb-tracking":{"orig_head":18446625427471709696,"skb":18446625427490994944,"timestamp":610794283209}}
{"common":{"timestamp":610794731861},"kernel":{"probe_type":"kprobe","symbol":"skb_free_head"},"skb":{"cloned":false,"daddr":"1.1.1.1","dataref":1,"dev_name":"eth0","dport":53,"dst":"52:54:00:bc:4e:4a","etype":2048,"fclone":false,"ifindex":2,"ip_version":4,"l3_len":79,"netns":4026531840,"protocol":17,"rx_ifindex":19,"saddr":"172.16.42.90","sport":45070,"src":"52:54:00:0b:99:9e","udp_len":59,"users":1},"skb-tracking":{"orig_head":18446625427471709696,"skb":18446625427490994944,"timestamp":610794283209}}
```

`saddr` does not match the filter in the last two events, because the packet was NATed.

Based on #102.


src/collect/collector.rs
```diff
@@ -151,6 +152,10 @@ impl Collectors {
             }
         }
 
+        // Initialize tracking & filters.
+        init_tracking(&mut self.probes)?;
```
amorenoz, line 156: We're changing the default behavior here. Now we're attaching the skb-tracking-related probes.
I'm not saying it's necessarily bad, just pointing it out. Do you see any reason why this would not be wanted? do we need a flag?

amorenoz (commented):
Maybe skb_tracking.rs can be now renamed to skb.rs?

amorenoz:
While testing this in combination with OVS module I found a slightly weird behavior:

Use case:
I want to debug some OVS issue so I prepare my test and before running it I type:
`retis collect -c ovs -f "tcp"` what should  I expect?

What I expect:
Only ovs module enabled so  I expect only ovs-related events. Also, I  only want TCP packets.

What I see:
Hundreds of events from all the TCP packets being freed and cloned.

So my thoughts here:
While moving skb tracking to the core is the right thing to do, I would make skb-tracking events an opt-in thing.
One approach is to have a global flag. Another approach is to keep a module called "skb-tracking" that just edits some config or something that makes the trackint-related probes send an event.

What do you think?

src/core/probe/common.rs
```diff
@@ -30,6 +30,10 @@ pub(crate) fn get_ebpf_debug() -> bool {
     }
 }
 
+/// Options to act on per-probe features.
 pub(crate) enum ProbeOption {
+    /// Control stack trace retrieval.
     StackTrace(bool),
+    /// Control if an event is being sent.
+    SendEvent(bool),
```
amorenoz, line 38: I haven't looked deeply at the new version but I'm wondering why you chose to do it this way rather than splitting the hook in two.

amorenoz:
I have rebased some work I have on top of this series and OVS tracking broke.
While bisecting, I found that:
80877416d39eae537418890bd3e23d50935d1840: core: probe: kernel: allow not to report events

does not build:
```
cargo build
   Compiling retis v0.1.0 (/home/amorenoz/src/retis)
error[E0609]: no field `global_probes_options` on type `&mut ProbeManager`
  --> src/core/probe/manager.rs:88:14
   |
88 |             .global_probes_options
   |              ^^^^^^^^^^^^^^^^^^^^^ unknown field
   |
   = note: available fields are: `dynamic_probes`, `dynamic_hooks`, `filters`, `config_map`, `options` ... and 2 others

error[E0609]: no field `global_probes_options` on type `&mut ProbeManager`
  --> src/core/probe/manager.rs:95:14
   |
95 |         self.global_probes_options.push(opt);
   |              ^^^^^^^^^^^^^^^^^^^^^ unknown field
   |
   = note: available fields are: `dynamic_probes`, `dynamic_hooks`, `filters`, `config_map`, `options` ... and 2 others

For more information about this error, try `rustc --explain E0609`.
error: could not compile `retis` due to 2 previous errors
```

src/core/probe/probe.rs
```diff
@@ -43,14 +44,49 @@ impl Probe {
...
+    }
+}
 
 impl Probe {
+    /// Use the underlying symbol to get its name and use it as a key which can
+    /// be used to differenciate between probes.
+    pub(crate) fn key(&self) -> String {
+        match &self {
+            Probe::Kprobe(probe) | Probe::Kretprobe(probe) | Probe::RawTracepoint(probe) => {
+                probe.symbol.name()
```
amorenoz, line 82: I'm afraid this will not work (in fact, it breaks ovs module).
kretprobes and kprobes might have the same name (because they attach to the same symbol), so this will store them under the same key replacing the first probe info with the last one that is added.
I'd suggest making Probe drive from Hash, Eq and PartialEq so it can directly be used in HashMaps

src/core/probe/manager.rs
```diff
@@ -249,9 +229,13 @@ impl ProbeManager {
 
         // First, handle generic probes.
         let mut set = ProbeSet {
-            probes: self.generic_probes.clone(),
+            probes: self
+                .probes
+                .clone()
+                .into_iter()
+                .filter(|(_, p)| p.hooks_len() == 0)
```
amorenoz: Maybe add a helper function in `Probe` such as `is_generic() -> bool` so that these filters are easier to understand?

src/core/probe/manager.rs
```diff
@@ -249,9 +229,13 @@ impl ProbeManager {
 
         // First, handle generic probes.
         let mut set = ProbeSet {
-            probes: self.generic_probes.clone(),
+            probes: self
+                .probes
+                .clone()
```
amorenoz: Not sure if this comment will be fixed later on in the same series but IIUC the ProbeSet is now a temporary object so we could give it a reference to probes and avoid cloning.

src/core/probe/manager.rs
```diff
@@ -263,13 +247,22 @@ impl ProbeManager {
...
 
         // Then targeted ones.
-        self.targeted_probes
+        self.probes
             .iter_mut()
-            .try_for_each(|set| -> Result<()> {
-                set.hooks
+            .filter(|(_, p)| p.hooks_len() != 0)
+            .try_for_each(|(_, p)| -> Result<()> {
+                p.hooks
```
amorenoz: could we move this to a method in `Probe`? e.g: `Probe::add_maps()`?

src/core/probe/manager.rs
```diff
@@ -321,9 +299,20 @@ impl ProbeSet {
         })?;
 
         // All probes loaded, issue an info log.
-        info!("{} probe(s) loaded", self.probes.len());
+        info!("{} probe(s) loaded", probes.len());
```
amorenoz, line 304: nit: I've found it confusing that this line is printed many times. Maybe we could consider moving it to ProbeManager and just printing the total  number of loaded probes?

src/core/probe/kernel/kernel.rs
```diff
@@ -19,35 +19,26 @@ use crate::core::kernel::inspect;
...
-            nargs: desc.nargs,
-            config: desc.probe_cfg,
-        })
+        Ok(KernelProbe { symbol })
     }
 
-    /// Set, for probes only, a single config option used to change the default probe behavior.
-    pub(crate) fn set_option(&mut self, opt: &ProbeOption) -> Result<()> {
-        match opt {
+    /// Generate the probe BPF configuration from a list of options.
```
amorenoz, line 30: nit: it's not really (only) "BPF" configuration any longer, right?

src/core/probe/manager.rs
```diff
@@ -207,7 +207,7 @@ impl ProbeManager {
             &mut self
                 .probes
                 .values_mut()
-                .filter(|p| p.hooks_len() == 0)
+                .filter(|p| p.hooks_len() == 0 && p.supports_generic_hooks())
```
amorenoz, line 210: Could we move this to a helper function in `Probe`?

src/core/probe/kernel/bpf/include/common.h
```diff
@@ -5,15 +5,11 @@
...
 #include <events.h>
 #include <helpers.h>
 #include <packet-filter.h>
-
-enum kernel_probe_type {
-	KERNEL_PROBE_KPROBE = 0,
-	KERNEL_PROBE_KRETPROBE = 1,
-	KERNEL_PROBE_TRACEPOINT = 2,
-};
+#include <skb-tracking.h>
```
amorenoz, line 12: do we have / need a guideline regarding casing of headers? I see both kebab and snake casing

amorenoz (commented):
I have mostly style or nit comments. I really like the probe manager rework!!!

src/core/probe/kernel/kernel.rs
amorenoz: Ok, I was thinking more in terms of BTF config

src/core/probe/manager.rs
```diff
@@ -209,132 +193,135 @@ impl ProbeManager {
...
+
+                    // Initialize the probe builder, only once for all targets.
+                    builder.init(map_fds.clone(), self.hooks.clone(), filters.to_owned())?;
+
+                    builders.insert(probe.type_key(), builder);
+                }
+                true => (),
+            }
+            // Unwrap as we just made sure the probe builder would be available.
+            let builder = builders.get_mut(&probe.type_key()).unwrap();
```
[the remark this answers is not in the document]
amorenoz: Good point

amorenoz (approved):
I don't have any further comments. @vlrpl you?

PR #119, opened 2023-04-20
Collector: allow wildcards in user defined probes (for kprobes)

atenart:
This PR allows to do things like `retis collect -p tcp_*`. One thing to notice is the test_data files should always be updated at once as there are multiple inter-dependencies.

Only kprobes are supported as it is the only probe type loading once for all probes. Other probe types could be supported but that should be done with extra care and would probably fail in many cases with errors like "too many opened files".

Based on #102 .

### DIFF
--- a/src/core/probe/manager.rs
+++ b/src/core/probe/manager.rs
@@ -251,11 +251,13 @@ impl ProbeManager {
 
     /// Attach all probes.
     pub(crate) fn attach(&mut self) -> Result<()> {
+        let mut attached = 0;
+
         // Take care of generic probes first.
         for set in self.dynamic_probes.iter_mut() {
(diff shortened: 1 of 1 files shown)


src/core/kernel/inspect.rs
```diff
@@ -216,6 +217,27 @@ pub(crate) fn get_name_offt_from_addr_near(addr: u64) -> Result<(String, u64)> {
     ))
 }
 
+/// Find functions matching a given pattern.
```
amorenoz, line 220: nit. maybe specify the supported patterns in the comment?

amorenoz (commented):
Only a small nit. Otherwise LGTM

amorenoz (approved):

PR #120, opened 2023-04-20
Use a text format for drop reasons in events

atenart:
This has multiple advantages:

- Ease to understand the reason why a packet was dropped.
- Better compatibility with different kernel versions.

Example:
```
# retis collect -c skb-drop
{"common":{"timestamp":3748134824705},"kernel":{"probe_type":"kprobe","symbol":"kfree_skb_reason"},"skb-drop":{"drop_reason":"tcp old data"}}
{"common":{"timestamp":3748134840945},"kernel":{"probe_type":"kprobe","symbol":"kfree_skb_reason"},"skb-drop":{"drop_reason":"not specified"}}
{"common":{"timestamp":3748223136046},"kernel":{"probe_type":"kprobe","symbol":"kfree_skb_reason"},"skb-drop":{"drop_reason":"tcp invalid sequence"}}
{"common":{"timestamp":3903142678490},"kernel":{"probe_type":"kprobe","symbol":"kfree_skb_reason"},"skb-drop":{"drop_reason":"no socket"}}
```

Based on #102.

### DIFF
--- a/src/module/skb_drop/event.rs
+++ b/src/module/skb_drop/event.rs
@@ -1,10 +1,17 @@
+use std::collections::HashMap;
+
 use anyhow::Result;
+use btf_rs::Type;
+use log::warn;
 use plain::Plain;
 
 use crate::{
-    core::events::{
-        bpf::{parse_single_raw_section, BpfRawSection},
-        *,
+    core::{
+        events::{
+            bpf::{parse_single_raw_section, BpfRawSection},
+            *,
+        },
+        kernel::inspect,
     },
     event_section, event_section_factory,
     module::ModuleId,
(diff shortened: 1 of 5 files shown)


src/module/skb_drop/event.rs
```diff
@@ -13,24 +20,59 @@ use crate::{
...
+        if let (btf, Type::Enum(r#enum)) =
+            inspect::btf_info()?.resolve_type_by_name("skb_drop_reason")?
+        {
+            for member in r#enum.members.iter() {
+                reasons.insert(
+                    member.val(),
+                    btf.resolve_name(member)?
+                        .trim_start_matches("SKB_")
+                        .trim_start_matches("DROP_REASON_")
+                        .replace('_', " ")
```
amorenoz, line 47: This makes it human readable but more difficult difficult for programmatic matching (e.g: postprocessing).
I'm wondering if there's a way to keep the integer as data representation and only do this when marshaling/unmarshaling  or maybe having a type that is convertible to/from string...

We would not be able to use BTF in that case though...

amorenoz: Maybe keeping the enum name and only making it human-readable for printing? Not ideal but I have no more ideas.

amorenoz: > If you see a value in providing the raw value, we could have both in the event, "skb-drop":{"drop_reason":"tcp old data", "drop_reason_raw":42}}.

If the numeric value is not stable that won't add much.
I was thinking that if we implement a reason filter: `retis collect --reason-filter="TCP_OLD_DATA"` is sligthly nicer than `retis collect --reason-filter="tcp old data"` ... Maybe it's just because I'm thinking in reasons as error codes (in my mind I'd filter by EINVAL not "invalid argument")... not a super strong issue though

amorenoz (approved):
lgtm

Issue #138, opened 2023-05-08
retis does not build with rust 1.65


amorenoz:
It yields the following error:

```
error[E0658]: custom discriminant values are not allowed in enums with tuple or struct variants
  --> src/core/filters/packets/ebpfinsn.rs:12:10
   |
11 |     Bpf(BpfJmpOp),
   |     ------------- tuple variant defined here
12 |     Ne = bpf_sys::BPF_JNE,
   |          ^^^^^^^^^^^^^^^^ disallowed custom discriminant
   |
   = note: see issue #60553 <https://github.com/rust-lang/rust/issues/60553> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `retis` due to previous error
root@823493a14136:/src# rustc --version
rustc 1.65.0 (897e37553 2022-11-02)
root@823493a14136:/src# 
```

Do we want to support it?
Note it's the one supported by latest Ubuntu LTS...

vlrpl:
fixed with #139

PR #139, opened 2023-05-08
fix build with rustc 1.65

vlrpl:
as reported in #138 by @amorenoz, the build fails using rustc 1.65.

This happens because custom discriminant values are not allowed in enums with tuple or struct variants in rust in versions < 1.66 which is of course the one we need :)

Fix it by removing the discriminant as it was fundamentally introduced for consistency.


amorenoz:
Fixes the issue, thanks for the quick patch @vlrpl

amorenoz (approved):

Issue #100, opened 2023-03-16
core: provide kernel version and loaded modules


amorenoz:
As discussed in  https://github.com/retis-org/retis/pull/90#discussion_r1138455343_

Modules should check if their dependencies are satisfied and exit early (and gracefully)

PR #145, opened 2023-05-10
Allow Retis & its modules to check for prerequisites

atenart:
In this PR:

- The inspector is moved to its own crate, and reworked a bit for easier use.
- New helpers are added to provide info on the kernel version, configuration, OS, loaded modules.
- A match engine for the kernel version is provided.

This is a prerequisite for #100 and for #115. Current form should be good enough to use as a base for both PRs.


src/module/ovs/ovs.rs
```diff
@@ -45,6 +46,33 @@ impl Collector for OvsCollector {
...
+                    if inspector.kernel.is_module_loaded("openvswitch") == Some(false) {
+                        bail!("Kernel module 'openvswitch' is not loaded")
+                    }
+                }
+                _ => (),
+            }
+        }
+
+        // Then check if the openvswitch user process is running.
+        if !proc::find_process_by_name("ovs-vswitchd")? {
```
amorenoz, line 69: Technically, the module can run even if ovs-vswitchd is not present. Only `--track` will not work. Not sure if we need to express that somehow here.

Also, ovs kernel module is ns-aware. So it's common to run several! ovs-vswitchd instances in containers, all of them using the same kernel module. Again, --track will not work but still kernel module can be traced

src/module/ovs/ovs.rs
```diff
@@ -45,6 +46,33 @@ impl Collector for OvsCollector {
...
     }
 
+    // Check if the OvS collector can run. Some potential errors are silenced,
+    // to avoid returning an error if we can't inspect a given area for some
+    // reasons.
+    fn can_run(&self) -> Result<()> {
+        let inspector = inspect::inspector()?;
+
+        // First check the OvS kernel module is available.
+        if let Ok(kconf) = inspector.kernel.get_config_option("CONFIG_OPENVSWITCH") {
```
amorenoz, line 56: OVS did maintain for some time an oot kernel module (in OVS's tree) that could be used. Not that it's very common AFAIK (plus, it was removed from OVS's tree earlier this year), but: wouldn't the presence of the module be enough?

src/core/inspect/check.rs
```diff
@@ -0,0 +1,54 @@
...
+use anyhow::{bail, Result};
+use log::warn;
+
+use super::{inspect, kernel_version::KernelVersionReq};
+
+/// Check various prerequisites for Retis to properly work, allowing to bail out
+/// early and to explain what isn't compatible to the user. This helper can also
+/// be used to show warning or information that do not prevent Retis from
+/// starting.
+pub(crate) fn collection_prerequisites() -> Result<()> {
```
amorenoz, line 12: Something that would be absolutely awesome would be to verify the process has CAP_EBPF! Not necessarily for this PR

src/module/ovs/ovs.rs
amorenoz: > > Technically, the module can run even if ovs-vswitchd is not present. Only `--track` will not work. Not sure if we need to express that somehow here.
> 
> So maybe we should have the cli as an argument to `can_run` because some capabilities might depends on some flags. And then we can add the check only if `--track` was used.
> 

Having the cli as an argument might be cool (although we might end have with a chicken-and-egg situation).

> Out of curiosity, can the ovs module handle packets if the userspace part is not running?
> 

Yes. Temporarily when you restart the userspace part, the kernel works headless. Upcalls are dropped but currently established traffic flows

> > Also, ovs kernel module is ns-aware. So it's common to run several! ovs-vswitchd instances in containers, all of them using the same kernel module. Again, --track will not work but still kernel module can be traced
> 
> Interesting, so if we want to do this correctly we should also check if those processes match the current ns. (Isn't something we should do in `Process` too to avoid the `PID_ALL` case?).

The common use case (at least for me) is to use [kind](https://kind.sigs.k8s.io/). In such cases the module will just not find the process by cmd and --track will not work. Still it's good to see what goes through the kernel datapath. Even more when we have ns filters. 
To find the right process we'd need to jump into the pid-ns of the container. Not sure how to do it cleanly so I don't think we need to make "--track" work on kind today. For now, it's just another case in which the userspace might not be available but kernel-only ovs module should be able to work.

amorenoz: Cool. BTW, I think it's OK to fail if "--track" has been provided and "ovs-vswitchd" is not present (or more than one is) but I guess we'll fail when the module is initialized anyway and the error is not very bad.

amorenoz:
I wanted to rebase #115 on top of this to test but it's proven complicated because it also breaks #155 . So I'd say we can merge this and I can fix any issue I find afterwards, wdty?

src/core/inspect/check.rs
vlrpl: as a note, at least access to debugfs and kallsyms (kptr_restrict = 0, perf_event_paranoid = 0) is needed.
Not sure there's something else missing. I don't think it's a goal for the time being, maybe for the future.

vlrpl (approved):

PR #149, opened 2023-05-16
Unflatten skb event & report more fields in the event

atenart:


src/module/skb/event.rs
```diff
@@ -6,61 +6,104 @@ use crate::{
     event_section, event_section_factory,
 };
 
-/// Skb event section
-/// TODO: unflatten?
+/// Skb event section.
 #[event_section]
 pub(crate) struct SkbEvent {
-    // L2 fields
+    eth: Option<SkbEthEvent>,
```
amorenoz, line 12: Shouldn't these fields be `pub(crate)` (and documented accordingly)?

amorenoz (approved):

PR #150, opened 2023-05-16
Fix core::user::proc::tests::get_threads test

atenart:
The `core::user::proc::tests::get_threads` test sometimes failed. It turns out `get_thread_info` could throw errors when the expected file/dir tree was a bit off (no such file or directory).


vlrpl (commented):
this fixes the failure and lgtm.

vlrpl (approved):

PR #152, opened 2023-05-16
Minor follow-ups on thread termination

atenart:
Mainly:
- A message for the user as the termination process can take up to ~1sec.
- Divided by 5 the skb tracking GC termination time.

### DIFF
--- a/src/core/tracking/skb_tracking.rs
+++ b/src/core/tracking/skb_tracking.rs
@@ -218,9 +218,18 @@ pub(crate) fn init_tracking(
     let gc = thread::spawn(move || {
         let tracking_map = tracking_map.get_mut();
 
-        while run_state.running() {
+        let running = || -> bool {
             // Let's run every SKB_TRACKING_GC_INTERVAL seconds.
-            thread::sleep(Duration::from_secs(SKB_TRACKING_GC_INTERVAL));
+            for _ in 0..SKB_TRACKING_GC_INTERVAL {
+                thread::sleep(Duration::from_secs(1));
+                if !run_state.running() {
+                    return false;
+                }
+            }
+            true
+        };
+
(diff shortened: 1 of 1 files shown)


src/collect/collector.rs
```diff
@@ -249,6 +253,14 @@ impl Collectors {
         let mut json = JsonFormat::default();
         let mut processors = get_processors(&mut json, collect.args()?)?;
 
+        // Let the user know we got a termination signal and we're processing
```
vlrpl, line 256: not sure if there's some limitation that prevents it, but have you considered replacing register_signal() with something similar to the code below?

src/core/tracking/skb_tracking.rs
```diff
@@ -220,7 +220,13 @@ pub(crate) fn init_tracking(
 
         while run_state.running() {
             // Let's run every SKB_TRACKING_GC_INTERVAL seconds.
-            thread::sleep(Duration::from_secs(SKB_TRACKING_GC_INTERVAL));
+            for _ in 0..SKB_TRACKING_GC_INTERVAL {
+                thread::sleep(Duration::from_secs(1));
+                if !run_state.running() {
+                    break;
```
vlrpl, line 226: Mostly cosmetic (and optional), but I guess a return here makes it more obvious (it is fairly obvious already, tbf :)
Also (optional :), if moving the running() check before sleeping on top of it, allows to get rid of the "while" condition and just loop {} which looks a bit redundant. Not exactly the same, but should be ok.
But if you prefer to keep it so, lgtm.

vlrpl: ack, looks good

vlrpl (approved):
lgtm, approved!

PR #156, opened 2023-05-24
Do not install probes with unknown types

atenart:


src/collect/collector.rs
```diff
@@ -324,11 +324,10 @@ impl Collectors {
                     break;
                 }
             }
+            // Skip probes which won't generate events from the collectors.
             if !valid {
-                warn!(
-                    "A probe to symbol {} is attached but no collector will retrieve data from it, only generic information will be retrieved",
-                    symbol
-                );
+                info!("Not probe was attached to {symbol} as no collector could retrieve data from it");
```
amorenoz, line 329: s/Not/No/?

amorenoz (commented):
I remember we discussed this in the chat but cannot recall the details.

Do you think a user might want to get events (just basic Common+Kernel sections) from probes that are relevant? Maybe we should then add a "generic" collector or something like that?

I'm not saying we should block this PR, just thinking about  possible use cases that might need "thin" events.

amorenoz (approved):
/lgtm

PR #159, opened 2023-05-25
Event display

### LINKED ISSUE #113: Improve stdout event output
Use a better format than JSON and potentially allow users to select which fields to display.

atenart:
Implements the core logic and then a proper output for all event sections. I tried to minimize the length while keeping it understandable.

RFC as:

- I did my best (reusing most of @amorenoz first implementation) for the OvS events. One enum variant is partially implemented though. I'm sure the ovs event can be improved but that require ovs knowledge.
- I did not check all fields reported by the skb events where genuine, even though it should be ~close to the truth.

But those are "minor" issues, and the series can be reviewed.

Fixes #113.


src/core/events/bpf.rs
```diff
@@ -299,6 +299,14 @@ pub(crate) struct CommonEvent {
 
 unsafe impl Plain for CommonEvent {}
 
+impl EventFmt for CommonEvent {
+    fn event_fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // Don't put an extra space at the beginning as this section is 1)
+        // always displayed and 2) always first.
+        write!(f, "{}", self.timestamp)
```
amorenoz, line 306: To me it feels weird that Sections have to know their relative position or account for the format of others. Not that I see a cleaner way to do without it.

src/module/skb/event.rs
```diff
@@ -33,7 +33,156 @@ pub(crate) struct SkbEvent {
...
+                if dataref.fclone > 0 {
+                    write!(f, "fclone {} ", dataref.fclone)?;
+                }
+                write!(f, "users {} dataref {}", dataref.users, dataref.dataref)?;
+            }
+
+            write!(f, "]")?;
+        }
+
+        Ok(())
```
amorenoz, line 185: Essentially this is formatting 100% of the event info. As soon as we add some other event sections this will not fit in a single terminal line, in which case the output starts to get messy.
Do you think it makes sense to make an opinionated split and only print a subset of the fields by default allowing for "-v", "-vv", "-vvv" to increase the information shown?

amorenoz:
Should we have a default way to print out stacktraces?

src/core/probe/kernel/kernel.rs
```diff
@@ -59,6 +59,12 @@ pub(crate) struct KernelEvent {
     pub(crate) stack_trace: Option<Vec<String>>,
 }
 
+impl EventFmt for KernelEvent {
+    fn event_fmt(&self, f: &mut fmt::Formatter, _: DisplayFormat) -> fmt::Result {
+        write!(f, " {:?}", self)
```
amorenoz: don't we need to remove the space now that Event::event_fmt is in charge of high level formatting?

src/module/ovs/event.rs
```diff
@@ -107,6 +134,18 @@ pub(crate) struct UpcallEvent {
 }
 unsafe impl Plain for UpcallEvent {}
 
+impl EventFmt for UpcallEvent {
+    fn event_fmt(&self, f: &mut fmt::Formatter, _: DisplayFormat) -> fmt::Result {
+        write!(
+            f,
+            "UPCALL{} port {} cpu {}",
```
amorenoz, line 141: The rest of the events are using lowercase. OVS should be kept that way as well.

src/collect/cli.rs
```diff
@@ -58,6 +58,13 @@ Example: --probe tp:skb:kfree_skb --probe kprobe:consume_skb"
         default_value = "false"
     )]
     pub(super) print: bool,
+    #[arg(
+        long,
+        help = "Pretty-print the event in a given format",
+        value_parser=PossibleValuesParser::new(["singleline", "multilines"]),
```
amorenoz, line 64: s/multilines/multiline?

amorenoz: Also, if you use the enum directly as the field type clap will generate the possible values for you (con: you need the num to derive from clap::ValueEnum)

log
```diff
@@ -0,0 +1,12 @@
+{"common":{"timestamp":32925833732191},"kernel":{"probe_type":"raw_tracepoint","symbol":"skb:kfree_skb"},"skb":{"ip":{"daddr":"2a00:1450:4007:813::200a","len":41,"protocol":17,"saddr":"2a01:e0a:2e3:74c2:b2d8:4bac:5ce:a752","version":6},"udp":{"dport":443,"len":41,"sport":42561}},"skb-drop":{"drop_reason":"NOT_SPECIFIED"},"skb-tracking":{"orig_head":18446616923382341120,"skb":18446616939029413632,"timestamp":32925833732191}}
```
amorenoz, line 1: Maybe you've accidentally committed this file?

amorenoz (commented):
In general, it's an a pretty good shape.

One question: How should EventSeries be formatted? should we add a function in Formatter to format and print a `Series` (adding a somewhat weird dependency unless `Series` is moved to core)?

src/core/probe/kernel/kernel.rs
amorenoz: Saw that it's removed in a later commit. Ignore this.

amorenoz:
> > One question: How should EventSeries be formatted? should we add a function in Formatter to format and print a `Series` (adding a somewhat weird dependency unless `Series` is moved to core)?
> 
> I'd say `Formatter` is the last piece of the chain we might want to change. I don't think we should handle series directly in the `Formatter`, so we might want to remove this abstraction.
> 
> IMHO it should be up to the command (whether `collect` or post-processing ones) to decide how to print the events (not how to format them). In the `sort` case, we probably want to have the first event non-indented and later ones with some indentation.
> 

For this to work in `Multiline` format we would need to pass down the indentation to `Event::event_fmt` right?
Something like 
```rust
pub(crate) enum DisplayFormat {
    SingleLine,
    #[default]
    MultiLine(u32),
}
```

> One option could be to ~move the Formatter trait into a post-process struct, something like `PrintSingle` and handle the output of series in the `sort` cmd implementation directly (as it's the only user for now). Then those could be used in post-process and/or collect commands. (And we would remove the `TextFormat` and `JsonFormat` as they would not provide anything at that point).
> 
> Thoughts?

Yep, sounds reasonable. I'll create a PR on top of this one implementing Series printing.

src/core/events/display.rs
```diff
@@ -0,0 +1,52 @@
+use std::fmt;
+
+#[allow(dead_code)] // FIXME
```
amorenoz: I think you can drop this line and cargo will still be happy

src/core/events/display.rs
```diff
@@ -0,0 +1,52 @@
...
+#[derive(Default, Clone, Copy)]
+pub(crate) enum DisplayFormat {
+    SingleLine,
+    #[default]
+    MultiLine,
```
amorenoz, line 8: BTW, I found a bit confusing that the default value on this enum is `Multiline` but on the CLI it's `SingleLine`. Why is that?

src/collect/cli.rs
```diff
@@ -58,6 +61,12 @@ Example: --probe tp:skb:kfree_skb --probe kprobe:consume_skb"
         default_value = "false"
     )]
     pub(super) print: bool,
+    #[arg(
+        long,
+        help = "Pretty-print the event in a given format",
```
amorenoz, line 66: I find this text a bit confusing. For me it means that if you don't add this arg, events are not printed but they are, this is just an opt to select the format not to select whether to print them or not (for that we have --print). How about "--format"?

amorenoz: So when we implement custom formatting we'll use `--custom-fmt "some format"`? If we use `--format` we'd need something like `--format custrom --custom-fmt "..."` or maybe just `--format "someformat"`...
Either work for me. My only concern is the verb in that sentence. For me, saying that the option "pretty-prints the event" implies that if not given, the even is not printed. Maybe something like "selects one of the predefined pretty formats for event printing" makes it a bit clearer that the event would be printed anyhow and that this option only allows you to say how.

src/module/ovs/event.rs
```diff
@@ -196,6 +259,22 @@ impl OperationEvent {
...
 
+impl EventFmt for OperationEvent {
+    fn event_fmt(&self, f: &mut fmt::Formatter, _: DisplayFormat) -> fmt::Result {
+        write!(
+            f,
+            "flow_{} q {} ts {} ({})",
+            OperationEvent::operation_str(self.op_type)
+                .unwrap_or("?")
+                .to_string()
+                .to_uppercase(),
```
amorenoz, line 270: I think we can also remove this "to_uppercase"

src/collect/cli.rs
```diff
@@ -58,6 +61,9 @@ Example: --probe tp:skb:kfree_skb --probe kprobe:consume_skb"
         default_value = "false"
     )]
     pub(super) print: bool,
+    #[arg(long, help = "Format used when printing an event.")]
+    #[clap(value_enum, default_value_t=DisplayFormat::MultiLine)]
+    pub(super) format: DisplayFormat,
```
amorenoz, line 66: tiny nit: I find more readable if help is just the doc-comment. feel free to ignore as the rest of the fields don't have.

amorenoz (approved):
/lgtm

PR #180, opened 2023-06-07
Add two drop monitor-like profiles

atenart:


profiles/drop.yaml
```diff
@@ -0,0 +1,18 @@
+version: 1.0
+name: drop
```
amorenoz, line 2: Does it make sense to name it "drop-mon"  or something similar that indicates that's a monitor-kind of profile?
I'm thinking that the name "drop" might make someone think that's just adding skb-drop and might combine it with lots of probes in, say, the TCP stack. The result would be quite noisy with lots of stack traces.

profiles/drop.yaml
```diff
@@ -0,0 +1,18 @@
...
+about: Drop monitor-like
+collect:
+  - name: Default
```
amorenoz, line 5: You can drop the "name: Default" if you like

Dockerfile
```diff
@@ -13,6 +13,9 @@ RUN dnf install -y \
 
 # Only download the dependencies for now so these steps can be cached.
 RUN cargo init
+COPY retis-derive retis-derive
+# `cargo -C <path>` is unstable for now.
+RUN (cd retis-derive && cargo fetch --locked)
```
amorenoz, line 20: Brackets are unnecessary

profiles/drop.yaml
```diff
@@ -0,0 +1,18 @@
...
+      collectors: skb-drop,skb
+      stack: ~
+---
+version: 1.0
+name: nft-drop
+about: Drop monitor-like for netfilter
+collect:
+  - name: Default
+    args:
+      collectors: nft,skb
```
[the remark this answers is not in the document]
amorenoz: Yep.

amorenoz (approved):
/lgtm

PR #184, opened 2023-06-09
profiles: warn in case of name conflict in the same src dir

atenart:


vlrpl (commented):
thanks @atenart and @amorenoz.
LGTM as well. Merged!

vlrpl:
it seems I arrived some seconds later :)

PR #185, opened 2023-06-09
module: skb: add more IP fields

atenart:


amorenoz:
My only comment is about the semantics/readability of the resulting json. Having:
```json
"skb": { "ip": { "version":  { "v4":  { "tos" : 0 } } } }
```
seems like a bit ugly and redundant  (the version field only has one member  that starts with "v").

I don't have a clear alternative though, I've tried adding #[serde(flatten)] to `version` so the result is:
```json
"skb": { "ip": { "v4":  { "tos" : 0 } } }
```
which imho is only sligthly better. If we were to handle the json object (or resulting python dict) we would have to test for the existence of a field. This would of course be fixed by a helper function at the Event level.

Another alternative would be to have something like:
```json
"skb": { "ip": { "version": "v4", "version_flags":  { "tos" : 0 } } }
```

But I'm not 100% convinced either.

Any thoughts?

amorenoz:
> > My only comment is about the semantics/readability of the resulting json. Having:
> > ```json
> > "skb": { "ip": { "version":  { "v4":  { "tos" : 0 } } } }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > seems like a bit ugly and redundant (the version field only has one member that starts with "v").
> 
> Agreed.
> 
> > I don't have a clear alternative though, I've tried adding #[serde(flatten)] to `version` so the result is:
> > ```json
> > "skb": { "ip": { "v4":  { "tos" : 0 } } }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which imho is only sligthly better. If we were to handle the json object (or resulting python dict) we would have to test for the existence of a field. This would of course be fixed by a helper function at the Event level.
> 
> Something like:
> 
> ```rust
> impl SkbIpEvent {
>     fn v4 -> Option<SkbIpv4Event> {}
>     fn v6 -> Option<SkbIpv6Event> {}
> }
> ```
> 
> ?
> 

For example, yes. So you think flattening the version struct is OK?

amorenoz (approved):
/lgtm

PR #200, opened 2023-06-14
profiles: add generic

atenart:
Provides a starting point for debugging the net stack.

Open question: is "generic" a good name? I'm not 100% sure :)

### DIFF
--- a/profiles/generic.yaml
+++ b/profiles/generic.yaml
@@ -1,0 +1,29 @@
+version: 1.0
+name: generic
+about: Generic set of probes, meant to be used as a starting point for debugging sessions
+collect:
+  - args:
+      collectors: skb,skb-tracking
+      probe:
+        - tp:skb:kfree_skb
+        - tp:net:netif_receive_skb
+        - tp:net:netif_rx
+        - kprobe:ip_rcv
+        - kprobe:ipv6_rcv
+        - kprobe:tcp_v4_rcv
+        - kprobe:tcp_v6_rcv
+        - kprobe:tcp_gro_receive
+        - kprobe:tcp6_gro_receive
+        - kprobe:tcp_rcv_established
+        - kprobe:tcp_rcv_state_process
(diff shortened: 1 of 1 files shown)


vlrpl:
> Provides a starting point for debugging the net stack.
> 
> Open question: is "generic" a good name? I'm not 100% sure :)

IMHO, doesn't sound bad. Could "base" be a decent alternative?

profiles/generic.yaml
```diff
@@ -0,0 +1,27 @@
+version: 1.0
+name: generic
+about: Generic set of probes, meant to be used as a starting point for debugging sessions
+collect:
+  - args:
```
amorenoz, line 5: Should we at least add the skb collector? and maybe a reasonable set of skb-sections?

amorenoz: IMHO the user experience is way better if users can just start by typing "retis -p generic collect" and start seeing something.

amorenoz: You don't have to combine with other profiles, running `retis -p generic collect -c skb-drop` won't print anything on all those probes. I think profiles should try no to make assumptions on what other profiles or other pieces of command line add.
That's why options such as `--collectors`, `--skb-sections`, etc can be repeated. Sure, the `nft` profile _might_ enable the `skb` collector, same as the `dropmon`, but if you select both or either one individually they all provide reasonable data.
Having reasonable defaults is a UX feature. It just makes users' life a bit easier by allowing them not to express an option but I don't think it should be part of the API (if profiles rely on it it would be).
I think having the skb collector enabled in the profile has no cons and closes gaps where the tool might not work as expected.

amorenoz (approved):
lgtm

profiles/generic.yaml
amorenoz: what, should we also add "dev" as skb-section? or do you think this might be an overkill?

PR #205, opened 2023-06-16
Small README improvements on the container section

atenart:


README.md
```diff
@@ -406,6 +406,8 @@ $ podman run --privileged --rm -it --pid=host -v /sys/kernel/debug/:/sys/kernel/
       quay.io/retis/retis:latest --help
 ```
 
+Or with Docker, replacing `podman` by `docker` in the above.
```
vlrpl, line 411: unreliable nit  (non-native speaker here), but I guess we could rephrase using the following:

Or using `docker` in place of `podman` in the above

otherwise

Or with Docker, by replacing `podman` with `docker` in the above.

PR #212, opened 2023-07-05
merge --ebpf-debug into --log-level

atenart:
Use `--log-level debug` to enable BPF debug in libbpf instead of a specific option. This makes thing more consistent and helps in removing the not perfect ebpf-debug handling internally. This prevents from only turning the BPF debug logging on or from only getting non-BPF debug messages, but IMHO that isn't really an issue.

### DIFF
--- a/src/collect/cli.rs
+++ b/src/collect/cli.rs
@@ -18,8 +18,6 @@ use crate::{
 
 #[derive(Args, Debug, Default)]
 pub(crate) struct CollectArgs {
-    #[arg(long, default_value = "false")]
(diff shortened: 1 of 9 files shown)


vlrpl:
I this libbpf_rs::set_print() fits perfectly with our purpose, and removing the option is a good thing (don't find it super convenient).
The only concern is about the amount of data printed in case of debug, but that could be expected using that.
It might potentially grow if bpf_sys starts collecting the log_buf (which might be useful in case a filter fails to load).

vlrpl (approved):
lgtm, thanks!

PR #209, opened 2023-06-22
Support more kernel configuration locations

atenart:
Fixes #207. Details in the commit messages.

PR #215, opened 2023-07-10
Add a --kconf option to provide a custom kernel configuration path

atenart:
Also modify the container image to always use `--kconf` and let the user map the right configuration file (and only that file, not a full directory).

Plus a few smalls changes.

Based on #209.

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,6 +403,17 @@ $ retis --help
 
 ### Container image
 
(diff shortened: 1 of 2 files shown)


vlrpl (approved):
lgtm. Thanks!
I'll let one of you merge as you might have some last-minute changes/suggestions.

PR #216, opened 2023-07-11
Upgrade to libbpf_rs v0.21.2

atenart:


amorenoz:
Isn't this ubuntu 23.04: https://app.vagrantup.com/ubuntu/boxes/lunar64 ?

src/core/tracking/gc.rs
```diff
@@ -26,6 +26,10 @@ pub(crate) struct TrackingGC {
     limit: u64,
     // The name of the thread
     name: String,
+    // Internal configuration map (optional). Here to keep a reference, not to
+    // be used.
+    #[allow(dead_code)]
+    config_map: Option<libbpf_rs::MapHandle>,
```
amorenoz, line 32: I don't think using the thread to hold on to arbitrary useless maps is very clean.

No module should need to use this as they are stateful objects that can store data, maybe we should do the same with this "core" modules or think of a better way for them to keep data around.

src/module/skb/bpf/skb_hook.bpf.c
```diff
@@ -111,6 +111,7 @@ struct skb_data_ref_event {
 static __always_inline int process_skb_l2_l4(struct retis_context *ctx,
 					     struct retis_raw_event *event,
 					     struct skb_config *cfg,
+					     u64 sections,
```
amorenoz, line 114: Whats the benefit of this? Does it compensate making the functions uglier and more confusing?

src/core/probe/kernel/kprobe.rs
```diff
@@ -28,7 +30,7 @@ impl ProbeBuilder for KprobeBuilder {
 
     fn init(
         &mut self,
-        map_fds: Vec<(String, i32)>,
+        map_fds: Vec<(String, RawFd)>,
```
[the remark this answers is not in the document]
amorenoz: You're right, not trivial. Let's tackle that in a different series.

amorenoz (approved):
LGTM

PR #222, opened 2023-07-27
Make skb-drop collector work on older kernels

atenart:
Do not fail if a kernel is old enough not to support skb drop reasons, so that the skb-drop reasons still work (installing probes at the right places) and especially for the dropmon profile to work (which still provides useful data).

### DIFF
--- a/src/module/skb_drop/bpf/skb_drop_hook.bpf.c
+++ b/src/module/skb_drop/bpf/skb_drop_hook.bpf.c
@@ -5,20 +5,19 @@
 #include <common.h>
 
 struct skb_drop_event {
-	u32 drop_reason;
+	s32 drop_reason;
 } __attribute__((packed));
 
 DEFINE_HOOK(F_AND, RETIS_F_PACKET_PASS,
 	struct skb_drop_event *e;
 
-	if (!retis_arg_valid(ctx, skb_drop_reason))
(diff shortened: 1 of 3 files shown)


amorenoz (commented):
I'm planning to test it (didn't have time yet). But, looking at the code in `src/module/skb_drop/bpf/skb_drop_hook.bpf.c`, if the enum is not defined and `kfree_skb` does not have that argument... will any drop event get sent to userspace?

amorenoz:
> > I'm planning to test it (didn't have time yet). But, looking at the code in `src/module/skb_drop/bpf/skb_drop_hook.bpf.c`, if the enum is not defined and `kfree_skb` does not have that argument... will any drop event get sent to userspace?
> 
> This is the tricky part, we won't get the `skb-drop` event. But we will still get one from the `skb` module if enabled. It's not perfect, but this is only to support older kernels and will be removed at some point. In such case on old kernels, the value of the skb-drop module is just to add a probe.

Would it be useful to generate a skb-drop event with an "UNKNOWN" reason? or try  to be forward compatible and fake a `SKB_DROP_REASON_NOT_SPECIFIED`? That way you don't require an extra module and the events have the same sections

amorenoz (approved):
LGTM. Thanks!

PR #223, opened 2023-07-27
ARP support

atenart:


src/module/skb/bpf/skb_hook.bpf.c
```diff
@@ -108,52 +108,19 @@ struct skb_data_ref_event {
 } __attribute__((packed));
 
 /* Must be called with a valid skb pointer */
-static __always_inline int process_skb_l2_l4(struct retis_raw_event *event,
-					     struct skb_config *cfg,
-					     struct sk_buff *skb)
+static __always_inline int process_skb_ip(struct retis_raw_event *event,
+					  struct skb_config *cfg,
+					  struct sk_buff *skb,
+					  unsigned char *head,
```
amorenoz, line 184: Why do we need to pass `head` here? just to save us from BPF_CORE_READ() it again?

src/module/skb/bpf/skb_hook.bpf.c
```diff
@@ -107,6 +119,64 @@ struct skb_data_ref_event {
...
+	/* Sender hardware address */
+	ptr = (unsigned char *)(arp + 1);
+	bpf_probe_read_kernel(&e->sha, sizeof(e->sha), ptr);
+
+	/* Sender protocol address */
+	ptr += 6 /* addr len */;
+	bpf_probe_read_kernel(&e->spa, sizeof(e->spa), ptr);
+
+	/* Target hardware address */
+	ptr += 4 /* addr len */;
```
amorenoz, line 170: nit, I'd put "hw addr len" or "proto addr len" or else it's a bit confusing that we sometimes add 4 and sometimes we add 6.

src/module/skb/event.rs
```diff
@@ -249,6 +269,21 @@ pub(crate) struct SkbEthEvent {
     pub(crate) dst: String,
 }
 
+/// ARP fields.
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
+pub(crate) struct SkbArpEvent {
+    /// Operation type (1: REQUEST, 2: REPLY).
+    pub(crate) operation: u16,
```
amorenoz, line 276: Can we make this an enum for easy matching / printing?

amorenoz (commented):
Apart from some small comments I have a suggestion: how about a functional test? :-D

src/module/skb/bpf/skb_hook.bpf.c
amorenoz: I compared the resulting ebpf with and without this change. I was curious to see if the "-O2" that we're giving clang is would optimize out the extra BPF_CORE_READ() but it seems that's not the case. I was trying to avoid making the functions uglier but in this case it seems there is a small gain after all.

amorenoz (approved):

PR #226, opened 2023-08-02
Runtime tests for skb and skb-drop

atenart:
Adding runtime tests for the `skb` and `skb-drop` modules:

- Sanity checks for both `skb` and `skb-drop`.
- TCP CC test for `skb`.

More tests can be added later (I'm thinking other protocols than TCP and `skb-tracking` in a more complex setup than two ns connected via a veth pair). But at least having some tests is better than nothing :)

### DIFF
--- a/tests/test_skb.py
+++ b/tests/test_skb.py
@@ -34,7 +34,7 @@ def test_skb_tcp_cc(two_ports_skb):
         "--skb-sections",
         "all",
         "-f",
-        "tcp port 80",
+        "tcp port 80 or arp",
         "-p",
         "tp:net:netif_rx",
     )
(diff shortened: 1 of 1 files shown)


tests/test_skb.py
```diff
@@ -0,0 +1,54 @@
...
+    retis = Retis()
+
+    retis.collect("-c", "skb", "-f", "icmp", "-p", "kprobe:ip_rcv")
+    print(ns.run("ns0", "ip", "link", "show"))
+    print(ns.run("ns0", "ping", "-c", "1", "10.0.42.2"))
+    retis.stop()
+
+    events = retis.events()
+    print(events)
+    execs = list(
```
amorenoz, line 17: not sure if the name of the variable is very meaningful, maybe a c-p leftover?

tests/test_skb.py
```diff
@@ -0,0 +1,54 @@
...
+    ipr.link('set', ifname='veth10', net_ns_fd='ns1')
+
+    # Setup ifaces
+    ns0.link('set', ifname='veth01', state='up')
+    ns1.link('set', ifname='veth10', state='up')
+    ns0.addr('add', index=ns0.link_lookup(ifname='veth01')[0], address='10.0.42.1', prefixlen=24)
+    ns1.addr('add', index=ns1.link_lookup(ifname='veth10')[0], address='10.0.42.2', prefixlen=24)
+
+    ipr.close()
+    yield netns
```
amorenoz: After yielding you can add code to clean things up, e.g: remove the interfaces.

amorenoz: oh, cool

amorenoz (approved):

PR #233, opened 2023-08-16
module: skb_drop: do not generate fake reason for all events

atenart:
Commit 915947cd1999 ("module: skb_drop: do not fail on old kernels") introduced a "fake" event to support old kernels not have drop reasons support. However this led to including the fake reason in every even on those old kernels, and for every event not including a genuine reason on newer kernels.

Fix this by:
- Only attaching the probe to the skb:free_skb tracepoint in case the kernel doesn't support drop reasons.
- Check the drop reasons enum exists in BPF side to better support both cases.

Fixes: 915947cd1999 ("module: skb_drop: do not fail on old kernels")

### DIFF
--- a/src/module/skb_drop/skb_drop.rs
+++ b/src/module/skb_drop/skb_drop.rs
@@ -14,11 +14,15 @@ use crate::{
     module::{Module, ModuleId},
 };
 
-pub(crate) struct SkbDropModule {}
+pub(crate) struct SkbDropModule {
+    reasons_available: bool,
+}
 
 impl Collector for SkbDropModule {
     fn new() -> Result<Self> {
-        Ok(Self {})
+        Ok(Self {
+            reasons_available: true,
+        })
     }
 
     fn known_kernel_types(&self) -> Option<Vec<&'static str>> {
@@ -54,6 +58,7 @@ impl Collector for SkbDropModule {
                     bail!("Could not retrieve skb drop reasons from the kernel");
                 } else {
                     warn!("This kernel doesn't provide skb drop reasons");
+                    self.reasons_available = false;
                 }
             }
             _ => (),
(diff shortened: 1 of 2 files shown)


amorenoz:
nit,  commit message:
> However this led to including the fake reason in every even on those old kernels, and for every event not including a genuine reason on newer kernels.

s/every even/every event/

src/module/skb_drop/skb_drop.rs
```diff
@@ -63,13 +68,23 @@ impl Collector for SkbDropModule {
     }
 
     fn init(&mut self, _: &CliConfig, probes: &mut ProbeManager) -> Result<()> {
-        let symbol = Symbol::from_name("skb:kfree_skb")?;
+        let mut probe = Probe::raw_tracepoint(Symbol::from_name("skb:kfree_skb")?)?;
+        let hook = Hook::from(skb_drop_hook::DATA);
 
-        if let Err(e) = probes.register_probe(Probe::raw_tracepoint(symbol)?) {
+        if self.reasons_available {
+            probes.register_kernel_hook(hook)?;
```
amorenoz, line 75: I'm wondering what other functions (apart from `kfree_skb_reason`) are we interested in probing?
If a kernel module does some internal processing where it uses core's `enum skb_drop_reason` , do we want to probe it?

Maybe it would be just easier to add a directed hook here as well?

amorenoz: I'm not suggesting we hook onto more probes. On the contrary, I'm asking why do we need to register the kernel hook for user-provided hooks. I just don't see the case where it'll be needed/wanted.

amorenoz: Right. I don't see much value and it can even be confusing if some module manipulates error codes without them being actually applied to kfree_skb_reason.

amorenoz: Yeah, I meant to say `-p dropmon collect -p...` or the  similar `-c skb,skb-drop -p...`
My general understanding is that skb-drop is all about skb drops, not any function that handles `skb_drop_reason`.
I just don't see a use-case for extracting that information from anything else but that one tracepoint. So we're adding that hook to every single probe and taking one out of the 10 slots we have without providing any real value.

amorenoz: > Maybe the correct answer here would be to have two collectors: skb-drop and drop-reason and to let the user combine them (and of course the dropmon profile in the first place). Or to simply remove the probe on tp:skb:kfree_skb and let that decision to the profile and/or user?

Ok, ok. I've failed to convince you to remove the generic hook, but let's not do the exact opposite xD

amorenoz (approved):

PR #236, opened 2023-08-22
module: skb: do not fail on invalid net device information

atenart:
Retrieving information from `skb->dev` is tricky as this is inside an union and there is no way we can know of the data is valid. Try our best to report an empty section if the data does not look like what it should.


amorenoz (approved):
LGTM

PR #237, opened 2023-08-22
build.rs: split tool and ebpf program build

vlrpl:
- build time halved
- finer-grained and better rebuild prereq checks
- clean target actually cleans the tree
- can be easier to use the available tooling (maybe linking progs, if we'll ever use that)
- could be easily extended to build ebpf programs in a container
- ebpf progs can be built independently
- simplifies build.rs (maybe :)
- can make it fairly easy to switch to GCC instead of clang

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -115,7 +115,7 @@ functional_task:
     - ssh ${DISTRO} 'sh -exc "uname -a"'
 
   test_script: &functional_test
-    - ssh -tt ${DISTRO} 'bash --login -exc "cd /vagrant/tests && cargo build"'
+    - ssh -tt ${DISTRO} 'bash --login -exc "cd /vagrant && cargo build"'
     - ssh -tt ${DISTRO} "cd /vagrant/tests && sudo python3 -m pytest ${TEST_EXTRA_ARGS}"
 
 # Manual trigger for non-PR branches.
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -11,22 +11,22 @@ cargo_task_tempate: &cargo_task_template
       - cat Cargo.lock
   setup_script:
     - apt-get update
-    - apt-get -y install llvm clang libelf-dev libpcap-dev python3-pip
+    - apt-get -y install llvm clang libelf-dev libpcap-dev python3-pip make jq
     - rustup component add rustfmt
     - rustup component add clippy
   before_cache_script: rm -rf $CARGO_HOME/registry/index
 
 unittest_task:
   << : *cargo_task_template
-  build_script: cargo build --verbose
-  test_script: cargo test --verbose
+  build_script: make V=1 CARGO_OPTS=--verbose
+  test_script: make test V=1 CARGO_OPTS=--verbose
   check_script:
     - cargo fmt --check
     - cargo clippy -- -D warnings
 
 benchmark_task:
   << : *cargo_task_template
-  build_script: cargo build -F benchmark --verbose --release
+  build_script: make bench V=1 CARGO_OPTS=--verbose
   test_script:
     - ./target/release/retis benchmark --ci events_parsing
     - ./target/release/retis benchmark --ci events_output
@@ -115,7 +115,7 @@ functional_task:
     - ssh ${DISTRO} 'sh -exc "uname -a"'
 
   test_script: &functional_test
-    - ssh -tt ${DISTRO} 'bash --login -exc "cd /vagrant && cargo build"'
+    - ssh -tt ${DISTRO} 'bash --login -exc "cd /vagrant && make V=1"'
     - ssh -tt ${DISTRO} "cd /vagrant/tests && sudo python3 -m pytest ${TEST_EXTRA_ARGS}"
 
 # Manual trigger for non-PR branches.
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -26,10 +26,10 @@ unittest_task:
 
 benchmark_task:
   << : *cargo_task_template
-  build_script: cargo build -F benchmark --verbose
+  build_script: cargo build -F benchmark --verbose --release
   test_script:
-    - ./target/debug/retis benchmark --ci events_parsing
-    - ./target/debug/retis benchmark --ci events_output
+    - ./target/release/retis benchmark --ci events_parsing
+    - ./target/release/retis benchmark --ci events_output
   check_script:
     - cargo clippy -F benchmark -- -D warnings
 
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,9 @@ export CLANG := clang
 export OBJCOPY := llvm-objcopy
 
 CARGO := cargo
+RELEASE_VERSION ?= $(shell $(CARGO) metadata --no-deps --format-version=1 | jq -r '.packages | .[] | select(.name == "retis") | .version')
+RELEASE_NAME ?= $(shell $(CARGO) metadata --no-deps --format-version=1 | jq '.packages | .[] | select(.name=="retis") | .metadata.misc.release_name')
+
 PRINT = echo
 
 VERBOSITY := $(filter 1,$(V))
@@ -73,27 +76,30 @@ else
     CARGO_JOBS := 1
 endif
 
-all: build
+all: debug
 
-install: build
-	$(CARGO) $(CARGO_OPTS) install $(CARGO_INSTALL_OPTS)
+install: release
+	$(CARGO) $(CARGO_OPTS) install $(CARGO_INSTALL_OPTS) --path=.
 
-build: ebpf
-	$(call out_console,CARGO,building retis ...)
+define build
+	$(call out_console,CARGO,$(strip $(2)) ...)
 	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
-	RETIS_PKG_VERSION=$(RELEASE_VERSION) \
+	RETIS_RELEASE_VERSION=$(RELEASE_VERSION) \
 	RETIS_RELEASE_NAME=$(RELEASE_NAME) \
-	$(CARGO) $(CARGO_OPTS) build $(CARGO_CMD_OPTS)
+	$(CARGO) $(CARGO_OPTS) $(1) $(CARGO_CMD_OPTS)
+endef
+
+debug: ebpf
+	$(call build, build, building retis (debug))
+
+release: ebpf
+	$(call build, build --release, building retis (release))
 
 test: ebpf
-	$(call out_console,CARGO,running tests ...)
-	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
-	$(CARGO) $(CARGO_OPTS) test $(CARGO_CMD_OPTS)
+	$(call build, test, building and running tests)
 
 bench: ebpf
-	$(call out_console,CARGO,building benchmarks ...)
-	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
-	$(CARGO) $(CARGO_OPTS) build -F benchmark --release $(CARGO_CMD_OPTS)
+	$(call build, build -F benchmark --release, building benchmarks)
 
 ifeq ($(NOVENDOR),)
 $(LIBBPF_INCLUDES): $(LIBBPF_SYS_LIBBPF_INCLUDES)
@@ -134,6 +140,7 @@ help:
 	$(PRINT) '	                  (eBPF only).'
 	$(PRINT) 'ebpf                --  Builds only the eBPF programs.'
 	$(PRINT) 'install             --  Installs Retis.'
+	$(PRINT) 'release             --  Builds Retis with the release option.'
 	$(PRINT) 'test                --  Builds and runs unit tests.'
 	$(PRINT)
 	$(PRINT) 'Optional variables that can be used to override the default behavior:'
@@ -150,4 +157,4 @@ help:
 	$(PRINT) 'NOVENDOR            --  Avoid to self detect and consume the vendored headers'
 	$(PRINT) '                        shipped with libbpf-sys.'
 
-.PHONY: all clean clean-ebpf ebpf $(EBPF_PROBES) $(GENERIC_HOOKS) help install $(OVS_HOOKS)
+.PHONY: all bench clean clean-ebpf ebpf $(EBPF_PROBES) $(GENERIC_HOOKS) help install $(OVS_HOOKS) release test
--- a/Makefile
+++ b/Makefile
@@ -73,12 +73,27 @@ else
     CARGO_JOBS := 1
 endif
 
-all: ebpf
+all: build
+
+install: build
+	$(CARGO) $(CARGO_OPTS) install $(CARGO_INSTALL_OPTS)
+
+build: ebpf
 	$(call out_console,CARGO,building retis ...)
 	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
 	RETIS_PKG_VERSION=$(RELEASE_VERSION) \
 	RETIS_RELEASE_NAME=$(RELEASE_NAME) \
 	$(CARGO) $(CARGO_OPTS) build $(CARGO_CMD_OPTS)
+
+test: ebpf
+	$(call out_console,CARGO,running tests ...)
+	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
+	$(CARGO) $(CARGO_OPTS) test $(CARGO_CMD_OPTS)
+
+bench: ebpf
+	$(call out_console,CARGO,building benchmarks ...)
+	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
+	$(CARGO) $(CARGO_OPTS) build -F benchmark --release $(CARGO_CMD_OPTS)
 
 ifeq ($(NOVENDOR),)
 $(LIBBPF_INCLUDES): $(LIBBPF_SYS_LIBBPF_INCLUDES)
@@ -112,23 +127,27 @@ clean: clean-ebpf
 
 help:
 	$(PRINT) 'all                 --  Builds the tool (both eBPF programs and retis).'
+	$(PRINT) 'bench               --  Builds benchmarks.'
 	$(PRINT) 'clean               --  Deletes all the files generated during the build process'
 	$(PRINT) '	                  (eBPF and rust directory).'
 	$(PRINT) 'clean-ebpf          --  Deletes all the files generated during the build process'
 	$(PRINT) '	                  (eBPF only).'
 	$(PRINT) 'ebpf                --  Builds only the eBPF programs.'
+	$(PRINT) 'install             --  Installs Retis.'
+	$(PRINT) 'test                --  Builds and runs unit tests.'
 	$(PRINT)
 	$(PRINT) 'Optional variables that can be used to override the default behavior:'
 	$(PRINT) 'V                   --  If set to 1 the verbose output will be printed.'
 	$(PRINT) '                        cargo verbosity is set to default.'
-	$(PRINT) '                        To override `cargo` behavior please refer to $$(CARGO_OPTS)'
-	$(PRINT) '                        and $$(CARGO_CMD_OPTS).'
+	$(PRINT) '                        To override `cargo` behavior please refer to $$(CARGO_OPTS),'
+	$(PRINT) '                        $$(CARGO_CMD_OPTS) and for the install $$(CARGO_INSTALL_OPTS).'
 	$(PRINT) '                        For further `cargo` customization please refer to configuration'
 	$(PRINT) '                        environment variables'
 	$(PRINT) '                        (https://doc.rust-lang.org/cargo/reference/environment-variables.html).'
 	$(PRINT) 'CARGO_CMD_OPTS      --  Changes `cargo` subcommand default behavior (e.g. --release for `build`).'
+	$(PRINT) 'CARGO_INSTALL_OPTS  --  Changes `cargo` install subcommand default behavior.'
 	$(PRINT) 'CARGO_OPTS          --  Changes `cargo` default behavior (e.g. --verbose).'
 	$(PRINT) 'NOVENDOR            --  Avoid to self detect and consume the vendored headers'
 	$(PRINT) '                        shipped with libbpf-sys.'
 
-.PHONY: all clean clean-ebpf ebpf $(EBPF_PROBES) $(GENERIC_HOOKS) help $(OVS_HOOKS)
+.PHONY: all clean clean-ebpf ebpf $(EBPF_PROBES) $(GENERIC_HOOKS) help install $(OVS_HOOKS)
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,134 @@
+ROOT_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+export LLC := llc
+export CLANG := clang
+export OBJCOPY := llvm-objcopy
+
+CARGO := cargo
+PRINT = echo
+
+VERBOSITY := $(filter 1,$(V))
+
+ifeq ($(VERBOSITY),)
+    Q=@
+    MAKE += -s
+    CARGO += -q
+define out_console
+    $(PRINT) -e "[$(1)]\t$(2)"
+endef
+
+.SILENT:
+endif
+
+ifeq ($(NOVENDOR),)
+    # This MUST be kept in sync with API_HEADERS under lib.rs in libbpf-sys
+    LIBBPF_API_HEADERS := bpf.h \
+                          libbpf.h \
+                          btf.h \
+                          bpf_helpers.h \
+                          bpf_helper_defs.h \
+                          bpf_tracing.h \
+                          bpf_endian.h \
+                          bpf_core_read.h \
+                          libbpf_common.h \
+                          usdt.bpf.h
+
+    LIBBPF_SYS_LIBBPF_BASE_PATH := $(dir $(shell cargo metadata --format-version=1 | jq -r '.packages | .[] | select(.name == "libbpf-sys") | .manifest_path'))
+    LIBBPF_SYS_LIBBPF_INCLUDES :=  $(wildcard $(addprefix $(LIBBPF_SYS_LIBBPF_BASE_PATH)/libbpf/src/, $(LIBBPF_API_HEADERS)))
+    LIBBPF_INCLUDES := $(ROOT_DIR)/src/.out
+endif
+
+FILTER_INCLUDES := src/core/filters/packets/bpf/include \
+                   src/core/filters/meta/bpf/include
+# Taking errno.h from libc instead of linux headers.
+# TODO: Remove when we fix proper header dependencies.
+INCLUDES_ALL := $(abspath $(wildcard src/core/probe/bpf/include \
+                                     src/core/probe/kernel/bpf/include \
+                                     src/core/probe/user/bpf/include \
+                                     src/core/events/bpf/include \
+                                     src/core/tracking/bpf/include \
+                                     /usr/include/x86_64-linux-gnu \
+                                     $(FILTER_INCLUDES)))
+INCLUDES_ALL += $(LIBBPF_INCLUDES)
+
+OVS_INCLUDES := $(abspath src/module/ovs/bpf/include)
+INCLUDES := $(addprefix -I, $(INCLUDES_ALL))
+
+EBPF_PROBES := $(abspath src/core/probe/kernel/bpf \
+                        src/core/probe/user/bpf)
+
+GENERIC_HOOKS := $(abspath src/module/skb/bpf \
+                           src/module/skb_drop/bpf \
+                           src/module/skb_tracking/bpf \
+                           src/module/nft/bpf \
+                           src/module/ct/bpf)
+
+OVS_HOOKS := $(abspath src/module/ovs/bpf)
+OUT_NAME := HOOK
+
+JOBS := $(patsubst -j%,%,$(filter -j%,$(MAKEFLAGS)))
+
+ifneq ($(JOBS),)
+    CARGO_JOBS := $(JOBS)
+else
+    CARGO_JOBS := 1
+endif
+
+all: ebpf
+	$(call out_console,CARGO,building retis ...)
+	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
+	RETIS_PKG_VERSION=$(RELEASE_VERSION) \
+	RETIS_RELEASE_NAME=$(RELEASE_NAME) \
+	$(CARGO) $(CARGO_OPTS) build $(CARGO_CMD_OPTS)
+
+ifeq ($(NOVENDOR),)
+$(LIBBPF_INCLUDES): $(LIBBPF_SYS_LIBBPF_INCLUDES)
+	-mkdir -p $(LIBBPF_INCLUDES)/bpf
+	cp $^ $(LIBBPF_INCLUDES)/bpf/
+endif
+
+$(OVS_HOOKS): INCLUDES_EXTRA := -I$(OVS_INCLUDES)
+
+$(EBPF_PROBES): OUT_NAME := PROBE
+
+ebpf: $(EBPF_PROBES) $(GENERIC_HOOKS) $(OVS_HOOKS)
+
+$(EBPF_PROBES) $(GENERIC_HOOKS) $(OVS_HOOKS): $(LIBBPF_INCLUDES)
+	$(call out_console,$(OUT_NAME),building $@ ...)
+	CFLAGS="$(INCLUDES) $(INCLUDES_EXTRA)" \
+	$(MAKE) -r -f $(ROOT_DIR)/ebpf.mk -C $@ $(TGT)
+
+clean-ebpf:
+	$(call out_console,CLEAN,cleaning ebpf progs...)
+	for i in $(EBPF_PROBES) $(GENERIC_HOOKS) $(OVS_HOOKS); do \
+	    $(MAKE) -r -f $(ROOT_DIR)/ebpf.mk -C $$i clean; \
+	done
+	-if [ -n "$(LIBBPF_INCLUDES)" ]; then \
+	    rm -rf $(LIBBPF_INCLUDES); \
+	fi
+
+clean: clean-ebpf
+	$(call out_console,CLEAN,cleaning retis ...)
+	$(Q)$(CARGO) clean
+
+help:
+	$(PRINT) 'all                 --  Builds the tool (both eBPF programs and retis).'
+	$(PRINT) 'clean               --  Deletes all the files generated during the build process'
+	$(PRINT) '	                  (eBPF and rust directory).'
+	$(PRINT) 'clean-ebpf          --  Deletes all the files generated during the build process'
+	$(PRINT) '	                  (eBPF only).'
+	$(PRINT) 'ebpf                --  Builds only the eBPF programs.'
+	$(PRINT)
+	$(PRINT) 'Optional variables that can be used to override the default behavior:'
+	$(PRINT) 'V                   --  If set to 1 the verbose output will be printed.'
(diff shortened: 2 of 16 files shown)


vlrpl:
> Overall, I think this is a good idea. Some high level comments:
> 
>     * Would be nice not to have two makefiles, removing `common.mk`.
> 

My idea, when adding common.mk, was to avoid to have a single root Makefile or one Makefile for each tgt directory (basically duplicated).
can you elaborate a bit on what you mean with 2 Makefiles?

>     * We could take the opportunity to set the release version + name in the `Makefile`. That would allow to build custom builds with a custom version/name.
> 

yup, that part is out ATM, but should be fairly easy to introduce.
In general, given the "ebpf" target, there should not be blocking parts (e.g. we can always issue make ebpf && cargo ...).

>     * We should rework what ends up in stdout when building BPF parts, a bit like what the kernel does.
> 

right, like quiet vs verbose, I suppose

>     * Collector hooks have to be defined twice. What if we could only define them in the `Makefile` and parse it from `build.rs` to avoid duplication? Alternatively an external file is probalby OK (but IMHO less friendly).
> 

I was thinking about env, but it doesn't really work as I think we should always keep a friendly support for the ebpf target + cargo build pair.
I think we can split this part of Makefile and put it into some `.hooks.mk `. 
Makefile includes it, while `build.rs` uses it to feed itself with hooks paths (probably some additional work could be needed). I don't love it, but at the moment I can't think of a better alternative. WDYT?

>     * `gen_probe_skel` logic should be in split from `gen_skel`. A little code duplication is OK for clarity.

yes, changing the previous bullet we can align things with this one in mind

vlrpl:
although beneficial, this doesn't seem a high priority.
Almost everything is there, but let's postpone this activity (there's one crucial point completely undiscussed still hanging).

vlrpl:
reviving this as we have consensus around it.

Makefile
```diff
@@ -10,6 +10,8 @@
 #                shipped with libbpf-sys
 #
 ROOT_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+RELEASE_VERSION ?= 1.3.0
```
vlrpl, line 13: not sure this is what was intended in the feedback (some time passed), but I put this together as a potential solution and added it here. Of course, I kept it split so it's more convenient to drop it and experiment with it.
Although we now have the version here, the manifest still requires it to be set.
Feedback is very welcome.

.cirrus.yml
```diff
@@ -11,22 +11,22 @@ cargo_task_tempate: &cargo_task_template
...
+  build_script: CARGO_BUILD_OPTS=-v make V=1
   test_script: cargo test --verbose
   check_script:
     - cargo fmt --check
     - cargo clippy -- -D warnings
 
 benchmark_task:
   << : *cargo_task_template
-  build_script: cargo build -F benchmark --verbose
+  build_script: make ebpf V=1 && cargo build -F benchmark --verbose
```
atenart, line 29: I think we should support every common target in the Makefile directly, at least `test` and probably `benchmark`. Otherwise the `build` vs other cargo targets makes things inconsistent and let the user think using cargo directly is supported; whereas it's now a low-level tool.

Same applies for `cargo build --release`, which should have instead a `release` target in the Makefile.

Having said that, it's very nice to have an `ebpf` target and be able to use cargo, but as low level tools.

Makefile
```diff
@@ -0,0 +1,125 @@
...
+export OBJCOPY := llvm-objcopy
+
+PRINT=echo
+
+ifeq ($(V),)
+	Q=@
+	MAKE += -s
+	suppress=$(Q)>/dev/null 2>&1 
+define out_console
+	$(PRINT) -e "\033[33m [$(1)] $(2)\033[0m"
```
atenart, line 27: Nit: we can use a tab to have the output better aligned. Something like `"\033[33m $(1)\t$(2)\033[0m"`.

Also not sure about using a color. It's nice, but depending the ones' terminal bg, this can be unreadable.

atenart (changes requested):
I think it's going in the right directly, mostly high level comments below.

Makefile
```diff
@@ -0,0 +1,125 @@
...
+export LLC := llc
+export CLANG := clang
+export OBJCOPY := llvm-objcopy
+
+PRINT=echo
+
+ifeq ($(V),)
+	Q=@
+	MAKE += -s
+	suppress=$(Q)>/dev/null 2>&1
```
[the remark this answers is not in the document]
vlrpl: makes sense. Will take a look.

Makefile
```diff
@@ -0,0 +1,125 @@
+# Two main optional variables are required:
+# - $(V): set or unset. If set the verbose output will be printed.
+#         cargo verbosity is set to default. To override `cargo build`
+#         behaviour please refer to $(CARGO_COMMON_OPTS) for common
+#         options and $(CARGO_BUILD_OPTS) for build options.
+#         For further `cargo` customization please refer to
+#         configuration environment variables
+#         (https://doc.rust-lang.org/cargo/reference/environment-variables.html).
+# - $(NOVENDOR): avoid to self detect and consume the vendored headers
+#                shipped with libbpf-sys
```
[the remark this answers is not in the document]
vlrpl: agreed. It's something we might even want to expand in the future

.cirrus.yml
vlrpl: yes, that's my idea as well (for the final result)

Makefile
vlrpl: I agree. I would happily avoid adding any kind of detection or similar things. The best option is to remove it.

Makefile
atenart: I think this comes from an attempt to have a version & name set based on the current git tree status and not based on the cargo target (dev vs release). Eg.
- `retis vx.y.x "name"` if the tree is pointing to a vx.y.z tag.
- `retis vx.y.z+` otherwise.

This can only be done by using git and looking at the current tree status, was experimented in `build.rs` but that turned out to be impossible to do. Now with a `Makefile` that might be possible.

Ideally we would have the version and name only defined in the `Makefile`, but I had a look at this and we can't use env variables in `Cargo.toml`. This means we would have to extract the version automatically from `Cargo.toml` (so we don't have to maintain two copies) but still set the version name in `Makefile`, which is not perfectly clean.

vlrpl: you mean something like this?

`cargo metadata --format-version=1 | jq '.packages | .[] | select(.name=="retis") | .version'`

atenart: If we agree to continue in this direction, yes, that would be nice to avoid duplicating the version in two files.

vlrpl: I updated my series (and also found a way for the `release_name`) just to make it stand out and get the chance to discuss about it

.cirrus.yml
vlrpl: about cargo release, I'm not sure we need a separate target.
I think we can simply do something like:

`make CARGO_BUILD_OPTS=--release V=1 -j$(nproc)`
instead of:
`make release V=1 -j$(nproc)`

WDYT?

common.mk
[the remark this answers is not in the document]
vlrpl: I see. Normally local Makefile are put under the target directory, whereas in this case, it was possible to tackle everything in a single common.mk (so instead of having empty Makefiles including a single one we use directly the separate one `make -C -f ...`).
I think that, based on the needs, we could eventually end up having local Makefiles (maybe with the majority simply including common.mk) with custom logic. Having a single one for the whole project could become a bit hard to handle if we need to approach some targets differently.

The second point here is that having the second (or local Makfiles) changes a bit the way we handle paths (we CWD) and so the Makefile itself. I think it could end up requiring decent changes (didn't look into it, but I guess some targets in common.mk cannot be simply copied, e.g. pattern rules, and this may or may not require some additional work which may lead to a more "static" way of handling subdirs).

.cirrus.yml
atenart: > Just for testing, I verified that even more complex commands still work (although can be run separately): `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER=sudo RUST_BACKTRACE=1 make test CARGO_CMD_OPTS="--features=test_cap_bpf,debug core::probe::kernel::raw_tracepoint::tests::init_and_attach -- --nocapture --show-output"`

Nice!

> about cargo release, I'm not sure we need a separate target. I think we can simply do something like:
> 
> `make CARGO_BUILD_OPTS=--release V=1 -j$(nproc)` (in the latest iteration should be `s/CARGO_BUILD_OPTS/CARGO_CMD_OPTS/`) instead of: `make release V=1 -j$(nproc)`

While I agree `--release` is just a flag we can give to Cargo, it's really a target we use a lot and IMO it's easier and more user friendly to maintain it as a target.

Makefile
atenart: With both variables defined in `Cargo.toml` this is quite nice!

.cirrus.yml
vlrpl: release target is now part of the series

common.mk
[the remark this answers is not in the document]
vlrpl: just to make it clear, my preference here is to keep these files split (one global and one for probes/hooks that could be converted in a per-directory Makefile, if needed). Not sure this is a blocker for this PR.

Makefile
```diff
@@ -0,0 +1,154 @@
...
+	CARGO += -q
+define out_console
+	$(PRINT) -e "[$(1)]\t$(2)"
+endef
+
+.SILENT:
+endif
+
+ifeq ($(NOVENDOR),)
+    # This MUST be kept in sync with API_HEADERS under lib.rs in libbpf-sys
```
atenart, line 25: That looks a bit dangerous. What about:

```
LIBBPF_API_HEADERS := $(shell sed -n "s/header!(\"\(.*\)\"),/\1/p" $(LIBBPF_SYS_LIBBPF_BASE_PATH)/src/lib.rs)
```

(untested)

Makefile
```diff
@@ -0,0 +1,154 @@
...
+                          bpf_tracing.h \
+                          bpf_endian.h \
+                          bpf_core_read.h \
+                          libbpf_common.h \
+                          usdt.bpf.h
+
+    LIBBPF_SYS_LIBBPF_BASE_PATH := $(dir $(shell cargo metadata --format-version=1 | jq -r '.packages | .[] | select(.name == "libbpf-sys") | .manifest_path'))
+    LIBBPF_SYS_LIBBPF_INCLUDES :=  $(wildcard $(addprefix $(LIBBPF_SYS_LIBBPF_BASE_PATH)/libbpf/src/, $(LIBBPF_API_HEADERS)))
+    LIBBPF_INCLUDES := $(ROOT_DIR)/src/.out
+    LIBBPF_INCLUDE_FILES := $(addprefix $(LIBBPF_INCLUDES)/bpf/,$(LIBBPF_API_HEADERS))
```
atenart, line 40: This doesn't seem to be used.

Makefile
```diff
@@ -0,0 +1,154 @@
...
+FILTER_INCLUDES := src/core/filters/packets/bpf/include \
+                   src/core/filters/meta/bpf/include
+INCLUDES_ALL := $(abspath $(wildcard src/core/probe/bpf/include \
+                                     src/core/probe/kernel/bpf/include \
+                                     src/core/probe/user/bpf/include \
+                                     src/core/events/bpf/include \
+                                     src/core/tracking/bpf/include \
+                                     /usr/include/x86_64-linux-gnu \
+                                     vendor/linux/include \
+                                     vendor/linux/asm/x86/include \
```
atenart, line 52: What are those paths?

Makefile
```diff
@@ -0,0 +1,154 @@
...
+	$(CARGO) $(CARGO_COMMON_OPTS) test $(CARGO_CMD_OPTS)
+
+bench: ebpf
+	$(call out_console,CARGO,building benchmarks ...)
+	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
+	$(CARGO) $(CARGO_COMMON_OPTS) build -F benchmark $(CARGO_CMD_OPTS)
+
+ifeq ($(NOVENDOR),)
+$(LIBBPF_INCLUDES): $(LIBBPF_SYS_LIBBPF_INCLUDES)
+	-mkdir -p $(LIBBPF_INCLUDES)/bpf
```
atenart, line 106: This is not removed in `clean`.

Makefile
```diff
@@ -0,0 +1,154 @@
...
+    LIBBPF_INCLUDE_FILES := $(addprefix $(LIBBPF_INCLUDES)/bpf/,$(LIBBPF_API_HEADERS))
+endif
+
+FILTER_INCLUDES := src/core/filters/packets/bpf/include \
+                   src/core/filters/meta/bpf/include
+INCLUDES_ALL := $(abspath $(wildcard src/core/probe/bpf/include \
+                                     src/core/probe/kernel/bpf/include \
+                                     src/core/probe/user/bpf/include \
+                                     src/core/events/bpf/include \
+                                     src/core/tracking/bpf/include \
```
atenart, line 49: Maybe we could auto-detect all those `src/*/bpf/include` paths?

Makefile
```diff
@@ -0,0 +1,154 @@
...
+INCLUDES := $(addprefix -I, $(INCLUDES_ALL))
+
+EBPF_PROBES := $(abspath src/core/probe/kernel/bpf \
+                        src/core/probe/user/bpf)
+
+GENERIC_HOOKS := $(abspath src/module/skb/bpf \
+                           src/module/skb_drop/bpf \
+                           src/module/skb_tracking/bpf \
+                           src/module/nft/bpf \
+                           src/module/ct/bpf)
```
atenart, line 66: Maybe we could auto-detect all those `src/module/*/bpf/*.bpf.c` files?

Makefile
```diff
@@ -0,0 +1,154 @@
...
+
+test: ebpf
+	$(call out_console,CARGO,running tests ...)
+	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
+	$(CARGO) $(CARGO_COMMON_OPTS) test $(CARGO_CMD_OPTS)
+
+bench: ebpf
+	$(call out_console,CARGO,building benchmarks ...)
+	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
+	$(CARGO) $(CARGO_COMMON_OPTS) build -F benchmark $(CARGO_CMD_OPTS)
```
atenart, line 102: \+ `--release`

Makefile
```diff
@@ -0,0 +1,154 @@
...
+
+ifneq ($(JOBS),)
+    CARGO_JOBS := $(JOBS)
+else
+    CARGO_JOBS := 1
+endif
+
+all: build
+
+install: build
```
atenart, line 81: I don't think it changes anything, but for clarity (and potential future changes) I think it should depend on `release` instead of `build`.

Makefile
```diff
@@ -0,0 +1,154 @@
...
+	$(CARGO) $(CARGO_COMMON_OPTS) install $(CARGO_CMD_OPTS)
+
+release: CARGO_CMD_OPTS += --release
+release: build
+
+build: ebpf
+	$(call out_console,CARGO,building retis ($(if $(findstring --release,$(CARGO_CMD_OPTS)),release,debug))...)
+	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
+	RETIS_PKG_VERSION=$(RELEASE_VERSION) \
```
atenart, line 90: Nit: s/RETIS_PKG_VERSION/RETIS_RELEASE_VERSION/

I think the name/version should be set for all targets. But we should add logic (could be done later) to detect if HEAD points to the right version tag otherwise add `+` at the end of the version. Same, if uncommited changes are detected we should add `-dirty`.

common.mk
```diff
@@ -0,0 +1,33 @@
+ARCH ?= x86
```
atenart, line 1: How does this work with cargo and the Rust compilation?

common.mk
```diff
@@ -0,0 +1,33 @@
...
+all: $(ALL_REQ)
+
+$(OUT_DIR):
+	mkdir -p $(OUT_DIR)/
+
+$(OBJS): | $(OUT_DIR)
+
+$(OUT_DIR)/%.o: %.c
+	$(CLANG) $(CFLAGS) $(BPF_CFLAGS) $(INCLUDES_EXTRA) -MMD -c -g -o $@ $<
+	$(OBJCOPY) --strip-debug $@
```
atenart, line 26: Maybe only for `release`? Or is that never useful?

atenart (changes requested):
Overall on the functional level this looked good. Dived into more details for this review; a few high level comments:
- Any change required in the docs regarding the prerequisites to build Retis?
- `tools/benchmark_*` scripts need to be updated to use `make`.

common.mk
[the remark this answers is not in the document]
atenart: I agree, no need to put a Makefile in each eBPF folder for now. For the Makefile/common.mk split, it's not a blocker. My personal preference would be not to split it, but it's your choice. If you keep the logic in two different files, could you rename `common.mk` to `ebpf.mk`?

Makefile
```diff
@@ -0,0 +1,154 @@
...
+	-if [ -n "$(LIBBPF_INCLUDES)" ]; then \
+	    rm -rf $(LIBBPF_INCLUDES); \
+	fi
+	$(call out_console,CLEAN,cleaning retis ...)
+	$(Q)$(CARGO) clean 
+
+help:
+	$(PRINT) 'all                 --  Builds the tool (both eBPF programs and retis)'
+	$(PRINT) 'bench               --  Builds benchmarks'
+	$(PRINT) 'clean               --  Deletes all the file generated during the build process'
```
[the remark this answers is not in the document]
vlrpl: makes sense

Makefile
vlrpl: it probably remained there after waltzing with vars target and prereq

Makefile
vlrpl: They come from `91510e8f4cdc ("module: ovs: introduce queue_id to track upcalls")`

Makefile
vlrpl: we could. This mostly follows the logic of the old `build.rs` (we went with explicit paths), the goal, of course, was to split eBPF prog build from rust build.
That said, detection establishes a naming convention. Probably, we already have it and so it's ok.
I'll take a look at this, but I'll add it in a separate patch (like any other out of PR-scope clean up).

Makefile
vlrpl: this means that `make` compiles with `--release`. Is that what we want?

Makefile
vlrpl: that's right. 
about `HEAD`, I agree, I would not leverage this PR for this. It's a good follow-up, though.

common.mk
vlrpl: cargo has its own env variables e.g. `CARGO_CFG_TARGET_ARCH`. It should not be affected by this. 
I think your point is a bit more than that. I think I got it, but can you please elaborate a bit here so we're sure we're not talking about different things? :)

common.mk
vlrpl: this is always useful. It strips dwarf info. BTF info is still there (func info and line info as well). 
You can tell the difference in the binary size (and skel creation) if you remove the line :)

with:
`110K Feb  3 12:35 ./core/probe/kernel/bpf/.out/kretprobe.bpf.o`

without:
`1.3M Feb  3 12:37 ./core/probe/kernel/bpf/.out/kretprobe.bpf.o`

common.mk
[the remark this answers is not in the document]
vlrpl: sgtm, thanks

Makefile
vlrpl: I was trying to avoid things like that because it's a bit fragile (if renamed or moved), but of course, feel free to disagree. 
The comment though is a bit misleading (it's not an all-or-nothing thing, but it's just what we need).
Here the only thing we should be careful of is that, if we have system headers and we miss exposing vendored ones, those would be "silently" used.

atenart: > Being library headers (local to the actual libbpf) they should remain as they are, so it should be safe. The comment though is a bit misleading (it's not an all-or-nothing thing, but it's just what we need). 

Makes sense, and is less scary.

> Here the only thing we should be careful of is that, if we have system wide headers and we miss exposing vendored ones, those would be "silently" used.

But still a bit :)

> I was trying to avoid things like that because it's a bit fragile (if renamed or moved or gets fed in a way that makes it fail), but of course, feel free to disagree.

Yeah, I agree, both ways are not perfect. Maybe one solution would be to keep the list hardcoded and use the above shell command to check it matches `libbpf-sys` definition? This way we can detect changes.

Makefile
atenart: But they are not used right?

Makefile
atenart: No, only for the `install` target.

common.mk
atenart: My question is can `ARCH` be set to something different and would that make the end result (BPF+Rust) compatible with that non x86 arch?

common.mk
atenart: Nice space saving. I was wondering if the unstripped version would be useful for debugging. My understanding of the above is it's not, so that's fine, thanks!

Makefile
```diff
@@ -0,0 +1,154 @@
...
+	$(PRINT) '	                  (eBPF and rust directory)'
+	$(PRINT) 'ebpf                --  Builds only the eBPF programs'
+	$(PRINT) 'install             --  Installs Retis'
+	$(PRINT) 'release             --  Builds Retis with the release option'
+	$(PRINT) 'test                --  Builds and runs unit tests'
+	$(PRINT)
+	$(PRINT) 'Optional variables that can be used to override the default behavior:'
+	$(PRINT) 'V                   --  If set the verbose output will be printed.'
```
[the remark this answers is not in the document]
vlrpl: how should we behave if it's set to anything else?

Makefile
vlrpl: > But they are not used right?

that's right. I don't recall/know if they are a leftover of a rework (will take a look).
@amorenoz, do you remember the original reason behind these include paths?

Makefile
vlrpl: should be removed here, isn't it?

```shell
	-if [ -n "$(LIBBPF_INCLUDES)" ]; then \
	    rm -rf $(LIBBPF_INCLUDES); \
	fi
```

Makefile
vlrpl: I guess this could be considered as part of this follow-up:

https://github.com/retis-org/retis/pull/237#discussion_r1477049620

Makefile
vlrpl: looking at our current CI, we don't add `--release`.
Maybe for this target, if we want a `benchmark` compiled as with the release flag we could use `$(CARGO_CMD_OPTS)`?

Makefile
vlrpl: > > Being library headers (local to the actual libbpf) they should remain as they are, so it should be safe. The comment though is a bit misleading (it's not an all-or-nothing thing, but it's just what we need).
> 
> Makes sense, and is less scary.
> 
> > Here the only thing we should be careful of is that, if we have system wide headers and we miss exposing vendored ones, those would be "silently" used.
> 
> But still a bit :)
> 

yeah, I have no perfect solution if compiled on a system with libbpf

> > I was trying to avoid things like that because it's a bit fragile (if renamed or moved or gets fed in a way that makes it fail), but of course, feel free to disagree.
> 
> Yeah, I agree, both ways are not perfect. Maybe one solution would be to keep the list hardcoded and use the above shell command to check it matches `libbpf-sys` definition? This way we can detect changes.

IMHO, I'm not sure having both would make it better.
In such a case, we'd know there's a discrepancy, but that doesn't mean our programs will be compiled against older headers. Not sure if a warning is enough of a failure is required (and if required, maybe it's better to go back to parsing `header!()`).

I guess for this as well, given there will be a follow-up for detections, we could defer this there as well (https://github.com/retis-org/retis/pull/237#discussion_r1477049620)

common.mk
vlrpl: > Nice space saving. I was wondering if the unstripped version would be useful for debugging. My understanding of the above is it's not, so that's fine, thanks!

yup

Makefile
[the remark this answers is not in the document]
atenart: The kernel has:

```
  make V=n   [targets] 1: verbose build
                       2: give reason for rebuild of target
                       V=1 and V=2 can be combined with V=12
```

Not sure this is applicable here. But idk, we might want one day to let cargo be verbose while hidding the BPF compilation details?

Makefile
atenart: > looking at our current CI, we don't add `--release`.

Well, this is a bug. The benchmark command really must be compiled as a release binary.

Makefile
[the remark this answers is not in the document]
vlrpl: I was about to propose

````Make
ifneq ($(VERBOSITY),0)
    Q=@
    MAKE += -s
    CARGO += -q
define out_console
    $(PRINT) -e "[$(1)]\t$(2)"
endef

.SILENT:
else
    ifeq ($(V),2)
        CARGO_CMD_OPTS += --verbose
    endif
endif
```

- V unset quiet
- set to 1: verbose (cargo not verbose)
- set to 2: V=1 + cargo build --verbose

common.mk
atenart: > this is for `PT_REGS_PARMx()` as they depend on the target.

I see, makes sense.

>  In general cross-compilation support should be possible, but should require pointing to the right backends (should not be a big deal).

Yeah that's fine, we don't support it for now. So maybe instead of `ARCH ?= ` we should use `ARCH = ` to make it clear we don't support anything else. Also a comment might be worth it (that some macros depends on this being set).

Makefile
atenart: > > > I was trying to avoid things like that because it's a bit fragile (if renamed or moved or gets fed in a way that makes it fail), but of course, feel free to disagree.
> > 
> > 
> > Yeah, I agree, both ways are not perfect. Maybe one solution would be to keep the list hardcoded and use the above shell command to check it matches `libbpf-sys` definition? This way we can detect changes.
> 
> IMHO, I'm not sure having both would make it better. In such a case, we'd know there's a discrepancy, but that doesn't mean our programs will be compiled against older headers.

We can make the build fail. Yes the build might succeed and the output might be OK, but IMO in such case the build should be fixed so that it has the right expectations. Or am I missing a use case?

> I guess for this as well, given there will be a follow-up for detections, we could defer this there as well ([#237 (comment)](https://github.com/retis-org/retis/pull/237#discussion_r1477049620)).
> 
> WDYT?

Sure, we can do this by steps :)

Makefile
atenart: Oops, sorry for the noise.

Makefile
vlrpl: ok. Dangerous context was close :)

Makefile
vlrpl: had a look and didn't find any particular reason they were present. My bet is they were included as part of a different approach and slipped out. Anyways, given we never use them I removed both.

vlrpl:
> Overall on the functional level this looked good. Dived into more details for this review.
> 
> A few high level comments first:
> 
>     * Any change required in the docs regarding the prerequisites to build Retis?
> 
>     * `tools/benchmark_*` scripts need to be updated to use `make`.

both needed an update, thanks.

Makefile
```diff
@@ -0,0 +1,160 @@
...
+
+test: ebpf
+	$(call out_console,CARGO,running tests ...)
+	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
+	$(CARGO) $(CARGO_OPTS) test $(CARGO_CMD_OPTS)
+
+bench: ebpf
+	$(call out_console,CARGO,building benchmarks ...)
+	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
+	$(CARGO) $(CARGO_OPTS) build -F benchmark --release $(CARGO_CMD_OPTS)
```
atenart, line 104: Suggestion: I think the following would remove duplication and make sure we have all set the right way for every variant; eg. release name/version or cargo jobs # would always be set — future options too.

```make
all: debug

install: release
	$(CARGO) $(CARGO_OPTS) install $(CARGO_CMD_OPTS) --path=.

define build
	$(MAKE) ebpf
	$(call out_console,CARGO,$(2)...)
	$(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
	RETIS_RELEASE_VERSION=$(RELEASE_VERSION) \
	RETIS_RELEASE_NAME=$(RELEASE_NAME) \
endef

debug:
	$(call build, build, building retis (debug))

release:
	$(call build, build --release, building retis (release))

test:
	$(call build, test, building and running tests)

bench:
	$(call build, build -F benchmark --release, building benchmarks)
```

Makefile
```diff
@@ -0,0 +1,160 @@
...
+$(EBPF_PROBES): OUT_NAME := PROBE
+
+ebpf: $(EBPF_PROBES) $(GENERIC_HOOKS) $(OVS_HOOKS)
+
+$(EBPF_PROBES) $(GENERIC_HOOKS) $(OVS_HOOKS): $(LIBBPF_INCLUDES)
+	$(call out_console,$(OUT_NAME),building $@ ...)
+	CFLAGS="$(INCLUDES) $(INCLUDES_EXTRA)" \
+	$(MAKE) -r -f $(ROOT_DIR)/ebpf.mk -C $@ $(TGT)
+
+clean-ebpf:
```
atenart, line 123: Looks like it's missing from the help.

src/main.rs
```diff
@@ -21,7 +22,8 @@ use crate::{
 // Re-export derive macros.
 use retis_derive::*;
 
-const VERSION_NAME: &str = "bredele";
+static VERSION_NAME: Lazy<&str> =
+    Lazy::new(|| option_env!("RETIS_RELEASE_NAME").unwrap_or("unreleased"));
```
atenart, line 26: Sorry, didn't see this before. Shouldn't this be defined alongside the version in `cli.rs`? It was defined here historically as it seemed a better place for something we had to modify for each release, but thanks to your work it's now defined in Cargo.toml so we can safely move it. That would also save a once_cell use.

nit: I think it's safe to keep the version & name mandatory (another comment suggested to always set those variables).

Makefile
```diff
@@ -0,0 +1,160 @@
...
+ifneq ($(JOBS),)
+    CARGO_JOBS := $(JOBS)
+else
+    CARGO_JOBS := 1
+endif
+
+all: build
+
+install: release
+	$(CARGO) $(CARGO_OPTS) install $(CARGO_CMD_OPTS) --path=.
```
atenart, line 84: I don't think we can use the same `CARGO_CMD_OPTS` that is used for building. We could special case the above (`CARGO_INSTALL_OPTS`) or remove it.

atenart (changes requested):
Still some comments, but I'm confident we'll converge soon.

Makefile
```diff
@@ -0,0 +1,160 @@
...
+endif
+
+FILTER_INCLUDES := src/core/filters/packets/bpf/include \
+                   src/core/filters/meta/bpf/include
+INCLUDES_ALL := $(abspath $(wildcard src/core/probe/bpf/include \
+                                     src/core/probe/kernel/bpf/include \
+                                     src/core/probe/user/bpf/include \
+                                     src/core/events/bpf/include \
+                                     src/core/tracking/bpf/include \
+                                     /usr/include/x86_64-linux-gnu \
```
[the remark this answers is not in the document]
atenart: "this" meaning the last line only, `/usr/include/...`.

Makefile
```diff
@@ -0,0 +1,160 @@
...
+    MAKE += -s
+    CARGO += -q
+define out_console
+    $(PRINT) -e "[$(1)]\t$(2)"
+endef
+
+.SILENT:
+else
+    ifeq ($(VERBOSITY),2)
+        CARGO_CMD_OPTS += --verbose
```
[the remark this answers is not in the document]
vlrpl: The same goes for release vs build + CARGO_OPTS.
That was an attempt to humor V=1,2 request :)

If the suggestion is still `V=1, V=2, V=12`, I think that any other further behavior finds better room in follow-ups otherwise the time for the PR stretches unreasonably adding more and more features/customizations (given in the old way of building we had no features like that).
Not sure we want to drop this V={1,2} part, though :D

Makefile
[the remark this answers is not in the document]
vlrpl: off the top of my head, I think this should be because of Ubuntu. There's a failure because we miss something (related to `errno.h`)

Makefile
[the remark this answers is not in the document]
atenart: > The same goes for release vs build + CARGO_OPTS.

A suggestion below would solve that ;)

> If the suggestion is still `V=1, V=2, V=12`, I think that any other further behavior finds better room in follow-ups otherwise the time for the PR stretches unreasonably adding more and more features/customizations (given in the old way of building we had no features like that). Not sure we want to drop this V={1,2} part, though :D

No please don't add more verbose levels :) As for V=2 it's up to you, if you think it has value (I truly have no idea; I guess for CI logs it's often nice to have more things logged in general).

Makefile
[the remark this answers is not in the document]
vlrpl: my bad :)
Will add it to the Makefile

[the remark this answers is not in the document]
atenart: Thanks!

Makefile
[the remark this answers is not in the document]
vlrpl: I don't think it hurts but at the same time, it's not super useful. Let me have a look at it (maybe we can sync offline)

Makefile
vlrpl: makes sense

Makefile
vlrpl: ah, thanks

Makefile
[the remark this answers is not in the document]
vlrpl: ok, so we went back to the initial approach. V=1 only this time, anything other than 1 defaults to quiet.

Makefile
vlrpl: I'm ok with factoring out some things, but I prefer to keep the prereq visible, hiding it in a macro call doesn't really make it as obvious as it should be

src/main.rs
vlrpl: yeah, among the many things to take care of in this PR I miss this. That was a good catch, though. It's time to remove this, indeed (not super neat, but the reasoning is/was clear).

vlrpl: I pushed an updated version keeping things as it was, but let's not consider this resolved until we reach consensus :)

Makefile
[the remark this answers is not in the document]
atenart: That works for me. The original comment was just about explicitly listing `1` as the expected value in the help.

Makefile
atenart: Works for me, that's the part I wasn't sure about :)

src/main.rs
atenart: > Jokes aside the reason of option_env!() over env!() is practical and it's about development. The split build (make ebpf; cargo build) is simply more convenient instead of being forced to specify both env variables.

Makes sense, the current status is good.

Makefile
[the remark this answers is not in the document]
atenart: Had a look at the latest version, `V=2` was removed but `--verbose` is added in `CARGO_OPTS`. Isn't that making a case for supporting `V=2`; if that is useful in a common use case?

Sorry if we're going circles on this... :(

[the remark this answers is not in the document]
vlrpl: > Had a look at the latest version, `V=2` was removed but `--verbose` is added in `CARGO_OPTS` in the CI. Isn't that making a case for supporting `V=2`; if that is useful in a common use case?
> 
> Sorry if we're going circles on this... :(

that was the "use case" I went/proposed that logic (and mostly because there weren't others really useful).
The point was that we used to build with a simple `cargo build --verbose` so I remapped this with the makefile logic.
I don't know, V=2 used in that way is not that useful, so I really have no strong preference on this.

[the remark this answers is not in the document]
atenart: Let's keep it this way then, we can add it back in the future if we every see the need.

atenart (approved):
LGTM, thanks, that really improves the development experience!

PR #246, opened 2023-09-12
core: inspect: remove os inspection helpers

atenart:
The os_id() and os_version() helpers aren't used atm and shouldn't be used as the information they report cannot be used in a reliable way. The use of containers is the main culprit, where the /etc/os-release file will report values different from the underlying system. This is not an issue when working with user programs in the container alone, but in our case (kernel, inspection of pids outside the container) those values do not mean anything valuable. Remove them.


vlrpl (approved):
code clean-up is always nice!
LGTM, thanks!

PR #247, opened 2023-09-13
Add a sanity CI test for skb tracking

atenart:


tests/conftest.py
```diff
@@ -83,6 +83,99 @@ def two_ports_skb(netns):
...
+    ns1.addr(
+        "add",
+        index=ns1.link_lookup(ifname="lo")[0],
+        address="10.0.255.1",
+        prefixlen=32,
+    )
+
+    # Enable forwarding and DNAT in ns1
+    netns.run("ns1", "sysctl", "-w", "net.ipv4.conf.all.forwarding=1")
+    netns.run(
```
vlrpl, line 158: not a blocking comment as this could be considered once we end up having an actual need, but I was wondering if it would make sense to split the fixture into multiple ones and have basic building blocks that could be reused. Like: 3ns + dnat + vip.

vlrpl: > Or even `ns(3) + dnat(1) + vip(1)`. Could be, but only if we find it more maintainable and if we see a real value in factorizing this. It's often tricky and tests setups are not easily split into reusable parts in my experience (like, there's always some small particularities that makes it special).

yes, no need to split now, as there's no actual need. Probably the ns part will be used in multiple tests. About the others, I agree that are more susceptible to slightly different characterizations.

The rest LGTM, so I'm going to approve.

vlrpl: cannot be approved for the cherry-picked commit. Let's just consider this approved by me :)

vlrpl: Not really, I must have seen grayed-out things when they were not.

PR #256, opened 2023-09-20
module: skb_drop: do sent empty event sections

atenart:
The previous fix wasn't complete and in case a kernel was supporting drop reasons but the hook was attached to a probe not having access to one, empty zeroed event sections were still reported.


vlrpl (approved):
lgtm

PR #262, opened 2023-09-26
skb tracking improvements / fixes

atenart:
Properly taking care of the `net:net_dev_xmit` tracepoint, which can access already freed skbs.

Before this PR, tracking entries were being removed by the tracking gc when using a probe on `net:net_dev_xmit`, like the `generic` profile does.


vlrpl (approved):
Nice catch. The series looks good to me.

PR #265, opened 2023-10-03
pcap-ng support

### LINKED ISSUE #224: PCAP output
Potential solutions need to be discussed, will provide more info later.

atenart:
Fixes #224. Marking as RFC to get avoid merging it while waiting for feedback; but code is in a good shape.

Two majors changes:

1. The `skb` module gained support for reporting raw packets from the BPF probes. This is disabled by default and enabled by using the `--skb-sections packet` option (note the recommendation and what is used in the new `pcap` profile is `packet,dev,ns`).
2. A post-processing command was added, `pcap`, to generate a `pcap-ng` formatted output (stdout or a file) from a *single* probe. This probe is selected at post-processing time (a single set of collected events can be used to generate different pcap outputs).

The pcap-ng output includes a definition of net devices seen in the events, including a name of the form `<name> (<netns>)` and a description of the form `ifindex=<i>`. A fake interface and/or netns is used if the event did not contained one (as the interface definition is mandatory in pcap-ng).

The pcap-ng packets include a comment of the form `probe=<probe def>`.

This works as follow:

```
$ retis -p pcap,generic collect -o retis.data --cmd 'dig google.com @8.8.8.8'

$ retis pcap --probe tp:net:netif_receive_skb -o retis.pcap retis.data
15:13:42 [INFO] 1 event(s) were processed
$ wireshark retis.pcap
...
$ tcpdump -nnr retis.pcap
reading from file retis.pcap, link-type EN10MB (Ethernet), snapshot length 65535
18:40:40.598012 IP 8.8.8.8.53 > 192.168.0.42.12345: 59700 1/0/1 A 216.58.214.78 (55)

$ retis pcap --probe kprobe:udp_rcv retis.data | tcpdump -nnr -
15:14:10 [INFO] 1 event(s) were processed
reading from file -, link-type EN10MB (Ethernet), snapshot length 65535
18:40:40.610832 IP 8.8.8.8.53 > 192.168.0.42.12345: 59700 1/0/1 A 216.58.214.78 (55)
$ retis pcap --probe tp:net:net_dev_start_xmit retis.data | tcpdump -nnr -
15:14:10 [INFO] 1 event(s) were processed
reading from file -, link-type EN10MB (Ethernet), snapshot length 65535
18:40:26.575099 IP 192.168.0.42.12345 > 8.8.8.8.53: 59700+ [1au] A? google.com. (51)
```

Currently at max 256 bytes are reported from the BPF probes. The exact value should be better defined, but the idea is to report headers from the packets, not necessarily their content. We could as a follow-up look into user-defined capture length.

All comments are welcomed, including non-code reviews.

### DIFF
--- a/src/module/skb/bpf/skb_hook.bpf.c
+++ b/src/module/skb/bpf/skb_hook.bpf.c
@@ -124,7 +124,7 @@ static __always_inline int process_skb_arp(struct retis_raw_event *event,
 					   struct skb_config *cfg,
 					   struct sk_buff *skb,
 					   unsigned char *head,
-					   u16 etype, u16 mac, u16 network)
+					   u16 mac, u16 network)
 {
 	struct skb_arp_event *e;
 	struct arphdr *arp;
@@ -182,7 +182,7 @@ static __always_inline int process_skb_ip(struct retis_raw_event *event,
 					  struct skb_config *cfg,
 					  struct sk_buff *skb,
 					  unsigned char *head,
-					  u16 etype, u16 mac, u16 network)
+					  u16 mac, u16 network)
 {
 	u8 protocol, ip_version;
 	u16 transport;
@@ -362,10 +362,10 @@ static __always_inline int process_skb_l2(struct retis_raw_event *event,
 
 	/* IPv4/IPv6 and upper layers */
 	if (etype == bpf_ntohs(ETH_P_IP) || etype == bpf_ntohs(ETH_P_IPV6))
-		return process_skb_ip(event, cfg, skb, head, etype, mac, network);
+		return process_skb_ip(event, cfg, skb, head, mac, network);
 	/* ARP */
 	else if (etype == bpf_ntohs(ETH_P_ARP))
-		return process_skb_arp(event, cfg, skb, head, etype, mac, network);
+		return process_skb_arp(event, cfg, skb, head, mac, network);
 
 	/* Unsupported etype */
 	return 0;
--- a/src/module/skb/bpf/skb_hook.bpf.c
+++ b/src/module/skb/bpf/skb_hook.bpf.c
@@ -5,6 +5,7 @@
 #include <common.h>
 
 #define BIT(x) (1 << (x))
+#define min(a, b) ((a) < (b) ? (a) : (b))
 
 #define ETH_P_IP	0x0800
 #define ETH_P_ARP	0x0806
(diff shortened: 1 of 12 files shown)


src/module/skb/skb.rs
```diff
@@ -56,7 +56,8 @@ impl Collector for SkbModule {
         let mut sections: u64 = 0;
         for category in args.skb_sections.iter() {
             match category.as_str() {
-                "all" => sections |= !0_u64,
+                // Do not include the raw packet in 'all'.
+                "all" => sections |= !SECTION_PACKET,
```
vlrpl, line 60: isn't this ~11?

src/module/skb/bpf/skb_hook.bpf.c
```diff
@@ -371,6 +382,85 @@ static __always_inline int process_skb_l2(struct retis_raw_event *event,
...
+	if (data_offset < 0) /* Keep the verifier happy */
+		return 0;
+
+	mac = BPF_CORE_READ(skb, mac_header);
+	network = BPF_CORE_READ(skb, network_header);
+	len = BPF_CORE_READ(skb, len);
+	linear_len = len - BPF_CORE_READ(skb, data_len); /* Linear buffer size */
+
+	/* Best case: mac offset is set and valid */
+	if (mac != (u16)~0U && (!network || (network && mac != network))) {
```
vlrpl, line 407: other places requiring valid and existing mac are simply considering `(mac != (u16)~0U)`, should we also extend (maybe the existing helper) the check for those as well?

src/process/cli/pcap.rs
```diff
@@ -0,0 +1,324 @@
...
+/// Generate a PCAP file from a collection of recorded events.
+#[derive(Parser, Debug, Default)]
+#[command(name = "pcap")]
+pub(crate) struct Pcap {
+    #[arg(
+        short,
+        long,
+        help = "Filter events from this probe. Probes should follow the TYPE:TARGET pattern. See `retis collect --help` for more details on the probe format."
+    )]
+    pub(super) probe: String,
```
vlrpl, line 212: apparently, wireshark supports filtering for this:
`frame.comment == "probe=raw_tracepoint:net:netif_receive_skb"`

Not explicitly asking for it, but it might be convenient in some cases to have all packets from different probes (although they could appear as duplicated) and then filter from the GUI. It would spare some back and forth.

vlrpl:
general question: should we add an entry in the doc for this?

vlrpl (approved):
This is a very nice feature.
The change lgtm!

PR #266, opened 2023-10-04
Default the event file to retis.data

atenart:
In both `collect -o` and in post-process commands, default the events files to `retis.data`.


vlrpl (approved):
the change lgtm, thanks!

PR #268, opened 2023-10-06
Cargo.toml: update dependencies

atenart:
Only one incompatible change: nix stopped enabling features by default, so had to enable the ones we use by hand.


vlrpl (approved):
lgtm, thanks!

PR #270, opened 2023-10-06
Revert "Default the event file to retis.data"

atenart:
Reverts retis-org/retis#266

This introduced a requirement for the `--out` parameter to use `=` when setting a value, which we find not practical. The other commits are OK but for now let's revert all those patches and spend some time to think about what could be done and do better tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 /target
 .out
 __pycache__/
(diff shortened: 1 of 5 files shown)


vlrpl (approved):
lgtm, approved!

PR #271, opened 2023-10-11
Default the event file to retis.data

atenart:
Second tentative of reverted PR #266.

@vlrpl suggested two solutions: removing `require_equals` and defaulting `--out` to `retis.data` (not in an `Option`). The former solution is chosen here, but the whole situation is discussed in length in the first commit message.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 .out
 __pycache__/
+retis.data


vlrpl:
@atenart, thanks for working on this.
The PR lgtm. Just one question about clap and potential breakage.
Any reason you chose to rely on default_missing_value instead of moving that logic into our code to avoid potential (yet very unlikely) future breakage?
Is the idea to leave room for any other potentially different approach?

amorenoz (approved):
In general the PR looks good to me.

When testing, the only thing that came up to me as a suggestion (maybe for a follow up patch is), now that the file name is not explicit, it might be worth adding an INFO level log after the capture has finished that says something like:
```
[INFO] Events written to retis.data
```
Given it's explained in the `help` and the option is opt-in, I don't consider this a blocking issue.

amorenoz:
> One thing that should be done however is to also convert the pcap command now that it was merged, so it also uses retis.data as the default input. Should I add this to this PR (and rebase) or do this in a following one?

Whatever you prefer.


> 
> > Given it's explained in the `help` and the option is opt-in, I don't consider this a blocking issue.
> 
> Also my reasoning was when looking at similar behavior in existing tools, for such core parameter, it's expected that the users will know the default log file name. To add to this, not only the collecting part is using this name, but the post-processing ones too. Which makes a classic use of Retis not to wonder about the file name being used.
> 

I agree.

PR #275, opened 2023-10-23
Support non-core drop reasons

atenart:


src/module/skb_drop/event.rs
```diff
@@ -47,9 +92,59 @@ impl SkbDropEventFactory {
...
     pub(crate) fn bpf() -> Result<Self> {
-        let reasons = Some(parse_enum("skb_drop_reason", &["SKB_", "DROP_REASON_"])?);
+        let subsys = parse_enum("skb_drop_reason_subsys", &["SKB_DROP_REASON_SUBSYS_"])?;
+
+        // Parse each sub-system drop reasons.
+        let mut reasons = HashMap::new();
+
+        if !subsys.is_empty() {
+            subsys.iter().try_for_each(|(id, name)| -> Result<()> {
+                if name != "NUM" {
```
vlrpl, line 102: is parsing the guard in the first place needed?

src/core/probe/kernel/inspect.rs
```diff
@@ -30,6 +30,22 @@ pub(super) fn inspect_symbol(symbol: &Symbol) -> Result<ProbeConfig> {
...
 
+/// Find the offset of drop reason enums. We do not distinguish them as they're
+/// part of the same (virtual) enum.
+fn drop_reason_offset(symbol: &Symbol) -> Result<Option<i8>> {
+    Ok(
+        if let Some(offset) = symbol.parameter_offset("enum skb_drop_reason")? {
+            Some(offset as i8)
+        } else if let Some(offset) = symbol.parameter_offset("enum mac80211_drop_reason")? {
+            Some(offset as i8)
+        } else {
```
vlrpl, line 41: why not an explicit check here as well?

vlrpl: sure. Had no doubt about the behavior, the point was the readability as you said.
Let's just make Clippy happy without adding exceptions.

vlrpl: ok :)
I was assuming you didn't want to add the exception. Had no strong feelings about it, but I also slightly prefer to allow this case. Thanks

vlrpl (approved):
thanks for the respin.
lgtm. Merged, thanks.

PR #277, opened 2023-11-07
tree: add skb offset validation helpers

atenart:
Validating offsets in an skb is no easy task. Add helpers to make checks more consistent in our probes and do the best we can do to guess if offsets are valid or not.

Note that this cannot be 100% failsafe. When converting Retis to parse packets in the Rust side we'll be able to improve further offset detection by using the BPF side as a hint only.


vlrpl (commented):
lgtm

vlrpl (approved):

PR #279, opened 2023-11-07
Add documentation for all missing modules

atenart:
Result can be seen at https://retis.readthedocs.io/en/at-skb-doc/


docs/README.md
```diff
@@ -97,6 +97,18 @@ $ retis sort retis.data
     if 2 (eth0) rxif 2 172.16.42.1.40532 > 172.16.42.2.8080 ttl 64 tos 0x0 id 14042 off 0 [DF] len 32 proto UDP (17) len 4
 ```
 
+An other post-processing command, `pcap`, can be used to generate `pcap-ng`
```
vlrpl, line 100: s/An other/another/

vlrpl:
Thanks @atenart. I find this extremely important and in good shape.
I know it is a bit out of scope, but I guess that time is ripe to add Authors.md, and this PR seems ideal to snuck it in.
WDYT?

vlrpl:
> > I know it is a bit out of scope, but I guess that time is ripe to add Authors.md, and this PR seems ideal to snuck it in. WDYT?
> 
> I'm personally not sure about such a file, but also not against it. Having said that there are multiple questions/possibilities:
> 
>     * Using a dedicated file. Using which order? Alphabetical, by first contribution, # of commits? Should we include emails? GH handle? Names?
> 
>     * Using a changelog file, with a list of all authors of a given release. Including # of commits? Plus same questions as above. Or just using the GH release page?
> 
> 
> I think it needs to be discussed first, so let's discuss it offline first :) If that's fine of course.

Totally fine. Better, I'd say. Let's sync offline.

vlrpl (approved):
lgtm, thanks!

PR #298, opened 2023-12-02
retis_in_container.sh: allow retis to fetch custom profiles

vlrpl:


vlrpl:
> > Not a fan of hidden automatic behaviors. What if one has local profiles but want to use shared profiles from the test team e.g. in a container? It should pollute its directory if the override is not allowed.
> 
> I wouldn't say this is hidden: it maps to the current behavior and `retis_in_container` aims at working as close as possible to a plain Retis binary. Also I think it 100% maps to our goal of using sane defaults to improve ux. Then and only then, if there is a need to manually set the value, we can always provide a way, which is the comment below,
> 

you're right, hidden doesn't fit well.

> > Can you elaborate the reason you're not sure and how overriding could be a problem/bad thing?
> 
> Not saying it would lead to bad things, more a comment on adding features only if there are users. If you think that is needed, I'm OK with that.
> 

yeah, probably not fitting wording as well here. With "bad things" I meant something I could have missed.
I edited my previous comment (you were faster in replying :) saying that both are fine (but I kind of think it is too much).

In my case, I needed a profile to create and throw away. Probably the '-p ./path/to/the/profile' would have been useful for me, but without that an override was more convenient (of course, the $HOME approach also would have worked).

Let's go with $HOME, it would be trivial to add the override if needed.

> > > Also, I think we should add support for providing profiles as a path, and this would make sense here too as we already map the current directory (for the log file initially). People could do `./retis_in_container.sh -p ./foo.yaml collect`.
> > 
> > 
> > I'm ok with this last if this is an addition (as it seems you intend). As a replacement it's not the same (only one file mapped) and probably too constraining.
> 
> Yes, as an addition.

PR #300, opened 2023-12-04
[RFC] profile: allow manually loading a profile

atenart:
This enables the Retis --profile cli parameter to load a profile from a file directly. This is useful when developing new profiles or when using a custom profile in a specific situation.

I'm not 100% sure if we want this, or want this in the current form. Some comments:

- As we support 1:n profiles in a single YAML definition, we can't use custom profiles with more than one defined profile. This makes the `--profile` cli command to be a little bit confusing, and makes using this option to develop profiles we'd like to integrate later a bit cumbersome.
- We already provide `.config/retis/profile` for use defined ones. The reasoning here is this could be handy for users to maintain their own set of profiles and share them in their own repositories. But still, for the above reason this might be preferred over this PR.
- This addition makes `--profile` values to have different meanings, files vs profile names, which is a bit of an issue.
- We could add an optional syntax in the `--profile` argument, eg `--profile ./foo.yaml:name` but that is a bit complex IMHO for an argument designed to symplify things.

Thoughts?

### DIFF
--- a/src/cli/cli.rs
+++ b/src/cli/cli.rs
@@ -203,9 +203,9 @@ pub(crate) struct MainConfig {
         long,
         short,
         value_delimiter = ',',
-        help = "Comma separated list of profile names to apply"
+        help = "Comma separated list of profile names or paths to apply"
     )]
-    pub(crate) profile: Vec<String>,
+    pub(crate) profile: Vec<PathBuf>,
     #[arg(
         long,
         help = "Path to kernel configuration (e.g. /boot/config-6.3.8-200.fc38.x86_64; default: auto-detect)"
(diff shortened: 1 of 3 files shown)


amorenoz:
I agree with your concerns. This makes the cli argument confusing and with double-meaning.
If we need  other  sources of profiles, could we have a `--extra-profile-path(s)` (or even an env RETIS_PROFILE_PATH) that extends the places retis looks for paths at? That would still allow fancy profile sharing and keep `-p` consistent.

vlrpl:
> I agree with your concerns. This makes the cli argument confusing and with double-meaning. If we need other sources of profiles, could we have a `--extra-profile-path(s)` (or even an env RETIS_PROFILE_PATH) that extends the places retis looks for paths at? That would still allow fancy profile sharing and keep `-p` consistent.

We all agree that we would avoid the double meaning.
What about simply having something like: `--local-conf-dir /path/to/...`?
This would replace `.config/retis` (maybe for specific cases or cases where no homedir is present at all)?

atenart:
Closing this for now.

PR #304, opened 2023-12-05
Raw packet capture fixes

atenart:
Two raw packet capture fixes for when:

- Capture length is the max allowed (was converted to 0...).
- No data lies in the skb linear buffer part.


vlrpl (approved):
lgtm, thanks!

PR #305, opened 2023-12-06
docs: fix navigation links to the index

atenart:
Mkdocs supports both index.md and README.md as the index. Both are converted to index.html. Other files, says foo.md and bar.md, are converted to foo/index.html and bar.index.html.

The above makes browsing URLs looking like:
/
/foo/
/bar/
...

We currently use README.md as our index. The issue is ReadTheDocs uses /README/ instead of / in some links, including the version ones. To fix this and avoid cases like this, rename the index to index.md, which should be better supported.

Note: using README.md allowed to browse the documentation in Github directly, but that was still limited as navigation wasn't there. With that it's safe to say the `docs` folder isn't meant to be browsed directly but serves as the documentation sources.

---

Another way to fix the above issue, which is already in place, is to add a redirection from /README/ to / in the ReadTheDocs interface. This is why this is an RFC as I'd like another take on this. IMHO, this PR looks safer in the long run, but you might thing otherwise.


vlrpl (dismissed):
lgtm, thanks!

vlrpl (approved):
looks good to me. Thanks!

PR #309, opened 2023-12-08
Auto-probe mode

### LINKED ISSUE #88: discovery mode: auto-attach to reported symbols in stack traces
- Keep a cache of functions we tried already.
- Exclude the current list of probes.
- Filter functions based on their parameters & if they are traceable.
- Might need to exclude things like BPF trampoline.

atenart:
This enables collection to use a new "auto-probe" mode, working as follow: stack traces are requested and functions found in those stack traces are evaluated and additional probes are added if they fit the requirements. Using this mode can help building a list of probes to use for following a given flow. More details about the implementation in the commit messages.

Part of this series is another rework of the probe manager. But fear not, there is no big functional changes and the code is very close to what it was. But the logic is better defined and that allows additional things, such as runtime registration of probes.

This is based on a few patches that are in other PRs (btf-rs update patches mainly). This is fine as they're all targeted at an earlier milestone.

One future possibility is the generation of a profile at the end of the collection, so it could be reused once where to probe is known.

`$ retis collect -a -f 'icmp and host 1.1.1.1' --cmd 'ping 1.1.1.1'`

Fixes #88.

### DIFF
--- a/src/collect/cli.rs
+++ b/src/collect/cli.rs
@@ -98,6 +98,24 @@ Examples of meta filters:
     #[arg(
         long,
         default_value = "false",
+        help = "When set, evaluates where Retis could add additional probes based on functions reported
+in the events stack traces (their display is still controlled by --stack). All matching
+functions are probed at runtime using kprobes.
+
+Notes:
+- Using a filter is required (--filter-packet and/or --filter-meta).
+- If no explicit probe is given, tp:skb:kfree_skb and tp:skb:consume_skb are used as a
+  starting point.
+- Additional probes are added only after events including them in their stack trace are
+  reported; this means the first packets hitting a probe won't be reported.
+- Packets will only be followed prior to the initial set of probes (as this mode uses
+  stack traces). This also means the filter must match packets as they appear in the
+  initial set of probes; packet transformation can't be automatically detected."
+    )]
+    pub(crate) probe_stack: bool,
+    #[arg(
+        long,
+        default_value = "false",
         help = r#"Allow the tool to setup all the system changes needed to make the tracing
 fully operational.
 
--- a/src/collect/collector.rs
+++ b/src/collect/collector.rs
@@ -30,8 +30,8 @@ use crate::{
             packets::filter::FilterPacket,
         },
         inspect::check::collection_prerequisites,
-        kernel::symbol::{matching_events_to_symbols, matching_functions_to_symbols},
-        probe::*,
+        kernel::symbol::{matching_events_to_symbols, matching_functions_to_symbols, Symbol},
+        probe::{kernel::probe_stack::ProbeStack, *},
         signals::Running,
         tracking::{gc::TrackingGC, skb_tracking::init_tracking},
     },
@@ -184,7 +184,14 @@ impl Collectors {
             .downcast_ref::<Collect>()
             .ok_or_else(|| anyhow!("wrong subcommand"))?;
 
-        if collect.args()?.stack {
+        if collect.args()?.probe_stack
+            && collect.args()?.packet_filter.is_none()
+            && collect.args()?.meta_filter.is_none()
+        {
+            bail!("Probe-stack mode requires filtering (--filter-packet and/or --filter-meta)");
+        }
+
+        if collect.args()?.stack || collect.args()?.probe_stack {
             self.probes
                 .builder_mut()?
                 .set_probe_opt(probe::ProbeOption::StackTrace)?;
@@ -249,6 +256,19 @@ impl Collectors {
             self.tracking_config_map = Some(map);
         }
         Self::setup_filters(self.probes.builder_mut()?, collect)?;
+
+        // If probe_stack is on and user hasn't provided a starting point, use
+        // skb:consume_skb & skb:kfree_skb.
+        if collect.args()?.probe_stack && collect.args()?.probes.is_empty() {
+            self.probes
+                .builder_mut()?
+                .register_probe(Probe::raw_tracepoint(Symbol::from_name(
+                    "skb:consume_skb",
+                )?)?)?;
+            self.probes
+                .builder_mut()?
+                .register_probe(Probe::raw_tracepoint(Symbol::from_name("skb:kfree_skb")?)?)?;
+        }
 
         // Setup user defined probes.
         collect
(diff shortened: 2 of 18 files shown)


amorenoz:
I haven't had the time to look deeply in to the code, but I have a question about user-friendliness.

Is it just me or does it feel strange to force users to write a packet filter?
I guess the reason is that functions that are hit by the "cmd" are very likely to be used by pretty much all traffic in the system and it's a way to avoid flooding the user with tons of events. But still, that seems to be pushing to the user some internal implementation detail.  

Maybe a possibility would be to support filtering by PID (which IMHO, has value in it's own). Combined with `--cmd`, it would allow the user to add many probes and no packet-filter without fearing event flooding or even run commands with still-unknown network behavior.

Also, not sure if the wording of the options conveys the fact that probes are added incrementally but _after_ they are hit at least once. If the user's "reproducer" is on the first packet (e.g: TCP SYN), they might loose some important events. In that case, the user would have to add probes manually so they are pre-loaded. I like the idea of automatically adding probes, I'm just concerned if the users will understand the implications.

atenart:
> Is it just me or does it feel strange to force users to write a packet filter? I guess the reason is that functions that are hit by the "cmd" are very likely to be used by pretty much all traffic in the system and it's a way to avoid flooding the user with tons of events. But still, that seems to be pushing to the user some internal implementation detail.

Right, the reason to require a filter is if no filter is used there are high chances the result would be similar to `retis collect --probe *` which we do not support (too many probes).

> Maybe a possibility would be to support filtering by PID (which IMHO, has value in it's own). Combined with `--cmd`, it would allow the user to add many probes and no packet-filter without fearing event flooding or even run commands with still-unknown network behavior.

That is a good idea, and it falls into the "a filter is required" IMHO. The intention is to allow all available filter methods for auto-probe, the packet one being the only one available when this was firstly written. Before merging this PR we should allow meta-filtering too. For PID filtering, it's a good follow-up IMHO but is just doesn't exist yet. Or maybe you had in mind using meta filtering for this?

> Also, not sure if the wording of the options conveys the fact that probes are added incrementally but _after_ they are hit at least once. If the user's "reproducer" is on the first packet (e.g: TCP SYN), they might loose some important events. In that case, the user would have to add probes manually so they are pre-loaded. I like the idea of automatically adding probes, I'm just concerned if the users will understand the implications.

You're right, we should make this more explicit. What about an additional sentence in the help, like "Note that additional probes are added only after events including them in their stack trace are reported first; this means the first packets hitting a probe won't be reported."

In addition, I'd like one day to be able to generate a profile based on the auto-probe state at the end of a capture; so we can generate a profile based on a scenario and then use it to see all the packets in a subsequent collection.

src/collect/cli.rs
```diff
@@ -95,6 +95,18 @@ Examples of meta filters:
 --filter-meta 'sk_buff.dev.nd_net.net.ns.inum == 4026531840'"#
     )]
     pub(super) meta_filter: Option<String>,
+    #[arg(
+        short,
+        long,
+        default_value = "false",
+        help = "Auto-probe mode evaluates where Retis could add additional probes based on functions reported
```
vlrpl, line 102: Needless to say this is a matter of personal taste (like always happens with names).
I don't find `auto-probe` very descriptive (in terms of this particular feature).
Have you considered something more explicit like `--follow-stack-trace`?
I can understand that the `short` option may "suffer" about this, but I guess it's fine (especially now that we have completion).

atenart: Or maybe `--probe-stack`? Having a long option only is fine I guess for this one.

vlrpl: sounds better to me

src/collect/cli.rs
```diff
@@ -95,6 +95,18 @@ Examples of meta filters:
...
+in the events stack traces ((their display is still controlled by --stack). All matching functions are probed
+at runtime using kprobes.
+
+Using a filter is required. If no explicit probe is given, skb:kfree_skb and skb:consume_skb
```
vlrpl, line 106: We cannot "track-back", meaning, the filter will match the packet as it is in the [automatic] probes set (e.g. `kfree_skb`, `consume_skb` or manually, e.g. in the tx path using `dev_hard_start_xmit` and so forth).

The filter then is also set globally, which means that the packet matched in `dev_hard_start_xmit` may not be matched before nat happens, assuming nat (same thing for the rx path).
If confirmed, should we "call out" this behavior?

atenart: You're right; what about:

```rust
    #[arg(
        long,
        default_value = "false",
        help = "When set, evaluates where Retis could add additional probes based on functions reported
in the events stack traces (their display is still controlled by --stack). All matching
functions are probed at runtime using kprobes.

Notes:
- Using a filter is required.
- If no explicit probe is given, tp:skb:kfree_skb and tp:skb:consume_skb are used as a
  starting point.
- Additional probes are added only after events including them in their stack trace are
  reported; this means the first packets hitting a probe won't be reported.
- Packets will only be followed prior to the initial set of probes (as this mode uses
  stack traces). This also means the filter must match packets as they appear in the
  initial set of probes; packet transformation can't be automatically detected."
```

vlrpl: yes, sgtm

atenart:
- Rebased on top of main (to resolve conflict).
- Renamed "auto probe" into "probe stack".
- Refined the help message.
- Initialized the probe stack tracing of probed functions with the current status.

src/collect/collector.rs
```diff
@@ -184,7 +184,14 @@ impl Collectors {
             .downcast_ref::<Collect>()
             .ok_or_else(|| anyhow!("wrong subcommand"))?;
 
-        if collect.args()?.stack {
+        if collect.args()?.probe_stack
+            && collect.args()?.packet_filter.is_none()
+            && collect.args()?.meta_filter.is_none()
+        {
+            bail!("Auto-probe mode requires filtering (`--filter-packet` or `--filter-meta`)");
```
vlrpl, line 191: at least one of, right?

atenart: Yes, at least one. Am I missing something? :)

vlrpl: no :) 
I meant as it is phrased (the part inside the parenthesis) it may seem one or the other while both are usable simultaneously.

atenart: IMO as this is the error message, that's fine. One filter is missing. However those options are not described in the `--probe-stack` help. (And it seems the error message above still refers to "auto-probe"). I'll fix that.

vlrpl: thanks!

vlrpl (approved):

PR #311, opened 2023-12-13
module: skb: fix (again) skb hook for old kernels

atenart:
Same as 1bb047ba166b ("module: skb: bpf: fix verifier issue on older kernels") but that had a flaw and was broken by a recent change that fixed it.

Could not find a better way. If not using &= no size check works. This has the downside of limiting the capture size of L3-only packets to 0x7f; but that is not that bad. Many packets headers should be within that limit. Another working solution would be to increase our buffer size to `PACKET_CAPTURE_SIZE + sizeof(ethhdr)`, which allows to capture up to PACKET_CAPTURE_SIZE of data in each case.

Fixes: 3afddf5ef4fa ("module: skb: fix raw packet retrieval for max capture size")

### DIFF
--- a/src/module/skb/bpf/skb_hook.bpf.c
+++ b/src/module/skb/bpf/skb_hook.bpf.c
@@ -387,13 +387,6 @@ static __always_inline int process_packet(struct retis_raw_event *event,
 	unsigned char *head, *data;
 	struct skb_packet_event *e;
 	u16 mac, network;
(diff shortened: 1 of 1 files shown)


vlrpl (commented):
the patch lgtm.
We had some discussion about it offline.
I have only one request. Can we add a chunk of the verifier's log in the commit description?

vlrpl (approved):
lgtm, thanks!

PR #321, opened 2023-12-20
module: skb: another fix for old verifiers

atenart:
Very similar to 98e9882ab666 and 3afddf5ef4fa, but on RHEL 8.9 this time. Fix this with a small trick for now, but we should investigate the root cause and come up with a real solution (even if the issue is the verifier).


vlrpl (approved):
As discussed offline, lgtm. Let's make it spin through the CI and merge.

PR #325, opened 2024-01-02
Remove all #[cfg(not(test))]

### LINKED ISSUE #103: Remove all #[cfg(not(test))]

liuhangbin:
Close #103

Before:
test result: ok. 56 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out; finished in 32.05s
After:
test result: ok. 56 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out; finished in 30.53s

### DIFF
--- a/src/collect/collector.rs
+++ b/src/collect/collector.rs
@@ -1,4 +1,3 @@
-#[cfg(not(test))]
 use std::os::fd::{AsFd, AsRawFd};
 use std::{
     collections::HashSet,
@@ -19,7 +18,6 @@ use crate::{
     process::display::PrintSingle,
 };
 
-#[cfg(not(test))]
 use crate::core::probe::kernel::{config::init_stack_map, kernel::KernelEventFactory};
 use crate::{
     cli::{dynamic::DynamicCommand, CliConfig, FullCli},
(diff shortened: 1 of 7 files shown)


liuhangbin:
> Looks like build (eg. `cargo check` is not passing). Making it explicit since it's your first PR; we usually wait for all CI tests to pass before making a review.

OK, I will check it.

> Still had a quick look at this and I don't think this can work (eg. BPF maps handling are not skipped anymore while testing). Tests might have worked when ran as root, but they should pass as a normal user.

Thanks for this info. I will run the test with normal user in future.

liuhangbin:
> > Still had a quick look at this and I don't think this can work (eg. BPF maps handling are not skipped anymore while testing). Tests might have worked when ran as root, but they should pass as a normal user.

OK, looks all the failures are because the bpf map init failed,  which calls `libbpf_rs::MapHandle::create()` and needs root
permission.  I don't have any idea how to avoid it except not build it via `#[cfg(not(test))]`, or define it as Option<libbpf_rs::MapHandle>. 

I don't know if convert `libbpf_rs::MapHandle` to `Option<libbpf_rs::MapHandle>` accepts. Because this would change the code a lot.

liuhangbin:
> So maybe the best way is close to the current implementation, which might be improved by better grouping those

Since this is hard to remove all `#[cfg(not(test))]` and we still need to use(or re-implement) the current implementation. I plan to close this PR first. I will re-work this when we have a better solution.

PR #326, opened 2024-01-04
Improve symbols validation

### LINKED ISSUE #322: Auto-collector mode fails to detect nft requirement mismatch
Seen on 8.4 but could be seen on other distros.

```
Error: Could not initialize the Nft collector: No type linked to name __nft_trace_packet
```

This looks like `Symbol::from_name()` call succeed in `can_run` but later use of the symbol in `init` fails as BTF information cannot be found. This could affect other modules too.

One solution would be to improve the check in `can_run` to use BTF inspection directly; but this might hide another issue if a probe is found to be valid in sysfs but no BTF info can be found. It might also be related to the btf-rs issue with strings.

atenart:
Fixes #322.

### DIFF
--- a/src/core/kernel/symbol.rs
+++ b/src/core/kernel/symbol.rs
@@ -18,19 +18,26 @@ impl Symbol {
     /// using different techniques depending on what is available.
     pub(crate) fn from_name(name: &str) -> Result<Symbol> {
         let mut debugfs = false;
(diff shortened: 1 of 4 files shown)


vlrpl (approved):
thanks for the improvement. LGTM, approved!

Issue #68, opened 2023-01-27
add bash completion script


vlrpl:
There's a chance we'll end up having a decent number of options and adding a bash completion file could be a nice to have

liuhangbin:
Hi @vlrpl @atenart 

I checked this feature. clap_complete has 3 ways to generate the shell completion.

The 1st way is using [generate_to](https://docs.rs/clap_complete/4.4.6/clap_complete/generator/fn.generate_to.html), which will generate the shell completion file at compile time. With this way we need to put the generator in `build.rs`.  But this way is very hard to implement as we need to `include!(src/cli/cli.rs)`, which also depends on a lot of modules.

The 2nd way is using [generate](https://docs.rs/clap_complete/4.4.6/clap_complete/generator/fn.generate.html), which could generate the completions file for a specified shell at runtime. But we need to add a `generate` parameter. e.g.
```Bash
# retis -h

Usage: retis [OPTIONS] <COMMAND>

Commands:
  collect  Collect network events
  print    Print stored events to stdout
  sort     Sort stored events in series based on tracking id
  pcap     Generate a PCAP file from stored events
  profile  Manage Profiles

Options:
      --log-level <LOG_LEVEL>  Log level [default: info] [possible values: error, warn, info, debug, trace]
  -p, --profile <PROFILE>      Comma separated list of profile names to apply
      --kconf <KCONF>          Path to kernel configuration (e.g. /boot/config-6.3.8-200.fc38.x86_64; default: auto-detect)
  -g, --generate <generator>   [possible values: bash, elvish, fish, powershell, zsh]
  -h, --help                   Print help (see more with '--help')
  -V, --version                Print version

# retis -g bash > bash_complete
# source bash_complete
```

The 3rd way is using the [unstalbe dynamic](https://docs.rs/clap_complete/4.4.6/clap_complete/dynamic/index.html) feature. With this way we can hide the generate parameter and user will not see it normally. e.g.
```
# retis -h
A subcommand definition to `flatten` into your CLI

Usage: retis [OPTIONS] <COMMAND>

Commands:
  collect  Collect network events
  print    Print stored events to stdout
  sort     Sort stored events in series based on tracking id
  pcap     Generate a PCAP file from stored events
  profile  Manage Profiles

Options:
      --log-level <LOG_LEVEL>  Log level [default: info] [possible values: error, warn, info, debug, trace]
  -p, --profile <PROFILE>      Comma separated list of profile names to apply
[...]

PR #329, opened 2024-01-08
cli: add shell completion subcommand

liuhangbin:
As the clap_complete dynamic module is unstable and hard-codes the
subcommand as `complete`, which conflicts with the current `collect`
subcommand. Let's still use the legacy static way to generate the
completion file.

Usage:
$ retis generate -h
Generate completions file for a specified shell

Usage: retis generate [OPTIONS]

Options:
      --shell <SHELL>        Specify shell to complete for [possible values: bash, elvish, fish, powershell, zsh]
      --register <REGISTER>  Path to write completion-registration to
  -h, --help                 Print help

$ retis generate --shell bash --register .
$ source retis.bash

Close: #68
---
v5: Remove `fn thin()` from `impl SubCommand`. Update doc description.
v4: Update docs/install.md about how to auto-complete. Fix `collect` subcommand auto-completion.
v3: Rename subcommand to `sh-complete`. Move the generate.rs to new folder generate/completion.rs. Rename `struct Generate` to `struct Complete` as we rename generate.rs to completion.rs. 
v2: Use static generator as dynamic generator use hard-code parameter `complete`, which conflicts with `collect`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ btf-rs = "1.0"
 byteorder = "1.5"
 caps = "0.5"
 clap = { version = "4.0", features = ["derive", "string"] }
+clap_complete = "4.4"
 elf = "0.7"
 flate2 = "1.0"
 libbpf-rs = "0.22"
--- a/src/cli/cli.rs
+++ b/src/cli/cli.rs
@@ -93,9 +93,6 @@ pub(crate) trait SubCommand {
     /// subcommand-specific functionality.
     fn as_any_mut(&mut self) -> &mut dyn Any;
 
-    /// Generate the clap Command to be used for "thin" parsing.
-    fn thin(&self) -> Result<Command>;
-
     /// Generate the clap Command to be used for "full" parsing.
     ///
     /// This method should be called after all dynamic options have been registered.
@@ -158,10 +155,6 @@ where
 
     fn as_any_mut(&mut self) -> &mut dyn Any {
         self
-    }
-
-    fn thin(&self) -> Result<Command> {
-        Ok(<Self as clap::CommandFactory>::command())
     }
 
     fn full(&self) -> Result<Command> {
@@ -270,9 +263,9 @@ impl ThinCli {
             .disable_help_subcommand(true)
             .infer_subcommands(true)
             .subcommand_required(true);
-        // Add thin subcommands so that the main help shows them.
+        // Add full subcommands so that the main help shows them.
         for sub in self.subcommands.iter() {
-            command = command.subcommand(sub.thin().expect("thin command failed"));
+            command = command.subcommand(sub.full().expect("full command failed"));
         }
 
         // Determine the subcommand that was run while ignoring errors from yet-to-be-defined
@@ -458,9 +451,6 @@ mod tests {
         }
         fn as_any_mut(&mut self) -> &mut dyn Any {
             self
-        }
-        fn thin(&self) -> Result<Command> {
-            Ok(Command::new("sub1").about("does some things"))
         }
         fn full(&self) -> Result<Command> {
             Ok(Sub1::augment_args(
--- a/src/cli/cli.rs
+++ b/src/cli/cli.rs
@@ -19,6 +19,7 @@ use super::dynamic::DynamicCommand;
 use crate::benchmark::cli::Benchmark;
 use crate::{
     collect::cli::Collect,
+    generate::Complete,
     module::{ModuleId, Modules},
     process::cli::*,
     profiles::{cli::ProfileCmd, Profile},
@@ -386,6 +387,10 @@ impl FullCli {
     pub(crate) fn get_subcommand_mut(&mut self) -> Result<&mut dyn SubCommand> {
         Ok(self.subcommand.as_mut())
     }
+
+    pub(crate) fn get_command(&self) -> Command {
+        self.command.clone()
+    }
 }
 
 /// CliConfig represents the result of the Full CLI parsing
@@ -421,6 +426,7 @@ pub(crate) fn get_cli() -> Result<ThinCli> {
     cli.add_subcommand(Box::new(Sort::new()?))?;
     cli.add_subcommand(Box::new(Pcap::new()?))?;
     cli.add_subcommand(Box::new(ProfileCmd::new()?))?;
+    cli.add_subcommand(Box::new(Complete::new()?))?;
 
     #[cfg(feature = "benchmark")]
     cli.add_subcommand(Box::new(Benchmark::new()?))?;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -295,6 +295,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.4.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "97aeaa95557bd02f23fbb662f981670c3d20c5a26e69f7354b28f57092437fcd"
+dependencies = [
+ "clap",
+]
+
+[[package]]
 name = "clap_derive"
 version = "4.4.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1144,6 +1153,7 @@ dependencies = [
  "byteorder",
  "caps",
  "clap",
+ "clap_complete",
  "elf",
  "flate2",
  "libbpf-cargo",
--- a/docs/install.md
+++ b/docs/install.md
@@ -106,3 +106,10 @@ only and Retis won't be able to fully filter probes.
 $ sudo setcap cap_sys_admin,cap_bpf,cap_syslog=ep $(which retis)
 $ retis collect
 ```
+
+### Shell auto-completion
+
+Retis can generate completion files for shells (Bash, Zsh, Fish...).
+For example to enable auto-completion of Retis command in Bash, you can
+add line `source <(retis sh-complete --shell bash)` in .bashrc, then
+the command parameter could be auto-completed when pressing <Tab>.
--- a/src/collect/cli.rs
+++ b/src/collect/cli.rs
@@ -148,10 +148,6 @@ impl SubCommand for Collect {
         })
     }
 
-    fn thin(&self) -> Result<Command> {
-        Ok(Command::new("collect").about("Collect network events"))
-    }
-
     fn name(&self) -> String {
         "collect".to_string()
     }
@@ -180,7 +176,7 @@ impl SubCommand for Collect {
             .collectors
             .command()
             .to_owned()
-            .about("Collect events")
+            .about("Collect network events")
             .long_about(long_about)
             .mut_arg("collectors", |a| {
                 a.value_parser(PossibleValuesParser::new(possible_collectors.clone()))
--- a/src/generate/completion.rs
+++ b/src/generate/completion.rs
@@ -1,0 +1,98 @@
+//! # Completion
+//!
+//! Generate a completions file for a specified shell at runtime.
+
+use std::{any::Any, io::Write, path::PathBuf};
+
+use anyhow::Result;
+use clap::{
+    error::Error as ClapError,
+    {value_parser, ArgMatches, Command, Parser},
+};
+use clap_complete::{generate, Generator, Shell};
+
+use crate::{cli::*, module::Modules};
+
+/// Generate completion file for a specified shell
+#[derive(Parser, Debug, Default)]
+#[command(name = "sh-complete")]
+pub(crate) struct Complete {
+    /// Specify shell to complete for
+    // We use an Option and require the parameter to be set here to allow
+    // deriving Default on Complete.
+    #[arg(long, required = true, value_parser(value_parser!(Shell)))]
+    shell: Option<Shell>,
+
+    /// Path to write completion-registration to
+    #[arg(long)]
+    register: Option<PathBuf>,
+}
+
+impl SubCommand for Complete {
+    fn new() -> Result<Self>
+    where
+        Self: Sized,
+    {
+        Ok(Self::default())
+    }
+
+    fn name(&self) -> String {
+        <Self as clap::CommandFactory>::command()
+            .get_name()
+            .to_string()
+    }
+
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn as_any_mut(&mut self) -> &mut dyn Any {
+        self
+    }
+
+    fn full(&self) -> Result<Command> {
+        Ok(<Self as clap::CommandFactory>::command())
+    }
+
+    fn update_from_arg_matches(&mut self, args: &ArgMatches) -> Result<(), ClapError> {
+        <Self as clap::FromArgMatches>::update_from_arg_matches(self, args)
+    }
+
+    fn runner(&self) -> Result<Box<dyn SubCommandRunner>> {
+        Ok(Box::new(CompleteRunner {}))
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct CompleteRunner {}
+
+impl SubCommandRunner for CompleteRunner {
+    fn check_prerequisites(&self) -> Result<()> {
+        Ok(())
+    }
+
+    fn run(&mut self, cli: FullCli, _modules: Modules) -> Result<()> {
+        let mut cmd = cli.get_command();
+        let matches = cli.get_command().get_matches();
+
+        if let Some(sub_m) = matches.subcommand_matches("sh-complete") {
+            if let Some(generator) = sub_m.get_one::<Shell>("shell") {
+                let mut buf = Vec::new();
+                let name = cmd.get_name().to_string();
+
+                generate(*generator, &mut cmd, name.clone(), &mut buf);
+                if let Some(out_path) = sub_m.get_one::<PathBuf>("register") {
+                    if out_path.is_dir() {
+                        let out_path = out_path.join(generator.file_name(&name));
+                        let _ = std::fs::write(out_path, buf);
+                    } else {
+                        let _ = std::fs::write(out_path, buf);
+                    }
+                } else {
+                    let _ = std::io::stdout().write_all(&buf);
+                }
+            }
+        }
+        Ok(())
+    }
+}
--- a/src/generate/mod.rs
+++ b/src/generate/mod.rs
@@ -1,0 +1,6 @@
+//! # Generate
+//!
+//! Generate a completions file for a specified shell at runtime.
+
+pub(crate) mod completion;
+pub(crate) use self::completion::*;
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,6 +4,7 @@ use log::{info, trace, warn, LevelFilter};
 mod cli;
 mod collect;
 mod core;
+mod generate;
 mod module;
 mod process;
 mod profiles;


liuhangbin:
Hmm, I saw error

```
error: accessing first element with `self
                           .untracked.get(0)`
  --> src/process/series.rs:84:19
   |
84 |                   < self
   |  ___________________^
85 | |                     .untracked
86 | |                     .get(0)
   | |___________________________^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#get_first
   = note: `-D clippy::get-first` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::get_first)]`
help: try
   |
84 ~                 < self
85 +                     .untracked.front()
   |

error: could not compile `retis` (bin "retis") due to previous error

```
when doing self-checking. But I didn't expect it to be caused by my patch. I will check it.

Cargo.toml
```diff
@@ -28,6 +28,7 @@ btf-rs = "1.0"
 byteorder = "1.5"
 caps = "0.5"
 clap = { version = "4.0", features = ["derive", "string"] }
+clap_complete = { version = "4.4.6", features = ["unstable-dynamic"] }
```
atenart, line 31: nit: Please use `x.y` numbers only in version (`4.4`).

src/cli/cli.rs
```diff
@@ -10,9 +10,11 @@ use clap::{
     builder::PossibleValuesParser,
     error::Error as ClapError,
     error::ErrorKind,
-    {ArgMatches, Args, Command, FromArgMatches},
+    {ArgMatches, Args, Command, FromArgMatches, Subcommand},
 };
+use clap_complete::dynamic::shells::CompleteCommand;
 use log::debug;
+use std::process;
```
atenart, line 17: nit: Please use the same include order as the rest of the project:

```
std includes
<newline>
ext crates includes
<newline>
this crate includes
```

atenart (changes requested):
While testing found a few issues or things that might be improved:

- Completion for the `retis collect` sub-command does not work.
- `complete` sub-command is hidden, which is good, but it is used while inferring other sub-commands. This makes `retis c` for collection to not work anymore. As `complete` is a "special" command it would be good to exclude it from that, or maybe the only way is to define a `_complete` sub-command instead.
- PR miss changes in `Cargo.lock`.

Once the above is done, please add a section in the documentation about how to enable auto-completion (I think it's as simple as adding `source <(retis complete --shell bash --register -)` in `.bashrc`.

Thanks!

atenart:
> Hmm, I saw error
> 
> ```
> error: accessing first element with `self
>                            .untracked.get(0)`
>   --> src/process/series.rs:84:19
>    |
> 84 |                   < self
>    |  ___________________^
> 85 | |                     .untracked
> 86 | |                     .get(0)
>    | |___________________________^
>    |
>    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#get_first
>    = note: `-D clippy::get-first` implied by `-D warnings`
>    = help: to override `-D warnings` add `#[allow(clippy::get_first)]`
> help: try
>    |
> 84 ~                 < self
> 85 +                     .untracked.front()
>    |
> 
> error: could not compile `retis` (bin "retis") due to previous error
> ```
> 
> when doing self-checking. But I didn't expect it to be caused by my patch. I will check it.

Welcome to the Rust world :D This is triggered by an update in the linter (`cargo clippy`). I have a similar fix in all of my PRs too, first one that will be merged will fix it.

Cargo.toml
liuhangbin: Got it

src/cli/cli.rs
liuhangbin: I'm not sure if this is a convention for `rust`. If not, maybe we can add this to `Coding style` part of CONTRIBUTING.md

liuhangbin:
> * Completion for the `retis collect` sub-command does not work.

I will check this. Looks like I need to register the dynamic cmd after `ran_subcommand`.

> * `complete` sub-command is hidden, which is good, but it is used while inferring other sub-commands. This makes `retis c` for collection to not work anymore. As `complete` is a "special" command it would be good to exclude it from that, or maybe the only way is to define a `_complete` sub-command instead.

Currently the `complete` is hard-coded in `clap_complete`. Do you mean to use the `static complete` method by adding a new parameter `_complete`?

Thanks
Hangbin

src/cli/cli.rs
atenart: Rust defines includes can be grouped but I don't think it defines the exact above order; the above is not uncommon though. Feel free to add a word about this in the contributing guide.

atenart:
> > * `complete` sub-command is hidden, which is good, but it is used while inferring other sub-commands. This makes `retis c` for collection to not work anymore. As `complete` is a "special" command it would be good to exclude it from that, or maybe the only way is to define a `_complete` sub-command instead.
> 
> Currently the `complete` is hard-coded in `clap_complete`. Do you mean to use the `static complete` method by adding a new parameter `_complete`?

No, I think the dynamic completion is quite nice and easier to maintain. Isn't there a way to rename the dynamic completion command or to define it explicitly rather than using the default one?

liuhangbin:
> No, I think the dynamic completion is quite nice and easier to maintain. Isn't there a way to rename the dynamic completion command or to define it explicitly rather than using the default one?

From the [code](https://github.com/clap-rs/clap/blob/master/clap_complete/src/dynamic/shells/mod.rs#L31) it looks like the parameter is hardcode. I have opened [an issue](https://github.com/clap-rs/clap/issues/5293) for clap, let's see what the maintainer replies.

liuhangbin:
> > Currently the `complete` is hard-coded in `clap_complete`. Do you mean to use the `static complete` method by adding a new parameter `_complete`?
> 
> No, I think the dynamic completion is quite nice and easier to maintain. Isn't there a way to rename the dynamic completion command or to define it explicitly rather than using the default one?

Emm, I just found that we can `hide` a parameter with `.hide(ture)` when adding a new parameter. So how about adding a hiding static parameter `generate` first? and change to dynamic when the dynamic completion supports self-defined parameter. With this, we will not affect the current usage result. e.g.

```bash
# retis -h
Trace packets on the Linux kernel

Usage: retis [OPTIONS] <COMMAND>

Commands:
  collect  Collect network events
  print    Print stored events to stdout
  sort     Sort stored events in series based on tracking id
  pcap     Generate a PCAP file from stored events
  profile  Manage Profiles

Options:
      --log-level <LOG_LEVEL>  Log level [default: info] [possible values: error, warn, info, debug, trace]
  -p, --profile <PROFILE>      Comma separated list of profile names to apply
      --kconf <KCONF>          Path to kernel configuration (e.g. /boot/config-6.3.8-200.fc38.x86_64; default: auto-detect)
  -h, --help                   Print help (see more with '--help')
  -V, --version                Print version

# source <(retis -g bash)
# retis -g <tab>
bash        elvish      fish        powershell  zsh
# retis --generate <tab>
bash        elvish      fish        powershell  zsh
# retis p<tab>
print    pcap     profile
```

atenart:
> > > Currently the `complete` is hard-coded in `clap_complete`. Do you mean to use the `static complete` method by adding a new parameter `_complete`?
> > 
> > No, I think the dynamic completion is quite nice and easier to maintain. Isn't there a way to rename the dynamic completion command or to define it explicitly rather than using the default one?
> 
> Emm, I just found that we can `hide` a parameter with `.hide(ture)` when adding a new parameter. 

I don't think it has to be hidden, but to ease the use of Retis the command name should not conflict with the other commands (at least `collect` and `sort`) to allow using `retis c` for example.

> So how about adding a hiding static parameter `generate` first? and change to dynamic when the dynamic completion supports self-defined parameter. With this, we will not affect the current usage result. e.g.

Let me summarize my understanding; we have two choices:

1. [clap_complete](https://docs.rs/clap_complete/latest/clap_complete/index.html)::[generator](https://docs.rs/clap_complete/latest/clap_complete/generator/index.html)::[generate](https://docs.rs/clap_complete/latest/clap_complete/generator/fn.generate.html#). This generates a standalone completion script at runtime.
1. [clap_complete](https://docs.rs/clap_complete/latest/clap_complete/index.html)::[dynamic](https://docs.rs/clap_complete/latest/clap_complete/dynamic/index.html)::[shells](https://docs.rs/clap_complete/latest/clap_complete/dynamic/shells/index.html)::[CompleteCommand](https://docs.rs/clap_complete/latest/clap_complete/dynamic/shells/enum.CompleteCommand.html#). This can be used to query Retis at runtime about completion and is not a stable API.

The big difference is the output of 1 needs to be generated for each upgrade in the user's machine to match current commands and options, while 2 queries the current Retis binary and is by definition always up to date.

Using `source <(retis complete)` in `.bashrc` can make 1 work a bit like 2 most of the time, except after an upgrade as the sourced script would not match the new version of the binary. This is not perfect but might be better than the alternatives.

From the maintainer of `clap`,

```
dynamic completions are also very early in their development so I also generally recommend people use static completions at this time but I encourage fixes and improvements on dynamic completions.
```

This makes me wonder if dynamic completion would still be a possibility. Looks like it's not advertised (I can understand why) but apart from potential bugs, are there missing functionalities (for Bash for example)?

> If choose to use this way, should I add the parameter as `COMMAND` or `OPTIONS`?

A dedicated command please. For the name, maybe `generate-completion` would be best, to 1) not conflict with other commands for inference and for completion and 2) not conflict with a broader `complete` or `_complete` command once dynamic support is there.

liuhangbin:
> > Emm, I just found that we can `hide` a parameter with `.hide(ture)` when adding a new parameter.
> 
> I don't think it absolutely has to be hidden, but to ease the use of Retis the command name should at least not conflict with the other commands (at least `collect` and `sort`) to allow using `retis c` for example.

OK

> 
> > So how about adding a hiding static parameter `generate` first? and change to dynamic when the dynamic completion supports self-defined parameter. With this, we will not affect the current usage result. e.g.
> 
> Let me summarize my understanding; we have two choices:
> 
> 1. [clap_complete](https://docs.rs/clap_complete/latest/clap_complete/index.html)::[generator](https://docs.rs/clap_complete/latest/clap_complete/generator/index.html)::[generate](https://docs.rs/clap_complete/latest/clap_complete/generator/fn.generate.html#). This generates a standalone completion script at runtime.
> 2. [clap_complete](https://docs.rs/clap_complete/latest/clap_complete/index.html)::[dynamic](https://docs.rs/clap_complete/latest/clap_complete/dynamic/index.html)::[shells](https://docs.rs/clap_complete/latest/clap_complete/dynamic/shells/index.html)::[CompleteCommand](https://docs.rs/clap_complete/latest/clap_complete/dynamic/shells/enum.CompleteCommand.html#). This can be used to query Retis at runtime about completion and is not a stable API.
> 
> The big difference is the output of 1 needs to be generated for each upgrade in the user's machine to match current commands and options, while 2 queries the current Retis binary and is by definition always up to date.
> 
> Using `source <(retis complete)` in `.bashrc` can make 1 work a bit like 2 most of the time, except after an upgrade as the sourced script would not match the new version of the binary. This is not perfect but might be better than the alternatives.

correctly

> 
> From the maintainer of `clap`,
> 
> ```
> dynamic completions are also very early in their development so I also generally recommend people use static completions at this time but I encourage fixes and improvements on dynamic completions.
> ```
> 
> This makes me wonder if dynamic completion would still be a possibility. Looks like it's not advertised (I can understand why) but apart from potential bugs, are there missing functionalities (for Bash for example)?

I'm not sure either. There is no document about it...

> 
> > If choose to use this way, should I add the parameter as `COMMAND` or `OPTIONS`?
> 
> A dedicated command please. For the name, maybe `generate-completion` would be best, to 1) not conflict with other commands for inference and for completion and 2) not conflict with a broader `complete` or `_complete` command once dynamic support is there.

OK, let me try it first.

liuhangbin:
> OK, let me try it first.

Tried to use subcommand but not work. I even copied the [clap_complete example](https://github.com/clap-rs/clap/blob/master/clap_complete/examples/completion.rs) and change the 
```rust
    Command::new("completion")
        .arg(
            Arg::new("generator")
                .long("generate")
                .value_parser(value_parser!(Shell)),
        )
        .subcommand(value_hint_command)
```
to 
``` rust
   let generate = Command::new("generate")
        .arg(
            Arg::new("shell")
                .long("shell")
                .value_parser(value_parser!(Shell)),
        );
    Command::new("completion")
        .subcommand(value_hint_command)
        .subcommand(generate)
```
It also does not work and reports error `Mismatch between definition and access of `shell`. Unknown argument or group id.`. I need to check how `matches.get_one::<Shell>("shell")` get the parameter.

liuhangbin:
> It also does not work and reports error `Mismatch between definition and access of 'shell'. Unknown argument or group id.`. I need to check how `matches.get_one::<Shell>("shell")` get the parameter.

OK, I read the doc and found that I need to use `subcommand_matches` when checking subcommand...

liuhangbin:
Hi @atenart , recently I use `generate` as subcommand becase it's shorter than `generate-completion`. The parameters are the same with dynamic `complete`, with `--shell` and `--register`. With this, there is no need to change if we want to move to dynamic complete in the future.

Please tell me if you still want to use `generate-completion` as subcommand.

src/process/cli/generate.rs
```diff
@@ -0,0 +1,100 @@
...
+};
+use clap_complete::{generate, Generator, Shell};
+
+use crate::{cli::*, module::Modules};
+
+/// Generate completion file for a specified shell
+#[derive(Parser, Debug, Default)]
+#[command(name = "generate")]
+pub(crate) struct Generate {
+    /// Specify shell to complete for
```
atenart, line 20: Can you add a comment (`//`) about the `required = true` + `Option` here (AFAICT so `Default` can be derived).

atenart (changes requested):
A few nits, but most importantly the completion for the `complete` command still does not work for me. Only `-h` and `--help` are available when hitting `<tab>` while we should see other parameters (eg. `--stack` or `--filter`):

```
# retis collect -<tab>
-h      --help
```

> Please tell me if you still want to use generate-completion as subcommand.

I find `generate` to be too generic for a command that "only" generates a completion script. The issue with `complete` would be a conflict with `collect` for the `c` use though. If we want to use a sort command name we could go with `_complete`, but I'm not super sure.

An alternative could be to make something like `retis generate completion ...`.

@vlrpl WDYT?

> There is no '-c' or '-o' auto-complete for collect subcommand. But I feel this is the collect parameter's problem.

Agreed. Same with `--profile`. This might be solved with dynamic completion once they also add support for custom implementation of authorized values.

src/process/cli/generate.rs
liuhangbin: Could you give me an example of the comment? I saw other parts do like `#[arg(long, default_value = "something")]` for default value.

atenart: Something simple really, like,

```rust
    /// Specify shell to complete for
+   // We use an Option and require the parameter to be set here to allow deriving Default on Generate.
    #[arg(long, required = true, value_parser(value_parser!(Shell)))]
    shell: Option<Shell>,
```

It's just to make obvious why an "optional" parameter is actually required.

vlrpl:
> A few nits, but most importantly the completion for the `complete` command still does not work for me. Only `-h` and `--help` are available when hitting `<tab>` while we should see other parameters (eg. `--stack` or `--filter`):
> 
> ```
> # retis collect -<tab>
> -h      --help
> ```
> 
> > Please tell me if you still want to use generate-completion as subcommand.
> 
> I find `generate` to be too generic for a command that "only" generates a completion script. The issue with `complete` would be a conflict with `collect` for the `c` use though. If we want to use a sort command name we could go with `_complete`, but I'm not super sure.
> 
> An alternative could be to make something like `retis generate completion ...`.
> 
> @vlrpl WDYT?
> 

I think that naming is always hard :)
I agree that generate is too vague (whereas makes sense in a cli context, instead)
`retis generate completion` makes sense but it's hard to say in advance if generate will serve any other purpose in the future. I would stick with "one" subcommand (as a personal preference).
Maybe `sh-completion` (or `sh-complete` or something similar)?

> > There is no '-c' or '-o' auto-complete for collect subcommand. But I feel this is the collect parameter's problem.
> 
> Agreed. Same with `--profile`. This might be solved with dynamic completion once they also add support for custom implementation of authorized values.

atenart:
The PR looks mostly good, but still have the below issue.

> A few nits, but most importantly the completion for the `complete` command still does not work for me. Only `-h` and `--help` are available when hitting `<tab>` while we should see other parameters (eg. `--stack` or `--filter`):
> 
> ```
> # retis collect -<tab>
> -h      --help
> ```

I had a quick look at this and I'm confident the issue is because of this:

```rust
impl SubCommand for Collect {
    fn thin(&self) -> Result<Command> {                                         
        Ok(Command::new("collect").about("Collect network events"))             
    }
}
```

IMO the correct fix would be to define the same command as what's done in the `new` function from the same impl. With that we should have the correct completion for all arguments of `collect`. @amorenoz does that sound right?

We should even be able to generate the list of modules and set a `PossibleValuesParser` so we also have completion for the available collectors.

liuhangbin:
> IMO the correct fix would be to define the same command as what's done in the `new` function from the same impl. With that we should have the correct completion for all arguments of `collect`. @amorenoz does that sound right?

Hi @atenart , do you want to do this on this PR or a separate PR?

atenart:
> > IMO the correct fix would be to define the same command as what's done in the `new` function from the same impl. With that we should have the correct completion for all arguments of `collect`. @amorenoz does that sound right?
> 
> Hi @atenart , do you want to do this on this PR or a separate PR?

The first part is a requirement for the `collect` command to be supported by the auto-completion, as it's the most used command by far I think it should be done in this PR.

The second part, about the auto-completion of collectors for the `-c` arg, can be done in a follow-up PR or in this one; your choice. I think it should be done by the time we release 1.4 though.

liuhangbin:
> The first part is a requirement for the `collect` command to be supported by the auto-completion, as it's the most used command by far I think it should be done in this PR.

OK, I will see how to fix it.
> 
> The second part, about the auto-completion of collectors for the `-c` arg, can be done in a follow-up PR or in this one; your choice. I think it should be done by the time we release 1.4 though.

I'd prefer to fix it on a separate PR. I will put it on my next week's todo list. When 1.4 will be released?

atenart:
> > The first part is a requirement for the `collect` command to be supported by the auto-completion, as it's the most used command by far I think it should be done in this PR.
> 
> OK, I will see how to fix it.

Thanks!

> > The second part, about the auto-completion of collectors for the `-c` arg, can be done in a follow-up PR or in this one; your choice. I think it should be done by the time we release 1.4 though.
> 
> I'd prefer to fix it on a separate PR. I will put it on my next week's todo list. When 1.4 will be released?

Target is end of March.

liuhangbin:
> Target is end of March.

OK. 2 months should be enough.

liuhangbin:
> > > IMO the correct fix would be to define the same command as what's done in the `new` function from the same impl. With that we should have the correct completion for all arguments of `collect`. @amorenoz does that sound right?
> 
> The first part is a requirement for the `collect` command to be supported by the auto-completion, as it's the most used command by far I think it should be done in this PR.

Hi @atenart , I checked this issue. As you said, the `collect` only returns `thin` parameter when built. To fix it, we need to get all the parameters. So either we return full parameter for `collect`, like

```diff
diff --git a/src/collect/cli.rs b/src/collect/cli.rs
index 93294883275c..fc348d035f27 100644
--- a/src/collect/cli.rs
+++ b/src/collect/cli.rs
@@ -149,7 +149,7 @@ impl SubCommand for Collect {
     }

     fn thin(&self) -> Result<Command> {
-        Ok(Command::new("collect").about("Collect network events"))
+        self.full()
     }

     fn name(&self) -> String {
```

Or let all subcommands return full parameter.

```diff
diff --git a/src/cli/cli.rs b/src/cli/cli.rs
index 459783eeaff1..eb4c1939153b 100644
--- a/src/cli/cli.rs
+++ b/src/cli/cli.rs
@@ -271,9 +271,9 @@ impl ThinCli {
             .disable_help_subcommand(true)
             .infer_subcommands(true)
             .subcommand_required(true);
-        // Add thin subcommands so that the main help shows them.
+        // Add full subcommands so that the main help shows them.
         for sub in self.subcommands.iter() {
-            command = command.subcommand(sub.thin().expect("thin command failed"));
+            command = command.subcommand(sub.full().expect("full command failed"));
         }

         // Determine the subcommand that was run while ignoring errors from yet-to-be-defined
```

I'd prefer the second way, which should resolve all subcommand auto-completion. What do you think?

> 
> The second part, about the auto-completion of collectors for the `-c` arg, can be done in a follow-up PR or in this one; your choice. I think it should be done by the time we release 1.4 though.

For this one, if we could get the possible values during `let mut cli = get_cli()?.build();`, then we can do auto-completion. But this looks not easy as we only load the modules and replace the ran subcommand with the full subcommand when call the subcommand. Here is the code

```
    /// Perform full CLI parsing and validation
    pub(crate) fn run(mut self) -> Result<CliConfig, ClapError> {
        self.enhance_profile().map_err(|err| {
            self.command
                .error(ErrorKind::InvalidValue, format!("{err}"))
        })?;

...

        // Replace the ran subcommand with the full subcommand.
        self.command = self
            .command
            .mut_subcommand(self.subcommand.name(), |_| self.subcommand.full().unwrap());

``` 

 Do you have any idea?

atenart (changes requested):
Looks good. I'm sorry, one last thing: can you remove `fn thin()` from `impl SubCommand` otherwise this PR would "introduce" dead code.

If you can while at it, can you ellaborate a bit more in the doc about auto-completion? Maybe "Retis can generate completion files for shells (Bash, Zsh, Fish...). For example to enable auto-completion of Retis command in Bash, ...".

Thanks!

atenart (approved):
LGTM, thanks a lot for your contribution!

PR #330, opened 2024-01-15
Meta data: initial unit tests

vlrpl:
- picked the first patch from #328 (no reason that one over #329's other than order and avoiding creating a third version :)
- the second patch bails on bitfields match: there's no separate bail message, but it's fine as this is temporary and bitfields support is the next in line
- the third patch includes a bunch of tests

### DIFF
--- a/src/core/filters/meta/filter.rs
+++ b/src/core/filters/meta/filter.rs
@@ -67,7 +67,7 @@ struct MetaTarget {
 }
 
 #[repr(C)]
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 struct MetaLoad {
     // Type of data we're going to load
     // bit 0-4: [char|short|int|long], bit5: reserved, bit6: is_ptr, bit7: sign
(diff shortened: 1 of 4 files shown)


src/core/filters/meta/filter.rs
```diff
@@ -456,7 +462,7 @@ impl FilterMeta {
 
                     stored_offset = offset;
                 }
-                None => bail!("{field} not found!"),
+                None => bail!("{field} not found or is a bitfield!"),
```
atenart, line 465: Those two reasons seems a bit unrelated. Maybe `walk_btf_node` could return a `Result` instead of an `Option`, with proper explanations on why it failed to find a field?

vlrpl: > Those two reasons seems a bit unrelated. Maybe `walk_btf_node` could return a `Result` instead of an `Option`, with proper explanations on why it failed to find a field?

that's what I meant with `there's no separate bail message, but it's fine as this is temporary and bitfields support is the next in line`. For that reason, it was not turned into `Result`.
This part will be removed anyway. We can change the message to "or not supported" instead of turning `Option` to `Result` otherwise I guess `Result` will be temporary but fine.

atenart: Ah sorry, got it. That make sense.

atenart (approved):
LGTM, thanks!

PR #336, opened 2024-01-23
Add an authors file

atenart:
Add a script to auto-generate and check an AUTHORS.md file and add a CI job to check if the authors file is up-to-date so we don't miss new contributors.

Currently the authors file lists authors based on their git name in alphabetical order, which value depends on how the author configured its git name (could be name, first name, surname, random, etc). Feel free to make any suggestion if you feel this should be done otherwise (by # of commits, first contribution date[1], etc).

[1] Note: using `awk '!x[$0]++'` instead of `sort`.

### DIFF
--- a/tools/authors.sh
+++ b/tools/authors.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+help() {
+	echo "Usage:"
+	echo "  $0 [opts]"
+	echo ""
+	echo "  Generate an alphabetically ordered list of authors and update the authors file"
+	echo ""
+	echo "  Options:"
+	echo "    -c		Checks the authors file is valid without modifying it."
+	echo "    -h		Show this help."
+}
+
+while getopts "ch" opt; do
+	case $opt in
+	c) check_only=1 ;;
+	*) help; exit ;;
+	esac
+done
+shift $(($OPTIND - 1))
+
+authors_file=$(git rev-parse --show-toplevel)/AUTHORS.md
+authors_list="$(git log --no-merges --format='%aN' | grep -v dependabot | sort -u)"
+
+out=$(cat <<-EOM
+# Authors
+
+Many thanks to everyone who contributed to Retis!
+
+In alphabetical order:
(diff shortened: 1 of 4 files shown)


atenart:
Rebased on top of main add synced with current list of contributors.

tools/authors.sh
```diff
@@ -0,0 +1,40 @@
...
+while getopts "ch" opt; do
+	case $opt in
+	c) check_only=1 ;;
+	*) help; exit ;;
+	esac
+done
+shift $(($OPTIND - 1))
+
+authors_file=$(git rev-parse --show-toplevel)/AUTHORS.md
+authors_list="$(git log --format='%aN' | grep -v dependabot | sort -u)"
```
vlrpl, line 23: a nit that doesn't really change anything (doesn't happen for sure at the moment):

`--no-merges` is theoretically needed here (maintainer that only merges)

other kind of contributions (still valuable) should be recognized in a separate section

atenart: IMO that doesn't make an "author". But yes, merging PRs is very valuable :) If we want to recognize this we can have a static list of maintainers; as you suggested, in a separate section.

vlrpl: > IMO that doesn't make an "author". But yes, merging PRs is very valuable :) If we want to recognize this we can have a static list of maintainers; as you suggested, in a separate section.

agreed, I was suggesting `--no-merges` for that reason

atenart: Oh I see sorry. I was absolutely certain the above was done with `--no-merges`; all my local tests were. No idea how this got dropped but that is definitively needed. Thanks!

atenart: Or maybe I only used it when testing ordering by # of commits... Anyway, thanks for spotting this :)

vlrpl: well, the result is the same :)

atenart:
Rebased on top of `main` to resolve a conflict in the contributing doc.

vlrpl (approved):
lgtm, thanks!

PR #337, opened 2024-01-24
Wait for all probes to be installed before starting the collection

### LINKED ISSUE #312: Early tracking might be wrong
Probes are installed sequentially and not in order. When matching packets while not all probes are installed, especially the ones required for the tracking infrastructure, we might have wrong tracking information as free events could be missed.

One easy way to see this is to install lots of probes (eg. `-p skb*`) while filtering on an IPv4 address. Before all probes are installed, if packets are matches, we might see unexpected events (here IPv6 or wrong address).

One solution might be to control the order in which those important probes are being installed.

atenart:
Not doing so lead to various issues such as:
- Inconsistency about what events are reported for a given flow.
- Tracking issues as not all tracking probes are installed (leading to unrelated flows sharing the same tracking id, or stalled entries in the tracking map).
    
This also opens the door for having other shared configuration options among all probes.
    
Fixes #312. Includes the same probe manager reworks as #309 but does not depend on it.

### DIFF
--- a/src/core/probe/kernel/bpf/include/common.h
+++ b/src/core/probe/kernel/bpf/include/common.h
@@ -210,9 +210,19 @@ static __always_inline int chain(struct retis_context *ctx)
 	 * event usage length read before and after the hook chain.
 	 */
 	struct common_task_event *ti;
+	static bool enabled = false;
 	volatile u16 pass_threshold;
 	struct common_event *e;
 	struct kernel_event *k;
+
+	/* Check if the collection is enabled, otherwise bail out. Once we have
+	 * a positive result, cache it.
+	 */
+	if (unlikely(!enabled)) {
+		enabled = collection_enabled();
+		if (!enabled)
+			return 0;
+	}
 
 	cfg = bpf_map_lookup_elem(&config_map, &ctx->ksym);
 	if (!cfg)
--- a/src/core/probe/user/bpf/usdt.bpf.c
+++ b/src/core/probe/user/bpf/usdt.bpf.c
@@ -52,12 +52,22 @@ int probe_usdt(struct pt_regs *ctx)
 {
 	u64 pid = bpf_get_current_pid_tgid();
 	u64 sym_addr = PT_REGS_IP(ctx);
-	volatile u16 pass_threshold;
-	struct common_task_event *ti;
-	struct common_event *e;
 	struct retis_raw_event *event;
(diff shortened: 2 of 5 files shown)


vlrpl:
can you please rebase this?

atenart:
> can you please rebase this?

Sure, thanks for pinging me. Turns out git can rebase this automatically but not merge :)

src/core/probe/kernel/bpf/include/common.h
```diff
@@ -210,10 +210,20 @@ static __always_inline int chain(struct retis_context *ctx)
...
 	struct common_task_event *ti;
+	static bool enabled = false;
 	volatile u16 pass_threshold;
 	struct common_event *e;
 	struct kernel_event *k;
 
+	/* Check if the collection is enabled, otherwise bail out. Once we have
+	 * a positive result, cache it.
+	 */
+	if (!enabled) {
```
vlrpl, line 221: what about adding this in a common header (common_defs.h)?:

```c
#ifndef likely
#define likely(x)	__builtin_expect(!!(x), 1)
#endif

#ifndef unlikely
#define unlikely(x)	__builtin_expect(!!(x), 0)
#endif
```

and use it here?

src/core/probe/user/bpf/usdt.bpf.c
```diff
@@ -52,12 +52,22 @@ int probe_usdt(struct pt_regs *ctx)
...
+	struct user_ctx uctx = {};
 	struct common_event *e;
-	struct retis_raw_event *event;
 	struct user_event *u;
-	struct user_ctx uctx = {};
+
+	/* Check if the collection is enabled, otherwise bail out. Once we have
+	 * a positive result, cache it.
+	 */
+	if (!enabled) {
```
vlrpl, line 66: ditto

src/core/probe/kernel/bpf/include/common.h
atenart: I think that makes sense; if that indeed makes a difference when building BPF programs (it seems some are using that so I guess so).

vlrpl: it makes a difference in the build

atenart:
Rebased on top of main to fix a linter issue. No other change intended.

vlrpl (approved):
lgtm, thanks!

PR #338, opened 2024-01-25
Parse protocols in Rust using the raw packets

### LINKED ISSUE #283: Evaluate parsing packets in the Rust side
With the `--skb-sections packet` option, we can now report raw packets to userspace. But asking for fields to be reported still enables parsing in the BPF side. We could instead always report raw packets and extra information from the Rust side directly.

A few extra points:
- We can optimize the section being reported from kernel space based on fields being asked.
- We might want to refine the `--skb-sections` option to have less control. Eg. only controlling the level of information to display (l2, l3, l4).
- Performances should be measured.
- But also doing this will enable parsing more data from packets, especially in early Rx where some offsets aren't set. So it's a balance.

atenart:
Fixes #283.

Instead of parsing protocol headers in the BPF side and reporting them through raw event sections, take advantage of the fact that we can report the (start of the) raw packet and parse it to generate the skb event (protocol) sections.

This has many advantages, summed up below:
- Using skb offsets in BPF limits what protocol sections can be reported, eg. in the IP layer in Rx we can't report TCP / UDP. Using the raw packet makes this possible as data is usually there and valid (raw packet retrieval doesn't have the same limitations regarding the skb offsets). For Tx this doesn't change much.
- This allows to have less instructions in BPF and to not report the same data twice from BPF to Rust.
- Surprisingly, performance in the Rust side only was better (parsing raw packet vs parsing multiple raw sections), see below.
- PCAP generation is now easier (raw packet is always included in the output file, no need to set specific options -some can improve the pcap generation experience but are not mandatory-).
- Easier to add support for new protocols (some will come once this is merged, that is the next step).

A few functional changes:
- Some `--skb-sections` values are now deprecated as they are always on. This is because for example we need to parse the IPv4 header anyway if we want TCP information. Copying a few fields isn't much. Only `eth` can be skipped as it's not always useful. Another reason is there are not many use cases for not reporting L3+L4 fields. If required we can add later on a limit on the max layer level to report.
- The raw packet is always reported (we can't only report the dev or ns part alone anymore). I took the opportunity to simplify things a bit since this was probably not useful.

### A note on the additional dependency

A dependency on `pnet` was added as it is supporting many protocols, is quite fast, and has ~lots of users. There was also no other crate I could find that checked all the boxes. Having said that, it's not as active as it was and the maintainer expressed their will to let someone take over (a long time ago but the issue is still opened). There was still a release in mid-2023.

Pros:
- It was easier to implement a first PR using an available crate, at least to validate the intent.
- It's not like already supported protocols in `pnet` will change a lot over the years.
- It's quite easy to replace the `pnet` use in Retis with another crate if the need arise.

Cons:
- It's not easy to find a replacement.
- Some doubts about long term efforts in `pnet`; eg. can we easily add support for parsing new protocols? Will it be maintained over time?
- `pnet` is not packaged in Fedora.

IMO things are not perfect, but are not too bad for now. It's probably acceptable. Tbf the other choice was to write a crate to parse packets myself (which I'm willing to do if the need arises or if you think we should do this from day one). If maintenance of `pnet` becomes an issue we could also vendor it, at least until we have a replacement.

### Performances of parsing events (Rust part only)

**Before**
Average first event parsing: 287µs
Average 1M events parsing: 1609080µs

**After**
Average first event parsing: 277µs
Average 1M events parsing: 1532422µs


amorenoz:
Sorry I haven't got the time to look at this deeply. I would like to understand the implications. Assuming previously users likely did not enable `packet` section:
- Events files are now bigger (they contain both raw and parsed headers). This will probably take longer to write (and not super fan of adding binary data as chars in JSON :-D)
- ebpf events are the same size (1024B). But, is there more or less space? Meaning: how does packet data size compares with all the previous section events?

Am I missing something?

atenart:
> Assuming previously users likely did not enable `packet` section:
> 
>     * Events files are now bigger (they contain both raw and parsed headers). This will probably take longer to write

That is right. This is not a requirement to get the functionality introduced by this PR but was more a simplification. This is handy and less error prone is various cases such as pcap generation or the upcoming python post-process.

However if there are concerns about performances writing the raw packet in the events file could still be opt-in.

> (and not super fan of adding binary data as chars in JSON :-D)

This is not adding this (pcap support did), but I get the point :) We can probably use `base64` instead (and break compatibility, not that I think it is an issue).

>     * ebpf events are the same size (1024B). But, is there more or less space? Meaning: how does packet data size compares with all the previous section events?

An Eth/IPv4/TCP packet w/o the raw packet was 57B. Now with the raw packet only it's 263B. In case the raw packet was retrieved previously, the size is now smaller. In many cases the raw packet length ending up in the events file is smaller than that. I intentionally did not try to get too smart in the BPF side about this as otherwise parsing the packet would depend on where we are in the stack.

Note that the raw packet section embedding 255B of the data is arbitrary and was chosen to be enough for encapsulation cases, but this could be revisited and we could probably chose something between 128B - 255B to save some space.

atenart:
Rebased on top of main to resolve merge conflicts.

src/module/skb/bpf.rs
```diff
@@ -379,12 +366,22 @@ pub(super) struct RawPacketEvent {
...
         len: raw.len,
         capture_len: raw.capture_len,
         packet: raw.packet[..(raw.capture_len as usize)].to_vec(),
-    })
+    });
+
+    // Then start parsing the raw packet to generate other sections.
+    let eth = match EthernetPacket::new(&raw.packet[..(raw.capture_len as usize)]) {
+        Some(eth) => eth,
+        None => return Ok(()),
```
amorenoz, line 274: Shouldn't we return some kind of error if we cannot parse the packet?

src/module/skb/bpf.rs
```diff
@@ -318,24 +296,40 @@ pub(super) fn unmarshal_packet(event: &mut SkbEvent, raw_section: &BpfRawSection
...
+                unmarshal_l3(event, ip.get_next_header(), ip.payload())?;
+            };
+        }
+        _ => (),
+    }
+
+    Ok(())
+}
+
+fn unmarshal_l3(
```
amorenoz, line 305: Funny name for a function that unmarshals the L4 header :-)

amorenoz (commented):
A couple of nits. Apart from that the code LGTM.
I would like to see a performance test that actually saves and reads from file which is now considerably bigger (on the default configuration). Do you think that's easy to do with current benchmark system?

src/module/skb/bpf.rs
atenart: :rofl:

atenart:
> I would like to see a performance test that actually saves and reads from file which is now considerably bigger (on the default configuration). Do you think that's easy to do with current benchmark system?

I think that's actually quite hard, the problem itself is not an easy one. We could see the impact of reading or writing more data, which can be done outside of Retis but that is not so easy I think, there are lots of things into play. Result might be similar, or a bit worst as we write more (note that with #377 raw packet is base64 encoded and is not that big).

But the real question is: is there an impact on the performance limiting part? We don't really know yet which one of the BPF<>Rust (+ parsing) part and the event printing + writing one is the limiting factor, although it's very likely the Rust side is (as it's the dequeuing/parsing side). Moreover, current benchmark commands are designed to optimize a specific part, but not really to study the impact of a change globally.

In addition, the whole skb BPF event was changed (less data) and its parsing is completely different (and moved in Rust). That also has an impact overall.

I did a test though: collection of events using the generic profile while using a high tput TCP transfer, and looked at dropped events. That's the best indicator we have IMO. Both versions (main and #377) had similar numbers (in the 300k-700k) with the main one having slightly better numbers in general.

On top of this looked at flamegraphs: the write is done by write_all -> write (blocking) and that takes only 3.5% of the total time spent.

I don't think ready/writing slightly more bits in the event file has a big impact on performances. But the overall change, with moving the packet parsing in Rust probably has. It might be more efficient in comparison, but this puts more pressure on the Rust side and IMO has an impact overall. This could be parallelized and would bring nice results, not that it's an easy topic (eg. how to order events).

amorenoz (approved):

PR #340, opened 2024-02-02
Bpf logger

vlrpl:
Add a dedicated trace pipe for better debugging eBPF progs and signaling unexpected errors.

### DIFF
--- a/src/core/events/bpf.rs
+++ b/src/core/events/bpf.rs
@@ -14,7 +14,7 @@ use std::{
 };
 
 use anyhow::{anyhow, bail, Result};
-use log::error;
+use log::{error, log, Level};
 use plain::Plain;
 
 use super::{Event, EventResult};
@@ -73,10 +73,12 @@ macro_rules! event_byte_array {
 #[cfg(not(test))]
 pub(crate) struct BpfEventsFactory {
     map: libbpf_rs::MapHandle,
+    log_map: libbpf_rs::MapHandle,
     /// Receiver channel to retrieve events from the processing loop.
     rxc: Option<mpsc::Receiver<Event>>,
     /// Polling thread handle.
     handle: Option<thread::JoinHandle<()>>,
+    log_handle: Option<thread::JoinHandle<()>>,
     run_state: Running,
 }
 
@@ -87,7 +89,6 @@ impl BpfEventsFactory {
             sz: mem::size_of::<libbpf_sys::bpf_map_create_opts>() as libbpf_sys::size_t,
             ..Default::default()
         };
-
         let map = libbpf_rs::MapHandle::create(
             libbpf_rs::MapType::RingBuf,
             Some("events_map"),
@@ -98,10 +99,26 @@ impl BpfEventsFactory {
         )
         .or_else(|e| bail!("Failed to create events map: {}", e))?;
 
+        let opts = libbpf_sys::bpf_map_create_opts {
+            sz: mem::size_of::<libbpf_sys::bpf_map_create_opts>() as libbpf_sys::size_t,
+            ..Default::default()
+        };
+        let log_map = libbpf_rs::MapHandle::create(
+            libbpf_rs::MapType::RingBuf,
+            Some("log_map"),
+            0,
+            0,
+            mem::size_of::<LogEvent>() as u32 * BPF_LOG_EVENTS_MAX,
+            &opts,
+        )
+        .or_else(|e| bail!("Failed to create log map: {}", e))?;
+
         Ok(BpfEventsFactory {
             map,
+            log_map,
             rxc: None,
             handle: None,
+            log_handle: None,
             run_state: Running::new(),
         })
     }
@@ -109,6 +126,42 @@ impl BpfEventsFactory {
     /// Get the events map fd for reuse.
     pub(crate) fn map_fd(&self) -> RawFd {
         self.map.as_fd().as_raw_fd()
+    }
+
+    /// Get the log map fd for reuse.
+    pub(crate) fn log_map_fd(&self) -> RawFd {
+        self.log_map.as_fd().as_raw_fd()
+    }
+
+    fn ringbuf_handler<CB>(
+        &self,
+        map: &libbpf_rs::MapHandle,
+        rb_handler: CB,
+    ) -> Result<thread::JoinHandle<()>>
+    where
+        CB: FnMut(&[u8]) -> i32 + 'static,
+    {
+        let mut rb = libbpf_rs::RingBufferBuilder::new();
+        rb.add(map, rb_handler)?;
+        let rb = rb.build()?;
+        let rs = self.run_state.clone();
+        // Start an event polling thread.
+        Ok(thread::spawn(move || {
+            while rs.running() {
+                if let Err(e) = rb.poll(Duration::from_millis(BPF_EVENTS_POLL_TIMEOUT_MS)) {
+                    match e.kind() {
+                        // Received EINTR while polling the
+                        // ringbuffer. This could normally be
+                        // triggered by an actual interruption
+                        // (signal) or artificially from the
+                        // callback. Exit without printing any error.
+                        libbpf_rs::ErrorKind::Interrupted => (),
+                        _ => error!("Unexpected error while polling ({e})"),
+                    }
+                    break;
+                }
+            }
+        }))
     }
 }
 
@@ -153,29 +206,52 @@ impl EventFactory for BpfEventsFactory {
             0
         };
 
-        // Finally make our ring buffer and associate our map to our event
-        // processing closure.
-        let mut rb = libbpf_rs::RingBufferBuilder::new();
-        rb.add(&self.map, process_event)?;
-        let rb = rb.build()?;
-        let rs = self.run_state.clone();
-        // Start an event polling thread.
-        self.handle = Some(thread::spawn(move || {
-            while rs.running() {
-                if let Err(e) = rb.poll(Duration::from_millis(BPF_EVENTS_POLL_TIMEOUT_MS)) {
-                    match e.kind() {
-                        // Received EINTR while polling the
-                        // ringbuffer. This could normally be
-                        // triggered by an actual interruption
-                        // (signal) or artificially from the
-                        // callback. Exit without printing any error.
-                        libbpf_rs::ErrorKind::Interrupted => (),
-                        _ => error!("Unexpected error while polling ({e})"),
-                    }
-                    break;
-                }
+        let run_state = self.run_state.clone();
+        // Closure to handle the log events coming from the BPF part.
+        let process_log = move |data: &[u8]| -> i32 {
+            if data.len() != mem::size_of::<LogEvent>() {
+                error!("Unexpected log event size");
+                return 0;
             }
-        }));
+            // If a termination signal got received, return (EINTR)
+            // from the callback in order to trigger the event thread
+            // termination. This is useful in the case we're
+            // processing a huge number of buffers and rb.poll() never
+            // times out.
+            if !run_state.running() {
+                return -4;
+            }
+
+            let mut log_event = LogEvent::default();
+            if let Err(e) = plain::copy_from_bytes(&mut log_event, data) {
+                error!("Can't read eBPF log event {:?}", e);
+                return 0;
+            }
+
+            let log_level = match log_event.level {
+                1 => Level::Error,
+                2 => Level::Warn,
+                3 => Level::Info,
+                4 => Level::Debug,
+                5 => Level::Trace,
+                l => {
+                    error!("Unexpected log level ({l}). Falling back to error level");
+                    Level::Error
+                }
+            };
+
+            match log_event.msg.to_string() {
+                Ok(msg) => log!(log_level, "[eBPF] {msg}"),
+                Err(e) => error!("Unable to convert eBPF log string: {e}"),
+            }
+
+            0
+        };
+
+        // Finally make our ring buffers and associate maps to their
+        // respective events processing closure.
+        self.handle = Some(self.ringbuf_handler(&self.map, process_event)?);
+        self.log_handle = Some(self.ringbuf_handler(&self.log_map, process_log)?);
 
         Ok(())
     }
@@ -183,14 +259,16 @@ impl EventFactory for BpfEventsFactory {
     /// Stops the event polling mechanism. The dedicated thread is stopped
     /// joining the execution
     fn stop(&mut self) -> Result<()> {
-        match self.handle.take() {
-            Some(th) => {
-                self.run_state.terminate();
-                th.join()
-                    .or_else(|_| bail!("while joining bpf event thread"))
-            }
-            None => Ok(()),
-        }
+        self.handle.take().map_or(Ok(()), |th| {
+            self.run_state.terminate();
+            th.join()
+                .map_err(|_| anyhow!("while joining bpf event thread"))
+        })?;
+
+        self.log_handle.take().map_or(Ok(()), |th| {
+            th.join()
+                .map_err(|_| anyhow!("while joining bpf log event thread"))
+        })
     }
 
     /// Retrieve the next event. This is a blocking call and never returns EOF.
@@ -443,6 +521,26 @@ impl EventFactory for BpfEventsFactory {
         Ok(())
     }
 }
+
+/// Size of msg for a single log event. Please keep in sync with its
+/// BPF counterpart.
+pub(super) const BPF_LOG_MAX: usize = 127;
+
+/// Max number of log events we can store at once in the shared map. Please keep in
+/// sync with its BPF counterpart.
+pub(super) const BPF_LOG_EVENTS_MAX: u32 = 32;
+
+event_byte_array!(BpfLogMsg, BPF_LOG_MAX);
+
+/// Log event. Please keep in sync with its BPF counterpart.
+#[derive(Default)]
+#[repr(C, packed)]
+pub(super) struct LogEvent {
+    level: u8,
+    msg: BpfLogMsg,
+}
+
+unsafe impl Plain for LogEvent {}
 
 /// Max number of events we can store at once in the shared map. Please keep in
 /// sync with its BPF counterpart.
--- a/src/core/events/bpf.rs
+++ b/src/core/events/bpf.rs
@@ -50,8 +50,8 @@ macro_rules! event_byte_array {
         #[allow(dead_code)]
         impl $name {
             fn to_string(&self) -> Result<String> {
-                Ok(std::str::from_utf8(&self.0)?
-                    .trim_end_matches(char::from(0))
+                Ok(std::ffi::CStr::from_bytes_until_nul(&self.0)?
+                    .to_str()?
                     .into())
             }
 
--- a/src/core/events/bpf/include/events.h
+++ b/src/core/events/bpf/include/events.h
@@ -3,6 +3,8 @@
 
 #include <vmlinux.h>
 
+#include <common_defs.h>
+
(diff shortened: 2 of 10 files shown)


atenart:
I just skimmed through the code; a few suggestion though:

- This functionality looks nice for `> debug` messages. However including debug messages unconditionally can fill the buffer and hide higher messages. Debug messages are really only needed for development, what about hiding it under a compilation flag (`-D`) only set for the debug build?
- Following the above, debug messages could end up being quite verbose in some cases. It's also logically tricky to use the same application we're debugging for reporting debug messages. What about using `bpf_printk` for debug messages?
- It looks like all eBPF messages are reported to userspace whatever the actual log level is. I think you could benefit from the global eBPF configuration introduced in #337 to configure the probes log level. This could be done using the same logic as in #337: once probes are enabled, we can set a global variable and use it afterwards).

vlrpl:
> I just skimmed through the code; a few suggestion though:
> 
>     * This functionality looks nice for `> debug` messages. However including debug messages unconditionally can fill the buffer and hide higher messages. Debug messages are really only needed for development, what about hiding it under a compilation flag (`-D`) only set for the debug build?
> 

debug messages must not be included unconditionally. This is for error path or dev usage (see the comment before the log macros).
We could strip them away from the compilation, or, considering this series is not in its final version (it lacks some usage and the "hiding" part) , the plan was to carry the loglevel down to the bpf progs and branch them behind a var in .rodata.

>     * Following the above, debug messages could end up being quite verbose in some cases. It's also logically tricky to use the same application we're debugging for reporting debug messages. What about using `bpf_printk` for debug messages?

not a fan of bpf_printk. It's mostly the same thing except it's a global trace pipe (also I'm not sure it's great from ux point of view).

> 
>     * It looks like all eBPF messages are reported to userspace whatever the actual log level is. I think you could benefit from the global eBPF configuration introduced in [Wait for all probes to be installed before starting the collection #337](https://github.com/retis-org/retis/pull/337) to configure the probes log level. This could be done using the same logic as in [Wait for all probes to be installed before starting the collection #337](https://github.com/retis-org/retis/pull/337): once probes are enabled, we can set a global variable and use it afterwards).

this is similar to what I had in mind.

atenart:
> > I just skimmed through the code; a few suggestion though:
> > ```
> > * This functionality looks nice for `> debug` messages. However including debug messages unconditionally can fill the buffer and hide higher messages. Debug messages are really only needed for development, what about hiding it under a compilation flag (`-D`) only set for the debug build?
> > ```
> 
> debug messages must not be included unconditionally. This is for error path or dev usage (see the comment before the log macros). We could strip them away from the compilation, or, considering this series is not in its final version (it lacks some usage and the "hiding" part) , the plan was to carry the loglevel down to the bpf progs and branch them behind a var in .rodata.

Makes sense. One way could be to set the rodata option before loading the BPF programs, this way debug messages are always compiled and available but are stripped when loading the program if disabled.

> > ```
> > * Following the above, debug messages could end up being quite verbose in some cases. It's also logically tricky to use the same application we're debugging for reporting debug messages. What about using `bpf_printk` for debug messages?
> > ```
> 
> not a fan of bpf_printk. It's mostly the same thing except it's a global trace pipe (also I'm not sure it's great from ux point of view).

I agree, handling debug messages in the same way we would handle other levels is good for UX and to keep things aligned. But that also makes the debug messages to rely on at least some parts of Retis to work properly. It's not a blocker, but wanted to have this discussion because it's a tradeoff to be aware of.

Maybe one way to still be able to debug Retis from an outside point of view would be to design this so that adding the following works:

```C
define debug bpf_printk
```

> > ```
> > * It looks like all eBPF messages are reported to userspace whatever the actual log level is. I think you could benefit from the global eBPF configuration introduced in [Wait for all probes to be installed before starting the collection #337](https://github.com/retis-org/retis/pull/337) to configure the probes log level. This could be done using the same logic as in [Wait for all probes to be installed before starting the collection #337](https://github.com/retis-org/retis/pull/337): once probes are enabled, we can set a global variable and use it afterwards).
> > ```
> 
> this is similar to what I had in mind.

Nice!

vlrpl:
> > > I just skimmed through the code; a few suggestion though:
> > > ```
> > > * This functionality looks nice for `> debug` messages. However including debug messages unconditionally can fill the buffer and hide higher messages. Debug messages are really only needed for development, what about hiding it under a compilation flag (`-D`) only set for the debug build?
> > > ```
> > 
> > 
> > debug messages must not be included unconditionally. This is for error path or dev usage (see the comment before the log macros). We could strip them away from the compilation, or, considering this series is not in its final version (it lacks some usage and the "hiding" part) , the plan was to carry the loglevel down to the bpf progs and branch them behind a var in .rodata.
> 
> Makes sense. One way could be to set the rodata option before loading the BPF programs, this way debug messages are always compiled and available but are stripped when loading the program if disabled.

yes, that's what I was thinking about.
I agree with all the points in the feedback and the idea is mostly to converge in that direction. Useful especially I wanted to discuss the `.rodata` approach as it kind of overlaps with other global configs (requiring a lookup).

> 
> > > ```
> > > * Following the above, debug messages could end up being quite verbose in some cases. It's also logically tricky to use the same application we're debugging for reporting debug messages. What about using `bpf_printk` for debug messages?
> > > ```
> > 
> > 
> > not a fan of bpf_printk. It's mostly the same thing except it's a global trace pipe (also I'm not sure it's great from ux point of view).
> 
> I agree, handling debug messages in the same way we would handle other levels is good for UX and to keep things aligned. But that also makes the debug messages to rely on at least some parts of Retis to work properly. It's not a blocker, but wanted to have this discussion because it's a tradeoff to be aware of.
> 

makes sense

> Maybe one way to still be able to debug Retis from an outside point of view would be to design this so that adding the following works:
> 
> ```c
> #define debug bpf_printk
> ```

I guess it would work, or at least can be done

src/core/events/bpf.rs
```diff
@@ -152,44 +205,69 @@ impl EventFactory for BpfEventsFactory {
...
+            // termination. This is useful in the case we're
+            // processing a huge number of buffers and rb.poll() never
+            // times out.
+            if !run_state.running() {
+                return -4;
+            }
+
+            let mut log_event = LogEvent::default();
+            if plain::copy_from_bytes(&mut log_event, data).is_err() {
+                error!("Can't read eBP log event");
```
atenart, line 225: s/eBP/eBPF/

src/core/events/bpf.rs
```diff
@@ -152,44 +205,69 @@ impl EventFactory for BpfEventsFactory {
...
+            // from the callback in order to trigger the event thread
+            // termination. This is useful in the case we're
+            // processing a huge number of buffers and rb.poll() never
+            // times out.
+            if !run_state.running() {
+                return -4;
+            }
+
+            let mut log_event = LogEvent::default();
+            if plain::copy_from_bytes(&mut log_event, data).is_err() {
```
atenart, line 224: Please use `if let Err(e) ...` so the error can be shown in the log msg.

src/core/events/bpf.rs
```diff
@@ -152,44 +205,69 @@ impl EventFactory for BpfEventsFactory {
...
+                l => {
+                    error!("Unexpected log level ({l}). Falling back to error level");
+                    Level::Error
                 }
+            };
+
+            if let Ok(msg) = log_event.data.to_string() {
+                log!(log_level, "[eBPF] {msg}");
+            } else {
+                error!("Unable to convert eBPF log string");
```
atenart, line 244: Ditto here. Using `match` can help getting the underlying error and show it in the log msg.

src/core/events/bpf.rs
```diff
@@ -109,6 +126,42 @@ impl BpfEventsFactory {
...
+    /// Get the log map fd for reuse.
+    pub(crate) fn log_map_fd(&self) -> RawFd {
+        self.log_map.as_fd().as_raw_fd()
+    }
+
+    fn ringbuf_handler<CB>(
+        &self,
+        map: &libbpf_rs::MapHandle,
+        rb_handler: CB,
+    ) -> Result<Option<thread::JoinHandle<()>>>
```
atenart, line 139: I would prefer to not use `Option` here as the function can't return `None`. Also, could you add a small comment above the fct definition?

src/core/events/bpf.rs
```diff
@@ -152,44 +205,69 @@ impl EventFactory for BpfEventsFactory {
...
-        }
+        self.handle.take().map_or(Ok(()), |th| {
+            self.run_state.terminate();
+            th.join()
+                .map_err(|_| anyhow!("while joining bpf event thread"))
+        })?;
+
+        self.log_handle.take().map_or(Ok(()), |th| {
+            th.join()
+                .map_err(|_| anyhow!("while joining bpf event log thread"))
```
atenart, line 269: s/event log/log/ or /log event/

src/core/events/bpf.rs
```diff
@@ -426,6 +504,26 @@ impl EventFactory for BpfEventsFactory {
...
 
+/// Size of data for a single log event. Please keep in sync with its
+/// BPF counterpart.
+pub(super) const BPF_LOG_MAX: usize = 128;
+
+/// Max number of log events we can store at once in the shared map. Please keep in
+/// sync with its BPF counterpart.
+pub(super) const BPF_LOG_EVENTS_MAX: u32 = 32;
+
+event_byte_array!(BpfLogData, BPF_LOG_MAX - 1);
```
atenart, line 515: Why `- 1`? Is it to align the log even taking into account the level? If so, `BPF_LOG_MAX` should include that -1.

Same in the BPF side.

src/core/probe/bpf/include/common_defs.h
```diff
@@ -32,6 +31,52 @@ static __always_inline bool collection_enabled() {
...
+ * {error,slow} path.
+ * Useful exceptions must use a high log level (ideally LOG_TRACE).
+ */
+#define retis_log(lvl, fmt, args...)					\
+({									\
+	struct retis_log_event *__log = NULL;				\
+	if (retis_log_enabled(lvl)) {					\
+		__log = bpf_ringbuf_reserve(&log_map, sizeof(struct retis_log_event), 0); \
+		if (__log) {						\
+			__builtin_memset(__log, 0, sizeof(struct retis_log_event)); \
```
atenart, line 65: We might be able to remove this, as `__log->level` and `log->data` are set below. BPF_SNPRINTF sets a `\0` at the end of the formatted string (or at the end of the buffer if the formatted str size equals or is above it).

src/core/events/bpf/include/events.h
```diff
@@ -3,11 +3,22 @@
...
 #define RAW_EVENT_DATA_SIZE	1024 - 2 /* Remove the size field */
 #define RETIS_MAX_COMM		64
 
+/* Please keep the below in sync with its Rust counterpart. */
+#define LOG_MAX			128
+#define LOG_EVENTS_MAX		32
+
+struct retis_log_event {
+	u8 level;
+	u8 data[LOG_MAX - 1];
```
atenart, line 19: s/data/message/?

src/core/probe/bpf/include/common_defs.h
```diff
@@ -32,6 +31,52 @@ static __always_inline bool collection_enabled() {
...
+/* Log macros must be used carefully and preferrably in the
+ * {error,slow} path.
+ * Useful exceptions must use a high log level (ideally LOG_TRACE).
+ */
+#define retis_log(lvl, fmt, args...)					\
+({									\
+	struct retis_log_event *__log = NULL;				\
+	if (retis_log_enabled(lvl)) {					\
+		__log = bpf_ringbuf_reserve(&log_map, sizeof(struct retis_log_event), 0); \
+		if (__log) {						\
```
atenart, line 64: If `__log` is `NULL`, could we use a counter to inform the user at the end of the collect? And we also probably don't need to return `__log` at the end of `retis_log`.

src/core/probe/bpf/include/common_defs.h
```diff
@@ -32,6 +31,52 @@ static __always_inline bool collection_enabled() {
...
+	LOG_INFO,
+	LOG_DEBUG,
+	LOG_TRACE,
+};
+
+/* Checks if the level will be reported. */
+#define retis_log_enabled(__lvl)			\
+({							\
+	bool active = false;				\
+	if (log_level && __lvl <= log_level)		\
```
atenart, line 50: Do you think allowing no log at all has benefits? Otherwise we can simplify the above (removing it and open coding it below) and defaults `log_level` to `LOG_ERROR`.

src/core/events/bpf/include/events.h
```diff
@@ -69,8 +86,10 @@ static __always_inline void *get_event_section(struct retis_raw_event *event,
 	u16 left = RAW_EVENT_DATA_SIZE - event->size;
 	void *section;
 
-	if (sizeof(*header) + size > left || event->size > sizeof(event->data))
+	if (sizeof(*header) + size > left || event->size > sizeof(event->data)) {
+		log_error("Failed to get event section");
```
atenart, line 90: \+ ": no space left"

atenart (changes requested):

src/core/events/bpf.rs
vlrpl: plain::error doesn't implement Display. Are you suggesting to use Debug?

src/core/events/bpf.rs
vlrpl: this is a bit different from the previous one. Ok for `match` usage.

src/core/probe/bpf/include/common_defs.h
vlrpl: uhm, that's right. I think I saw weird things w/o the `memset()` because of the improper cstr representation using `.trim_end_matches()`

src/core/events/bpf.rs
vlrpl: don't recall the exact reason I made I left it out. But I don't see any to keep it.

src/core/events/bpf.rs
atenart: Ah right. Up to you then, to keep it as-is or use Debug.

src/core/probe/bpf/include/common_defs.h
vlrpl: `log_level` should never be 0 unless we explicitly set it so with `set_max_level()`.
It always defaults to `log::max_level()` and so the same level of user space.
In general, I think it's nice to have such a direct mapping.

Maybe we want to rename `retis_log_enabled()` to `retis_log_allowed()` (I'm open to proposals :) as the current name doesn't perfectly describe well the macro?

atenart: > `log_level` should never be 0 unless we explicitly set it so with `set_max_level()`. It always defaults to `log::max_level()` and so the same level of user space. In general, I think it's nice to have such a direct mapping.

Right, the point was then we can remove the `if (log_level && ` part from above as it's always true. And then we can open code it for better readability. In addition, to make the default clear we can initialize `log_level` to `LOG_ERROR`.

> Maybe we want to rename `retis_log_enabled()` to `retis_log_allowed()` (I'm open to proposals :) as the current name doesn't perfectly describe well the macro?

I don't have a strong feeling on this one :)

src/core/events/bpf/include/events.h
vlrpl: renamed it to `msg`

src/core/probe/bpf/include/common_defs.h
vlrpl: the problem is that counters need pid (for usdt probes, 0 otherwise) and `sym_addr` (target).
Probably we should wrap logs functions and, based on the "return" increase the counter.
As of now, message drop should never happen (or if it happens we'll see a good number of repeating logs before).
I think this will have some follow-ups anyways, as we might want to consider some rate limit mechanism and at this point global counters (or different approaches). WDYT?

src/core/events/bpf.rs
vlrpl: Just to make it clear, are you suggesting:
```
self.handle = Some(self.ringbuf_handler(&self.map, process_event)?);
self.log_handle = Some(self.ringbuf_handler(&self.log_map, process_log)?);
```

`Option` is used for `BpfEventsFactory` (essentially `Default`)

src/core/events/bpf.rs
vlrpl: I think we should be fine with `Debug`

src/core/probe/bpf/include/common_defs.h
vlrpl: > > `log_level` should never be 0 unless we explicitly set it so with `set_max_level()`. It always defaults to `log::max_level()` and so the same level of user space. In general, I think it's nice to have such a direct mapping.
> 

makes sense to remove `log_level &&` as it's not really useful.

> Right, the point was then we can remove the `if (log_level && ` part from above as it's always true. And then we can open code `retis_log_enabled` for better readability. In addition, to make the default clear we can initialize `log_level` to `LOG_ERROR`.
> 
> > Maybe we want to rename `retis_log_enabled()` to `retis_log_allowed()` (I'm open to proposals :) as the current name doesn't perfectly describe well the macro?
> 
> I don't have a strong opinion on this one :)

Let's open code it and not think about the name  :)

src/core/probe/bpf/include/common_defs.h
atenart: Makes sense, let's not use a counter for now.

src/core/probe/bpf/include/common_defs.h
vlrpl: about `log_level` to `LOG_ERROR`, user space defaults to LOG_INFO. That would be misaligned.

vlrpl:
there still are some open questions (minor things).
Sent a new version in the meantime.

src/core/probe/bpf/include/common_defs.h
atenart: Right, we should make the two default to the same value at least here for the initial submission.

vlrpl:
should be ready now

src/core/probe/bpf/include/common_defs.h
```diff
@@ -32,6 +31,42 @@ static __always_inline bool collection_enabled() {
...
+	if (lvl <= log_level) {						\
+		__log = bpf_ringbuf_reserve(&log_map, sizeof(struct retis_log_event), 0); \
+		if (__log) {						\
+			__log->level = lvl;				\
+			BPF_SNPRINTF(__log->msg, sizeof(__log->msg), fmt, args); \
+			bpf_ringbuf_submit(__log, BPF_RB_FORCE_WAKEUP);	\
+		}							\
+	}								\
+	__log;								\
+})
```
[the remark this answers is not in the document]
vlrpl: if we want to wrap the macro and increment counters, we'd need `__log` (that's the reason it stayed).
Admittedly, initializing it to NULL is not what we want, but any other value now is a bit too early.
Let's "return" `__log` once/if we introduce counters.

atenart (approved):
LGTM, this will be very helpful, thanks!

atenart (approved):
LGTM, thanks!

src/core/filters/meta/bpf/include/meta_filter.h
```diff
@@ -218,8 +218,10 @@ unsigned int meta_filter(struct sk_buff *skb)
 
 	ctx.base = skb;
 
-	if (meta_process_ops(&ctx) < 0 || !ctx.data)
+	if (meta_process_ops(&ctx) < 0 || !ctx.data) {
+		log_error("Failed to process meta-filter ops");
```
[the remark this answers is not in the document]
atenart: This can be discussed in a dedicated PR.

PR #341, opened 2024-02-06
Remove no longer needed macro

vlrpl:
The macro is provided by the header we consume.
There's no longer a need to define a local version.

### DIFF
--- a/src/module/ovs/bpf/include/ovs_common.h
+++ b/src/module/ovs/bpf/include/ovs_common.h
@@ -2,15 +2,9 @@
 #define __MODULE_OVS_COMMON__
 
 #include <bpf/bpf_core_read.h>
+#include <bpf/bpf_helpers.h>
 
 #include "jhash.h"
-
-/* <bpf/bpf_helpers.h> now includes the below definition, define it here if not
- * found for the time being to allow a smooth transition.
- */
-#ifndef barrier_var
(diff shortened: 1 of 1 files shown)


atenart:
This was done because IIRC the definition of `barrier_var` in `bpf_helpers.h` was recent and not in every distro where Retis is built. I guess with #237 this will be resolved. So if I got this correctly, we can merge it once the cited PR is merged.

vlrpl:
> This was done because IIRC the definition of `barrier_var` in `bpf_helpers.h` was recent and not in every distro where Retis is built. I guess with #237 this will be resolved. So if I got this correctly, we can merge it once the cited PR is merged.

we already use vendored headers exposed by libbpf-sys, so we should be fine already. I would be surprised if we use system's

vlrpl:
> > This was done because IIRC the definition of `barrier_var` in `bpf_helpers.h` was recent and not in every distro where Retis is built. I guess with #237 this will be resolved. So if I got this correctly, we can merge it once the cited PR is merged.
> 
> we already use vendored headers exposed by libbpf-sys, so we should be fine already. I would be surprised if we use system's

just to clarify, I guess this got solved once we upgraded libbpf-{rs,sys} at some point

atenart (approved):
Thanks for the explanations, LGTM!

PR #344, opened 2024-02-15
Fix `make install` and use git to set version

atenart:


Makefile
```diff
@@ -78,9 +78,6 @@ endif
 
 all: debug
 
-install: release
```
vlrpl, line 81: I think it's fine to keep a separate target.
Explicit `--offline --locked` is what we need to avoid it to be rebuilt (instead of relying on `$CARGO_INSTALL_OPTS`)

Makefile
```diff
@@ -4,8 +4,8 @@ export CLANG := clang
 export OBJCOPY := llvm-objcopy
 
 CARGO := cargo
-RELEASE_VERSION ?= $(shell $(CARGO) metadata --no-deps --format-version=1 | jq -r '.packages | .[] | select(.name == "retis") | .version')
-RELEASE_NAME ?= $(shell $(CARGO) metadata --no-deps --format-version=1 | jq '.packages | .[] | select(.name=="retis") | .metadata.misc.release_name')
+RELEASE_VERSION = $(shell $(CARGO) metadata --no-deps --format-version=1 | jq -r '.packages | .[] | select(.name == "retis") | .version')
```
vlrpl: any specific reason behind this?
The idea here was to allow to create quick scratch builds without editing any file

atenart: The idea is to make the version predictable so we can trust it (while using retis, in bug reports, in discussions with users, etc). Allowing to set it to a different value makes sense for a custom build but with this PR such builds will have a specific version that we can use to distinguish a proper build (and the commit id should be included which also can help later on to be sure exactly what was built).

IMO this does not prevent to have a custom version, and even improves the situation as we have a better idea what was the tree like when building it.

Do you see other use cases not covered after this PR?

Makefile
atenart: Is there any advantage to have this two stage build & install for the `install` target? I'm feeling we would special case it and I'm not 100% sure `cargo install` is exactly (and will continue to be) equivalent to `cargo build --release && cargo install`.

Well, on the other hand when packaging it for a distro the two stage logic is used :)

vlrpl: Follows the makefile logic. 
Also, `$CARGO_INSTALL_OPTS` was intentionally added/special-cased (part of the feedback) but now we're potentially mixing it with `$CARGO_CMD_OPTS`.
Considering the installation as a separate step (the options above allow to make it logically separate), I'm not sure it can be considered special casing. The only common thing is that `cargo` is used (which I guess is theoretically non-mandatory).
At some point install would need to install `-man`, `-completion` and so forth (of course, those will be different targets.
I'll leave it up to you, though.

Just to clarify, what I had in mind was something (unverified, simple enough to probably work):

```diff
diff --git a/Makefile b/Makefile
index 85feab7..72a37bc 100644
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,13 @@
 ROOT_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
-export LLC := llc
-export CLANG := clang
-export OBJCOPY := llvm-objcopy
+LLC := llc
+CLANG := clang
+OBJCOPY := llvm-objcopy
 
 CARGO := cargo
-RELEASE_VERSION ?= $(shell $(CARGO) metadata --no-deps --format-version=1 | jq -r '.packages | .[] | select(.name == "retis") | .version')
-RELEASE_NAME ?= $(shell $(CARGO) metadata --no-deps --format-version=1 | jq '.packages | .[] | select(.name=="retis") | .metadata.misc.release_name')
+RETIS_RELEASE_VERSION ?= $(shell $(CARGO) metadata --no-deps --format-version=1 | jq -r '.packages | .[] | select(.name == "retis") | .version')
+RETIS_RELEASE_NAME ?= $(shell $(CARGO) metadata --no-deps --format-version=1 | jq '.packages | .[] | select(.name=="retis") | .metadata.misc.release_name')
+
+export LLC CLANG OBJCOPY RETIS_RELEASE_NAME RETIS_RELEASE_VERSION
 
 PRINT = echo
 
@@ -79,13 +81,11 @@ endif
 all: debug
 
 install: release
-       $(CARGO) $(CARGO_OPTS) install $(CARGO_INSTALL_OPTS) --path=.
+       $(CARGO) $(CARGO_OPTS) install $(CARGO_INSTALL_OPTS) --path=. --offline --frozen
 
 define build
        $(call out_console,CARGO,$(strip $(2)) ...)
        $(Q)CARGO_BUILD_JOBS=$(CARGO_JOBS) \
-       RETIS_RELEASE_VERSION=$(RELEASE_VERSION) \
-       RETIS_RELEASE_NAME=$(RELEASE_NAME) \
        $(CARGO) $(CARGO_OPTS) $(1) $(CARGO_CMD_OPTS)
 endef
```

Also, we could consider, but it's really unrelated:

```
s/RETIS_RELEASE_NAME/RELEASE_NAME/
s/RETIS_RELEASE_VERSION/RELEASE_VERSION/
```

Makefile
vlrpl: no, the PR logic is fine and it is a very nice improvement.
What I was trying to suggest/keep, given you can always change those variables by editing the Makefile (when built from sources :) was the ability to specify a custom name/version (probably the name makes more sense). `?=` doesn't do anything other than easily override the behavior introduced by this PR (something like: "scratch-build-with-this-nice-feature").
This is because -dirty doesn't tell the whole story.

Maybe a "LOCALVERSION" suffix (instead of dirty, if exists) could further improve it? WDYT?

Makefile
atenart: > Follows the makefile logic. Also, `$CARGO_INSTALL_OPTS` was intentionally added/special-cased (part of the feedback) but now we're potentially mixing it with `$CARGO_CMD_OPTS`. 

Oops :(

> Considering the installation as a separate step (the options above allow to make it logically separate), I'm not sure it can be considered special casing. The only common thing is that `cargo` is used (which I guess is theoretically non-mandatory). At some point install would need to install `-man`, `-completion` and so forth (of course, those will be different targets.

Makes sense. We might even not use `cargo install` for some cases.

> Just to clarify, what I had in mind was something (unverified, simple enough to probably work):

Looks like there are many changes but I guess only the following is related:

```diff
 install: release
-       $(CARGO) $(CARGO_OPTS) install $(CARGO_INSTALL_OPTS) --path=.
+       $(CARGO) $(CARGO_OPTS) install $(CARGO_INSTALL_OPTS) --path=. --offline --frozen
```

Will do the above.

> Also, we could consider (unrelated):
> 
> ```
> s/RETIS_RELEASE_NAME/RELEASE_NAME/
> s/RETIS_RELEASE_VERSION/RELEASE_VERSION/
> ```

Right, it's not like we have different projects in the same repo :)

atenart: Maybe we could default `CARGO_INSTALL_OPTS` to `--path=.` too?

Makefile
atenart: > What I was trying to suggest/keep, given you can always change those variables by editing the Makefile (when built from sources :)

Right :) Let's say allowing to override is a built-in feature.

> was the ability to specify a custom name/version (probably the name makes more sense). `?=` doesn't do anything other than easily override the behavior introduced by this PR (something like: "scratch-build-with-this-nice-feature"). This is because -dirty doesn't tell the whole story.
> 
> Maybe (instead of the `?=` thing) a "LOCALVERSION" suffix (as a replacement for dirty, if both exist) could further improve it? WDYT?

Good point, makes sense to allow overriding the version name to have a human readable string. We can keep the version built-in though, to avoid confusion when overriding the version name.

With that, I would keep the overriding of the name built-in and not introduce a localversion to keep things simple (for a feature which is not used too much).

Does that work?

Makefile
vlrpl: > > Follows the makefile logic. Also, `$CARGO_INSTALL_OPTS` was intentionally added/special-cased (part of the feedback) but now we're potentially mixing it with `$CARGO_CMD_OPTS`.
> 
> Oops :(
> 
> > Considering the installation as a separate step (the options above allow to make it logically separate), I'm not sure it can be considered special casing. The only common thing is that `cargo` is used (which I guess is theoretically non-mandatory). At some point install would need to install `-man`, `-completion` and so forth (of course, those will be different targets.
> 
> Makes sense. We might even not use `cargo install` for some cases.
> 
> > Just to clarify, what I had in mind was something (unverified, simple enough to probably work):
> 
> Looks like there are many changes but I guess only the following is related:
> 
> ```diff
>  install: release
> -       $(CARGO) $(CARGO_OPTS) install $(CARGO_INSTALL_OPTS) --path=.
> +       $(CARGO) $(CARGO_OPTS) install $(CARGO_INSTALL_OPTS) --path=. --offline --frozen
> ```
> 
> Will do the above.
> 

That's what I would have said, but seems that cargo is smart enough to detect that the env is different and then rebuilds accordingly. So it seems RELEASE_{VERSION,NAME} must be set.

> > Also, we could consider (unrelated):
> > ```
> > s/RETIS_RELEASE_NAME/RELEASE_NAME/
> > s/RETIS_RELEASE_VERSION/RELEASE_VERSION/
> > ```
> 
> Right, it's not like we have different projects in the same repo :)

atenart: > That's what I would have said, but seems that cargo is smart enough to detect that the env is different and then rebuilds accordingly. So it seems RELEASE_{VERSION,NAME} must be set.

Ah, I see, missed that form the diff.

vlrpl: > Maybe we could default `CARGO_INSTALL_OPTS` to `--path=.` too?

sorry, I missed this. Sounds good to me. If you decide to do so, would it be fine for you to add a note about this default value in the help target?

atenart (commented):
Self-commenting: looks like the two stage build/install does not work as expected. Version information looks good but the binary is not right.

atenart:
> Self-commenting: looks like the two stage build/install does not work as expected. Version information looks good but the binary is not right.

Turns out I'm just tired; everything looks good...

vlrpl (approved):
lgtm. Thanks!

PR #358, opened 2024-02-29
build.rs: add error hint when build before making ebpf

liuhangbin:
Since #237 the BPF part is not handled by `cargo build` anymore. Users who are not aware of this and run `cargo build` directly will get error:
```
  --- stderr
  thread 'main' panicked at build.rs:46:9:
  failed to generate `src/core/probe/kernel/bpf/.out/kprobe.bpf.o`

  Caused by:
      0: Failed to generate skeleton for src/core/probe/kernel/bpf/.out/kprobe.bpf.o
      1: No such file or directory (os error 2)
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The users could not get what should do to avoid this error. Let's add an error hint so they can know how to avoid it. Here is what it looks like:
```
  --- stderr
  thread 'main' panicked at build.rs:48:17:
  Unable to find src/core/probe/kernel/bpf/.out/kprobe.skel.rs, please try 'make ebpf' first

  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,4 @@
-use std::{env, fs::File, io::Write, path::Path};
+use std::{env, fs::File, io, io::ErrorKind, io::Write, path::Path};
 
 use libbpf_cargo::SkeletonBuilder;
 use memmap2::Mmap;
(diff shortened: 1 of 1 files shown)


atenart (commented):
Don't we need the same improved error message in `gen_hook_skel`?

liuhangbin:
> Don't we need the same improved error message in `gen_hook_skel`?

After `make ebpf`, `cargo build` will pass. So I think there is no need to add another hit.

atenart:
> > Don't we need the same improved error message in `gen_hook_skel`?
> 
> After `make ebpf`, `cargo build` will pass. So I think there is no need to add another hit.

That's right. But we might change the order of the calls in `build.rs`; or someone could manually remove one of the BPF object files, etc. I agree those are not likely events, but having the right error message from the beginning will help maintenance.

atenart (approved):
LGTM, thanks!

Issue #317, opened 2023-12-20
meta filter: advanced filters


vlrpl:
Consider extending the syntax to something like:

`-m sk_buff._nfct:{aritm,mask}_expr:resulting_type[.field] == <value>`

This would allow the user to follow non-trivial pointers embedding other info:

`-m sk_buff._nfct:&~0x7:nf_conn.mark == ...`

This should also optionally make it possible to match fields in the following way:

`-m sk_buff._nfct:&0x7 == 0 // IP_CT_ESTABLISHED`

vlrpl:
Closed with #392

vlrpl:
Closed with #392

PR #362, opened 2024-03-04
remove bound check in meta filter

vlrpl:
A clean-up chore removing unneeded check.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -235,6 +235,9 @@ The comparison operators are:
 4. ">" and ">=" for *greater than* and *greater than or equal to*
 
 At the moment, only number and string comparisons are supported.
+The right-hand side (rhs) of numeric matches must be expressed as
+literal and can be represented in either base 10 or base 16, with the
+latter starting with `0x` prefix.
 All the comparison operators support numbers (both signed and unsigned).
 For strings only the operator *equal to* is supported, furthermore, the
 string must be enclosed between *quotes*.
--- a/src/core/filters/meta/bpf/include/meta_filter.h
+++ b/src/core/filters/meta/bpf/include/meta_filter.h
@@ -139,7 +139,7 @@ static __always_inline
 bool filter_bytes(struct retis_meta_ctx *ctx)
 {
 	char val[META_TARGET_MAX];
-	u32 sz;
+	long sz;
 
 	/* if it is an array of chars use its size. Alternatively, use
 	 * the target size (probe read could fail in this case).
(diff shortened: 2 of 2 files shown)


docs/index.md
```diff
@@ -223,7 +223,7 @@ Meta filtering also automatically follows struct pointers, so indirect access to
 structures pointed by an `sk_buff` field is possible.
 
 ```none
-$ retis collect -m 'sk_buff.dev.nd_net.net.ns.inum == 4026531840'
+$ retis collect -m 'sk_buff.dev.nd_net.net.ns.inum == 0xF0000000'
```
atenart, line 226: Not sure about this one. While showing hex can be used is OK, netns ids are rarely used in the hex form IMO.

vlrpl: IDK, it's an inode number between:
`PROC_DYNAMIC_FIRST and 0xffffffff`

Multiple places refer to it in hex:
https://github.com/torvalds/linux/blob/90d35da658da8cff0d4ecbb5113f5fac9d00eb72/fs/nfs/callback.c#L176

Not sure this is bad. At least the upside is that shows how hexes are required in exprs.

vlrpl: anyways, no strong opinions about this. We can always add a separate line for it.

atenart: It's just that I always saw netns being referenced as a decimal value (eg. `lsns`) and as a user I find using an hexadecimal value surprising; but that is only my own experience, I could be wrong :)

vlrpl: > It's just that I always saw netns being referenced as a decimal value (eg. `lsns`) and as a user I find using an hexadecimal value surprising; but that is only my own experience, I could be wrong :)

It's mostly the reason I didn't bother converting it in the first place:

`ls -l /proc/$$/ns/` 

but I changed it here (instead of simply specifying that number can be hex or dec) because seemed a convenient way to show it is also possible to use that (after checking is not uncommon :)

Just to be clear, I'm ok to keep it decimal. I'll simply add a note separately.

atenart: > Just to be clear, I'm ok to keep it decimal. I'll simply add a note separately.

Looks good to me, I think the doc explicitly mentioning hex is good enough. But if you want, we can add another example with a more natural hex value. One that comes to mind would be `skb->hash` for example.

This made me realize we probably would like to support the `&` operator, as `skb->dev->features & 0xf00` would have been a good hex example. That is of course unrelated to this (and you probably already have it on your todo list) :)

vlrpl: That's certainly a nice improvement and should be fairly straightforward to add.

vlrpl: in the todo list there's something [involving '&'](https://github.com/retis-org/retis/issues/317) but not exactly fitting. I'll add this.

vlrpl: > > Just to be clear, I'm ok to keep it decimal. I'll simply add a note separately.
> 
> Looks good to me, I think the doc explicitly mentioning hex is good enough. 
> [...].
> 

I simply rephrased it to make it clearer.

atenart (approved):
LGTM, thanks!

PR #365, opened 2024-03-06
collect: support collect sub command auto complete

liuhangbin:
When generating the auto-complete file, we need to get full command parameters after `get_cli()?.build()`. But the collect subcommand only has a default collector setting at the first build round. It only generates all command parameters after running the subcommand specifically.

To generate all the parameters in the first round, we need to update the SubCommand full implementation to register all modules during building. With this, we can auto-complete all collect sub-parameters.

After update, there is no need to register again in command run.

At the same time, since we registered all modules during command build. In testing, if the collector only has a part of the modules, the init will fail with error "unknown collector {}". There is no need to initialize all collectors from args. We only need to initialize the collectors from the current modules.

---

Here is the output:

```Bash
# retis collect -<Tab>
-p                      -c                      --probe                 --filter-packet         --skb-sections          --help
-o                      -h                      --out                   --filter-meta           --nft-verdicts          
-f                      --stack                 --print                 --probe-stack           --ovs-track             
-m                      --cmd                   --format                --allow-system-changes  --collectors            
# retis collect --skb-sections <Tab>
all      eth      arp      ip       tcp      udp      icmp     dev      ns       meta     dataref  packet   
# retis collect --nft-verdicts <Tab>
all       continue  break     jump      goto      return    drop      accept    stolen    queue     repeat    
# retis collect --collectors <Tab>
skb-drop      skb           ct            nft           skb-tracking  ovs           
```

### DIFF
--- a/src/collect/cli.rs
+++ b/src/collect/cli.rs
@@ -178,7 +178,7 @@ impl SubCommand for Collect {
         Some(&mut self.collectors)
     }
 
-    fn full(&self) -> Result<Command> {
+    fn full(&mut self) -> Result<Command> {
         let long_about = "Collect events using 'collectors'.\n\n \
             Collectors are modules that extract \
             events from different places of the kernel or userspace daemons \
@@ -187,8 +187,13 @@ impl SubCommand for Collect {
 
         // Determine all registerd collectors and specify both the possible values and the default
         // value of the "collectors" argument
-        let possible_collectors =
-            Vec::from_iter(self.collectors.modules().iter().map(|x| x.to_str()));
+        let mut modules = crate::get_modules()?;
+        let collectors = modules.collectors();
+        collectors
+            .values()
+            .try_for_each(|c| c.register_cli(&mut self.collectors))?;
+        let possible_collectors: Vec<&'static str> =
+            collectors.keys().map(|m| m.to_str()).collect();
 
         let full_command = self
             .collectors
--- a/src/collect/collector.rs
+++ b/src/collect/collector.rs
@@ -114,20 +114,6 @@ impl Collectors {
         })
     }
 
-    // Register the dynamic commands with the cli and parse collector-specific arguments
-    fn register_cli(&mut self, mut cli: FullCli) -> Result<CliConfig> {
-        // Register all collectors' command line arguments. Cli registration
-        // errors are fatal.
-        let cmd = cli.get_subcommand_mut()?.dynamic_mut().unwrap();
-        self.modules
-            .collectors()
-            .iter()
-            .try_for_each(|(_, c)| c.register_cli(cmd))?;
-
-        // Now we can parse all parameters.
-        Ok(cli.run()?)
-    }
-
     /// Setup user defined input filter.
     fn setup_filters(probes: &mut ProbeBuilderManager, collect: &Collect) -> Result<()> {
         if let Some(f) = &collect.args()?.packet_filter {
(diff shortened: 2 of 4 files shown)


atenart (changes requested):

src/collect/collector.rs
```diff
@@ -115,6 +115,7 @@ impl Collectors {
     }
 
     // Register the dynamic commands with the cli and parse collector-specific arguments
+    #[allow(dead_code)]
     fn register_cli(&mut self, mut cli: FullCli) -> Result<CliConfig> {
```
atenart, line 119: Is this used anywhere? IMO looks like the function could be removed instead.

liuhangbin: > Is this used anywhere now? IMO looks like the function could be removed instead.

Yes, we still need it. e.g.register the collectors in full()
```rust
        let collectors = modules.collectors();
        let _ = collectors
            .values()
            .try_for_each(|c| c.register_cli(&mut self.collectors));
```
Or modules registers, e.g.

```rust
impl Collector for OvsModule {

    fn register_cli(&self, cmd: &mut DynamicCommand) -> Result<()> {
        cmd.register_module::<OvsCollectorArgs>(ModuleId::Ovs)
    }

```

atenart: But the above example is the `Collector` trait implementation, for actual collectors. The function I'm suggesting to remove is an implementation of the `Collectors` command, not a trait implementation. Things are a little bit confusing here, I admit :)

liuhangbin: Ah, got it. I will update the patch.

src/collect/cli.rs
```diff
@@ -187,8 +187,13 @@ impl SubCommand for Collect {
 
         // Determine all registerd collectors and specify both the possible values and the default
         // value of the "collectors" argument
-        let possible_collectors =
-            Vec::from_iter(self.collectors.modules().iter().map(|x| x.to_str()));
+        let mut modules = crate::get_modules()?;
+        let collectors = modules.collectors();
+        let _ = collectors
+            .values()
+            .try_for_each(|c| c.register_cli(&mut self.collectors));
```
atenart, line 194: I'm sorry, was about to merge this but only see this now. We should not silence the error here, please instead do:

```rust
collectors
            .values()
            .try_for_each(|c| c.register_cli(&mut self.collectors))?;
```

liuhangbin: Hmm, got error after update
```bash
# ./target/debug/retis collect -h
thread 'main' panicked at src/cli/cli.rs:351:80:
called `Result::unwrap()` on an `Err` value: module SkbDrop already registered
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Need to figure out how to avoid the module already registered issue.

atenart: Ah right, we have this two step registration for the cli. Since we don't do this anymore I'm wondering if this is still needed (to do the second one). The following removes the second registration (untested):
```diff
diff --git a/src/cli/cli.rs b/src/cli/cli.rs
index 1d5ebd50c372..f2c76bd98a36 100644
--- a/src/cli/cli.rs
+++ b/src/cli/cli.rs
@@ -345,11 +345,6 @@ impl FullCli {
                 .join(" ")
         );
 
-        // Replace the ran subcommand with the full subcommand.
-        self.command = self
-            .command
-            .mut_subcommand(self.subcommand.name(), |_| self.subcommand.full().unwrap());
-
         // Get the matches.
         let matches = match cfg!(test) {
             true => self.command.try_get_matches_from_mut(self.args.iter())?,
```

liuhangbin: Thanks! I just tested it and works!

atenart (approved):
LGTM, thanks!

PR #370, opened 2024-03-13
Makefile clean up and improvements

vlrpl:
- Introduces some rough ARCH logic (instead of the previous always hard-code value detached from the Rust build)
- Addresses the requests not covered during the first introduction of the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,33 +46,17 @@ ifeq ($(NOVENDOR),)
     LIBBPF_INCLUDES := $(ROOT_DIR)/src/.out
 endif
 
-FILTER_INCLUDES := src/core/filters/packets/bpf/include \
-                   src/core/filters/meta/bpf/include
 # Taking errno.h from libc instead of linux headers.
 # TODO: Remove when we fix proper header dependencies.
-INCLUDES_ALL := $(abspath $(wildcard src/core/probe/bpf/include \
-                                     src/core/probe/kernel/bpf/include \
-                                     src/core/probe/user/bpf/include \
-                                     src/core/events/bpf/include \
-                                     src/core/tracking/bpf/include \
-                                     /usr/include/x86_64-linux-gnu \
-                                     $(FILTER_INCLUDES)))
+INCLUDES_ALL := $(abspath $(wildcard $(shell find src/core -type d -path '*/bpf/include') \
+                                     /usr/include/x86_64-linux-gnu))
 INCLUDES_ALL += $(LIBBPF_INCLUDES)
 
-OVS_INCLUDES := $(abspath src/module/ovs/bpf/include)
 INCLUDES := $(addprefix -I, $(INCLUDES_ALL))
 
-EBPF_PROBES := $(abspath src/core/probe/kernel/bpf \
-                        src/core/probe/user/bpf)
+EBPF_PROBES := $(abspath $(wildcard src/core/probe/*/bpf))
 
-GENERIC_HOOKS := $(abspath src/module/skb/bpf \
-                           src/module/skb_drop/bpf \
-                           src/module/skb_tracking/bpf \
-                           src/module/nft/bpf \
-                           src/module/ct/bpf)
-
-OVS_HOOKS := $(abspath src/module/ovs/bpf)
-OUT_NAME := HOOK
+EBPF_HOOKS := $(abspath $(wildcard src/module/*/bpf))
 
 all: debug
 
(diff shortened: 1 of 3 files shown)


ebpf.mk
```diff
@@ -1,6 +1,14 @@
...
 # specific facility.
-ARCH := x86
+DEFAULT_ARCH := $(patsubst target_arch="%",%,$(filter target_arch="%",$(shell rustc --print cfg)))
+__ARCH := $(if $(CARGO_CFG_TARGET_ARCH),$(CARGO_CFG_TARGET_ARCH),$(DEFAULT_ARCH))
+x86_64 := x86
+aarch64 := arm64
+powerpc64 := powerpc
+s390x := s390
+# Mappings takes precedence over custom ARCH
+ARCH := $(if $($(__ARCH)),$($(__ARCH)),$(__ARCH))
```
atenart, line 10: If that's possible I would invert the logic here, to set `CARGO_CFG_TARGET_ARCH` using `ARCH`. This has two advantages:
- `ARCH` is commonly used in various projects for this.
- We would not rely on a Cargo specific variable and instead both targets (Rust & eBPF) would be abstracted.

Then we could run `ARCH=aarch64 make`.

Makefile
```diff
@@ -101,20 +89,20 @@ $(LIBBPF_INCLUDES): $(LIBBPF_SYS_LIBBPF_INCLUDES)
 	cp $^ $(LIBBPF_INCLUDES)/bpf/
 endif
 
-$(OVS_HOOKS): INCLUDES_EXTRA := -I$(OVS_INCLUDES)
+$(OVS_HOOK): INCLUDES_EXTRA := -I$(OVS_INCLUDES)
```
atenart, line 92: Instead of special-casing OvS, could we automatically add `-Isrc/module/<name>/bpf/include` (if path exists) when building each module?

build.rs
```diff
@@ -62,6 +62,35 @@ fn gen_probe_skel(source: &str) {
     println!("cargo:rerun-if-changed={source}");
 }
 
+// This establish a naming convention. eBPF objs MUST be suffixed with
```
atenart, line 65: s/establish/establishes/

atenart (changes requested):
Thanks for working on this!

ebpf.mk
vlrpl: I thought about it but I ended up with this because CARGO one makes it clear (they are documented) what are the arch names (then mapped to "ebpf" ones). 
 
If we go with `ARCH` the names must be documented (help, at least), but we claim (for good reasons) that no support is available yet, so it should not be documented at the same time (essentially we have to do it later).

Just sharing the reasoning that brought to this. I'm fine with the change.

atenart: > If we go with `ARCH` the names must be documented (help, at least), but we claim (for good reasons) that no support is available yet, so it should not be documented at the same time (essentially we have to do it later).

Yep, IMO that's fine. We'll have to document which ones are supported, so we'll have to list the right values anyway. If an arch is not supported and someone wants to build it, they'll have to look up what's needed on Cargo and BPF side anyway.

Makefile
```diff
@@ -83,6 +70,10 @@ define build
 	$(CARGO) $(CARGO_OPTS) $(1) $(CARGO_CMD_OPTS)
 endef
 
+define build_ebpf
+
+endef
```
atenart, line 75: What does this do?

build.rs
```diff
@@ -78,31 +104,13 @@ fn gen_bindings() {
...
-    gen_hook_skel("src/module/ovs/bpf/.out/kernel_upcall_ret.bpf.o");
-    gen_hook_skel("src/module/ovs/bpf/.out/user_recv_upcall.bpf.o");
-    gen_hook_skel("src/module/ovs/bpf/.out/user_op_exec.bpf.o");
-    gen_hook_skel("src/module/ovs/bpf/.out/user_op_put.bpf.o");
-    gen_hook_skel("src/module/nft/bpf/.out/nft.bpf.o");
-    gen_hook_skel("src/module/ct/bpf/.out/ct.bpf.o");
-
-    println!("cargo:rerun-if-changed={BINDGEN_HEADER}");
+    walk_gen_skels("src/core/probe/", &gen_probe_skel);
+    walk_gen_skels("src/module/", &gen_hook_skel);
```
atenart, line 115: Not a request for change, just that I like this change and using a callback! :)

Makefile
vlrpl: really?!? :)
Sorry, it probably was too late yesterday and I forgot to refresh this hunk removal.

atenart (approved):
LGTM, thanks!

Issue #373, opened 2024-03-25
retis on aarch64


vlrpl:
It would be nice to have Retis (at least) [cross-]compile on arm.

What's needed:
1. split the single `vmlinux.h` logic into `vmlinux-x86_64.h` and `vmlinux-aarch64.h` (for arch-dependent data access, e.g. `pt_regs`). Essentially, this requires to generate a new vmlinux.h for arm64
2. `vmlinux.h` that pulls the arch-specific `vmlinux*.h` depending on `__TARGET_ARCH_*`
3. [optional] This could be a good chance to reshape our vmlinux.h (unnecessarily bloated)
5. [optional] Ideally, CI with functional tests

atenart:
Done in #379.

PR #374, opened 2024-03-27
decouple arch from the build

vlrpl:
this is a preparation to complete  #373 
Only an arm `vmlinux.h` will be left for implementing the non-optional bullets.


atenart (dismissed):
One suggestion: using `ARCH/vmlinux.h` instead of `vmlinux-ARCH.h`. But not a blocker, so I'm approving this and you can either merge it or make the change :)

vlrpl:
> One suggestion: using `ARCH/vmlinux.h` instead of `vmlinux-ARCH.h`. But not a blocker, so I'm approving this and you can either merge it or make the change :)

I decided to go with `-arch` suffix because I don't expect other arch-specific files, so there was no strong opinion. 
Anyways, given there's a preference, I renamed the file and put it under `arc/` dir.

.gitattributes
```diff
@@ -1,2 +1,2 @@
-vmlinux.h linguist-vendored
+vmlinux*.h linguist-vendored
```
atenart, line 1: We probably need to update this as well.

vlrpl: right. Should work :) but it is unneeded now. Thanks

atenart (approved):
LGTM, thanks!

PR #375, opened 2024-04-02
Fix OvS runtime CI failure

atenart:
Haven't went deep into the why here, but this fixes our runtime CI.


vlrpl (approved):
Thanks @atenart, the change looks good to me

PR #376, opened 2024-04-03
Meta bitfields

vlrpl:
Adds bitfield support for LE ARCH

### DIFF
--- a/src/core/filters/meta/bpf/include/meta_filter.h
+++ b/src/core/filters/meta/bpf/include/meta_filter.h
@@ -1,6 +1,8 @@
 #ifndef __CORE_FILTERS_META_FILTER__
 #define __CORE_FILTERS_META_FILTER__
 
+#include <common_defs.h>
+
 /* Please keep in sync with its Rust counterpart. */
 #define META_OPS_MAX	32
 #define META_TARGET_MAX	32
@@ -26,6 +28,7 @@ union retis_meta_op {
 		u8 type;
 		u8 nmemb;
 		u16 offt;
+		u8 bf_size;
 	} l;
 	struct {
 		union {
@@ -54,6 +57,8 @@ struct retis_meta_ctx {
 	u8 type;
 	/* optional */
 	u8 nmemb;
+	/* optional bitfield size. */
+	u8 bfs;
 	/* Target Info */
 	/* actual data. */
 	void *data;
@@ -111,6 +116,7 @@ static __always_inline long meta_process_ops(struct retis_meta_ctx *ctx)
 		ctx->offset = val->l.offt;
 		ctx->type = val->l.type;
 		ctx->nmemb = val->l.nmemb;
+		ctx->bfs = val->l.bf_size;
 	}
 
 	return 0;
@@ -176,7 +182,23 @@ bool filter_bytes(struct retis_meta_ctx *ctx)
 	return !(*sp1 - *sp2);
 }
 
-static __always_inline u64 fixup_signed(u64 val, u32 sz)
+/* Removes prefix and suffix bits that can be part of a potential
+ * previous or following bitfield and extracts the bitfield into a
+ * u64.
+ * Right shift on a signed negative numbers is implementation-defined.
+ * Both clang and gcc act on them by sign extension.
+ * This allows to preseve sign while extracting the bitfield.
+ */
+static __always_inline
+u64 extract_bf(u64 val, bool has_sign, u16 bit_off, u16 bit_sz)
+{
+	val <<= (64 - bit_sz) - (bit_off  % 8);
+
+	return has_sign ? (s64)val >> (64 - bit_sz) : val >> (64 - bit_sz);
+}
+
+static __always_inline
+u64 fixup_signed(u64 val, u32 sz)
 {
 	u64 ret;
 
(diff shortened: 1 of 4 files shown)


atenart (commented):
A few high level comments:
- While it's not 100% related, I think allowing operands like `&` would be very nice when working with bitfields, as that would be the main use IMO. We should also find a way to test a bit is not set.
- First patch also includes unrelated changes; those could be split.
- Documentation should be updated too.
- I think some comments could be added here and there to help following the code.

src/core/filters/meta/bpf/include/meta_filter.h
```diff
@@ -190,23 +221,57 @@ static __always_inline u64 fixup_signed(u64 val, u32 sz)
...
 }
 
+/* Returns the size in bytes that fits the
+ * bitfield size also considering the offset.
+ */
+int get_bfs_byte_sz(u16 bio, u16 byo, u8 bit_sz)
+{
+	u16 bio_rel = bio - byo * 8;
+
+	if (bio_rel < 0) {
```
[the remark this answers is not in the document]
vlrpl: yep, this was changed last minute before pushing the draft to make clang front end happy and I didn't update the branch in the draft (should be the equivalent of if (byo * 8 > bio)).

vlrpl:
> A few high level comments:
> 
>     * While it's not 100% related, I think allowing operands like `&` would be very nice when working with bitfields, as that would be the main use IMO. We should also find a way to test a bit is not set.
> 

I think it's fairly easy to add simple '&' support (for unsigned only I'd say). 
The idea was to tackle this part with #317, but it will not make it for this release (could require parsing rework).
Not sure we might want a separate operator. I'll take a look.
At the moment, I open this PR as draft for 2 reasons: it's a draft (no surprise here :) , we should decide if we want to pick this for the next release.

>     * First patch also includes unrelated changes; those could be split.
> 
>     * Documentation should be updated too.
> 

both the above make sense

>     * I think some comments could be added here and there to help following the code.

will add some notes

atenart:
> > A few high level comments:
> > ```
> > * While it's not 100% related, I think allowing operands like `&` would be very nice when working with bitfields, as that would be the main use IMO. We should also find a way to test a bit is not set.
> > ```
> 
> I think it's fairly easy to add simple '&' support (for unsigned only I'd say). The idea was to tackle this part with #317, but it will not make it for this release (could require parsing rework). Not sure we might want a separate operator. I'll take a look. At the moment, I open this PR as draft for 2 reasons: it's a draft (no surprise here :) , we should decide if we want to pick this for the next release.

If we are planning to add support for `&` soon (ie. next release) I think we can add this for 1.4 as it'll still provide basic bitfield support.

vlrpl:
still just a draft. Will update to a formal version soon

vlrpl:
> > > A few high level comments:
> > > ```
> > > * While it's not 100% related, I think allowing operands like `&` would be very nice when working with bitfields, as that would be the main use IMO. We should also find a way to test a bit is not set.
> > > ```
> > 
> > 
> > I think it's fairly easy to add simple '&' support (for unsigned only I'd say). The idea was to tackle this part with #317, but it will not make it for this release (could require parsing rework). Not sure we might want a separate operator. I'll take a look. At the moment, I open this PR as draft for 2 reasons: it's a draft (no surprise here :) , we should decide if we want to pick this for the next release.
> 
> If we are planning to add support for `&` soon (ie. next release) I think we can add this for 1.4 as it'll still provide basic bitfield support.

yes, the plan is to introduce `&` plus other features targeting 1.5

src/core/filters/meta/bpf/include/meta_filter.h
```diff
@@ -168,7 +185,23 @@ bool filter_bytes(struct retis_meta_ctx *ctx)
...
+ * previous or following bitfield and extracts the bitfield into a
+ * u64.
+ * Right shift on a signed negative numbers is implementation-defined.
+ * Both clang and gcc act on them by sign extension.
+ * This allows to preseve sign while extracting the bitfield.
+ */
+static __always_inline
+u64 fixup_bf(u64 val, bool has_sign, u16 bit_off, u16 bit_sz)
+{
+	val <<= (64 - bit_sz) - (bit_off  % 8);
```
atenart, line 198: I don't get why the `%` above. Shouldn't we just remove the offset in bits?

Nit: would have name the function `extract_bf` for more clarity, but of course not a blocker at all, only personal taste :)

src/core/probe/bpf/include/common_defs.h
```diff
@@ -110,4 +110,6 @@ static __always_inline void err_report(u64 sym_addr, u32 pid)
 #define unlikely(x) __builtin_expect(!!(x), 0)
 #endif
 
+#define DIV_CEIL(m, n) (((m) + (n) - 1) / (n))
```
atenart, line 113: I think `m + n` could overflow here; instead `1 + ((m) - 1) / (n)` can be used. Not that it would happen here but if that is reused in other code paths.

src/core/filters/meta/bpf/include/meta_filter.h
```diff
@@ -103,6 +108,9 @@ static __always_inline long meta_process_ops(struct retis_meta_ctx *ctx)
 			    !ptr)
 				return -1;
 
+			if ((void *)ptr == NULL)
+				return 0;
```
atenart, line 112: Nit: I think this could be in a dedicated commit with an explanation about what this fixes.

src/core/filters/meta/bpf/include/meta_filter.h
vlrpl: we discussed offline the first part.
The new name sgtm.

src/core/filters/meta/bpf/include/meta_filter.h
vlrpl: ah, right. Forgot to split this.

vlrpl: looking at this twice, this is a leftover and should be dropped.
It was intended to be a replacement/slightly different approach for the previous `|| !ptr)`  (part of "bpf logger" and written/updated in parallel with this PR). 
This small redundant check fell through the cracks once eBPF logger got merged. Thanks for spotting this, though.

atenart (approved):
Nice improvement, thanks!

PR #377, opened 2024-04-04
Further skb module improvements

### LINKED ISSUE #244: ICMPv6 support
Exact implementation will depend on how #224 goes.

atenart:
Based on top of #338. Adds various improvements, including:
- Raw packet as base64 in the JSON event.
- ICMPv6 basic support.
- Improvement of early Tx capture.
- Bugs fixes.

Fixes #244.


src/module/skb/skb.rs
```diff
@@ -53,11 +54,10 @@ impl Collector for SkbModule {
         // First, get the cli parameters.
         let args = cli.get_section::<SkbCollectorArgs>(ModuleId::Skb)?;
 
-        let mut sections: u64 = 0;
+        let mut sections: u64 = 1 << SECTION_PACKET;
```
vlrpl, line 77: does this worth a note/update in the doc?
WDYT?

src/module/skb/skb.rs
```diff
@@ -54,23 +58,29 @@ impl Collector for SkbModule {
...
-                "icmp" => sections |= 1 << SECTION_ICMP,
                 "dev" => sections |= 1 << SECTION_DEV,
                 "ns" => sections |= 1 << SECTION_NS,
                 "meta" => sections |= 1 << SECTION_META,
                 "dataref" => sections |= 1 << SECTION_DATA_REF,
                 "gso" => sections |= 1 << SECTION_GSO,
-                "packet" => {
-                    warn!("Use of 'packet' in --skb-sections is depreacted (is now always set)")
+                "packet" | "eth" | "arp" | "ip" | "tcp" | "udp" | "icmp" => {
+                    warn!(
```
vlrpl, line 92: does this worth (maybe along with packet, if makes sense) a note/update in the doc?
WDYT?

atenart: Absolutely!

atenart:
Took the doc comment into account and rebased on top of main.

src/module/skb/bpf/skb_hook.bpf.c
```diff
@@ -80,6 +80,54 @@ struct skb_packet_event {
...
+	if (likely(protocol))
+		return protocol;
+
+	/* We're *most likely* in the Tx path; skb->protocol wasn't set yet.
+	 * Let's try to detect the protocol from the packet data.
+	 */
+
+	head = BPF_CORE_READ(skb, head);
+
+	/* L4 must be set as we derive L3 header len from it */
```
amorenoz, line 109: ni: full stop at the end.

src/module/skb/bpf/skb_hook.bpf.c
```diff
@@ -125,7 +173,7 @@ static __always_inline int process_packet(struct retis_raw_event *event,
 	 * the eth header.
 	 */
 	} else if (is_network_data_valid(skb)) {
-		u16 etype = BPF_CORE_READ(skb, protocol);
+		u16 etype = skb_protocol(skb);
```
amorenoz, line 176: While looking at this I thought: would it be desirable to flag the eth header info as "fake" in this case so we can represent it as ???? -> ??? instead of all zeros? for someone not knowing retis internals it might be confusing to see those headers and people might thing there is something wrong with the e.g: driver.

amorenoz (approved):
some nits and a non-related topic. otherwise, LGTM

src/module/skb/bpf/skb_hook.bpf.c
atenart: Makes sense. Instead of reporting `???` we can just avoid filling the eth information in the event. I'll add a commit to this PR.

vlrpl (approved):
change lgtm, thanks!

PR #380, opened 2024-04-11
packet: rework and clean-up call placeholder handling

vlrpl:
Coalesce L2 and L3 functions as they essentially share most of the code.


atenart (approved):
LGTM, thanks! Always happy to nice nice clean-ups :)

PR #381, opened 2024-04-17
Some authors file (logical) updates

atenart:
This makes first contributions easier for people. Also removes the burden from the contributors and makes us update the authors file (using a GH action).

Note that for this to work we must enable `Settings > Actions > General > Workflow permissions > Read and write permissions`. IMO that's OK, but feel free to disagree.

For an example of this in action, see https://github.com/atenart/test. Generated commits look like:

```
Author: github-actions <41898282+github-actions[bot]@users.noreply.github.com>
Date:   Wed Apr 17 12:10:25 2024 +0000

    Update authors file
```

### DIFF
--- a/tools/authors.sh
+++ b/tools/authors.sh
@@ -20,7 +20,7 @@ done
 shift $(($OPTIND - 1))
 
 authors_file=$(git rev-parse --show-toplevel)/AUTHORS.md
-authors_list="$(git log --no-merges --format='- %aN' | grep -v dependabot | sort -u)"
+authors_list="$(git log --no-merges --format='- %aN' | grep -v dependabot | sort -fu)"
 
 out=$(cat <<-EOM
 # Authors
--- a/tools/authors.sh
+++ b/tools/authors.sh
@@ -7,13 +7,13 @@ help() {
 	echo "  Generate an alphabetically ordered list of authors and update the authors file"
 	echo ""
 	echo "  Options:"
-	echo "    -c		Checks the authors file is valid without modifying it."
+	echo "    -n		Dry run (show what would be modified)."
 	echo "    -h		Show this help."
 }
 
-while getopts "ch" opt; do
+while getopts "hn" opt; do
 	case $opt in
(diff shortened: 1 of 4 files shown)


atenart:
Had some discussions offline and letting a bot making automatic changes for a low volume activity was deemed too much. Instead we just make sure to update the authors file when releasing new versions.

tools/authors.sh
```diff
@@ -7,13 +7,11 @@ help() {
 	echo "  Generate an alphabetically ordered list of authors and update the authors file"
 	echo ""
 	echo "  Options:"
-	echo "    -c		Checks the authors file is valid without modifying it."
```
vlrpl, line 10: @atenart, what about renaming this to -n (or any of your preference) and keeping it as a dry-run, instead?

atenart: I don't think we should keep it as dry run, as it's not only checking the different steps can be run but is also checking the end result. If you think we should keep this, we can keep it as is IMO.

vlrpl: > I don't think we should keep it as dry run, as it's not only checking the different steps can be run but is also checking the end result. If you think we should keep this, we can keep it as is IMO.

I mean something like this (adding  `-u` to `diff`):
```diff
❯ ./tools/authors.sh -c
--- /dev/fd/63  2024-04-19 16:08:45.856496295 +0200
+++ /dev/fd/62  2024-04-19 16:08:45.856496295 +0200
@@ -3,4 +3,8 @@
 Many thanks to everyone who contributed to Retis!
 
 In alphabetical order:
+- Adrian Moreno
+- Antoine Tenart
+- Daniel Mendes
 - Hangbin Liu
+- Paolo Valerio
```

atenart: `diff -u` is quite nice, thanks for the suggestion!

> about dry run or check, I'm fine with both. Up to you.

On a second thought, you're 100% right, this is a dry run. Let's go with that.

vlrpl (approved):
lgtm, thanks!

PR #383, opened 2024-04-24
Containerfile: update build command

atenart:
Take into account our build system modifications and make sure no .out directory was copied over from an earlier build.


vlrpl (approved):
approved, thanks!

Issue #347, opened 2024-02-19
Remove use of `path-clean`


atenart:
Remove the `path-clean` dependency in Retis and its use.

ffmancera:
I guess this can be closed now @atenart, thanks!

atenart:
Fixed in #384.

PR #384, opened 2024-04-26
process: cli: drop path-clean dependency

ffmancera:
Path-clean is a low-activity simple library that can be imported in our code directly to avoid adding it as a dependency. As we want to package Retis in Fedora we need to reduce the number of dependencies to the minimum possible, otherwise we would need to own and maintain them too in Fedora.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1009,12 +1009,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "path-clean"
-version = "1.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "17359afc20d7ab31fdb42bb844c8b3bb1dabd7dcf7e68428492da7f16966fcef"
-
-[[package]]
 name = "pcap"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1286,7 +1280,6 @@ dependencies = [
  "nix 0.28.0",
  "once_cell",
  "pager",
- "path-clean",
  "pcap",
  "pcap-file",
  "plain",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,6 @@ memoffset = "0.9"
 nix = { version = "0.28", features = ["feature", "time", "user"] }
 once_cell = "1.15"
 pager = "0.16"
-path-clean = "1.0"
 pcap = "2.0"
 pcap-file = "2.0"
 plain = "0.2"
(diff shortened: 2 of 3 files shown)


ffmancera:
Fixes: https://github.com/retis-org/retis/issues/347

ffmancera:
> I only had minor comments on the changes, but then thought about the real reason we had `path-clean`. When using `retis sort` if the input file is the same as the output file, we could lose data. `path-clean` was introduced to fix this when the output file does not exist yet ([696c780](https://github.com/retis-org/retis/commit/696c78043b2ced75a443b978e40eb91f839e0ba8)).
> 
> However now looking at the `path-clean` implementation and at the later use of the `PathBuf` `Eq` implementation this does not work for all cases. Firstly, `path-clean` is not equivalent to `canonicalize` and secondly the `Eq` implementation of `PathClean` can return false for the same file depending on how the path is expressed. It turns out my initial attempt is bugged and the current logic does not ensure the input and output files are the same in all cases.
> 

Hm that is right. I did not think about it before.

> I think instead of all this we should remove all the `path-clean` + canonicalization code, and instead use `OpenOptions::create_new` early on when opening/creating the output file. This would really ensure we won't override the input file. (Of course, this would need to be tested).
> 

That makes sense indeed. Let me do that. Thanks!

> I hope the above is clear :)
> 
> Sorry if I misled you when we quickly chatted about this offline.

No worries at all, I should have noticed that as well but didn't. Let me work on this, I think the `OpenOptions` approach is great!

ffmancera:
Hm. Maybe I am being naive here, but I think for the current use case we have, `canonicalize()` is enough. True, `canonicalize()` will fail if the path doesn't exist but if it doesn't exist it means it is not the input one as the input file must be something already exist. So no need for path-clean and we can keep the old logic probably..

Am I missing something? @atenart what do you think?

atenart:
> Hm. Maybe I am being naive here, but I think for the current use case we have, `canonicalize()` is enough. True, `canonicalize()` will fail if the path doesn't exist but if it doesn't exist it means it is not the input one as the input file must be something already exist. So no need for path-clean and we can keep the old logic probably..
> 
> Am I missing something? @atenart what do you think?

I think you're right and it's actually the right solution, as using `OpenOptions::create_new` is not good because the output file should be accepted if the user wants to overwrite it and in such case the file would exist already.

So we can remove `path-clean` and slightly update the current logic: `canonicalize()` must work on the input file, but if it fails on the output one it's fine; if it works on both the two paths can be compared. While at it, can you add a comment on top of this to explain the different cases and why failing to canonicalize the output file is fine? Thanks!

PR #386, opened 2024-04-26
core: inspect: kernel: drop bimap dependency

ffmancera:
The `bimap` usage can be replaced by a tuple of maps where the first element represent left->right and the sencond one represent right->left.

### DIFF
--- a/retis/src/helpers/bimap.rs
+++ b/retis/src/helpers/bimap.rs
@@ -1,0 +1,149 @@
+use std::{
+    borrow::Borrow,
+    collections::BTreeMap,
+    ops::{Bound, RangeBounds},
+    sync::Arc,
+};
+
+/// An iterator over a range of key->values pairs in a `BiBTreeMap`.
+pub(crate) struct KeyRange<'a, K: 'a, V: 'a> {
+    inner: std::collections::btree_map::Range<'a, Arc<K>, Arc<V>>,
+}
+
+impl<'a, K: Ord, V> KeyRange<'a, K, V> {
+    fn new(map: &'a BTreeMap<Arc<K>, Arc<V>>, range: (Bound<&'a K>, Bound<&'a K>)) -> Self {
+        Self {
+            inner: map.range::<K, (Bound<&K>, Bound<&K>)>(range),
+        }
+    }
+}
+
+impl<'a, K: Ord, V> Iterator for KeyRange<'a, K, V> {
+    type Item = (&'a K, &'a V);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner.next().map(|(k, v)| (k.as_ref(), v.as_ref()))
+    }
+}
+
+impl<'a, K: Ord, V> DoubleEndedIterator for KeyRange<'a, K, V> {
+    fn next_back(&mut self) -> Option<Self::Item> {
+        self.inner
+            .next_back()
+            .map(|(k, v)| (k.as_ref(), v.as_ref()))
+    }
+}
+
+/// A bimap backed by two `BTreeMap`.
+pub(crate) struct BiBTreeMap<K, V> {
+    /// A `BTreeMap` of `Arc` wrapped keys and values representing the key->value
+    /// relationship.
+    left2right: BTreeMap<Arc<K>, Arc<V>>,
+    /// A `BTreeMap` of `Arc` wrapped keys and values representing the value->key
+    /// relationship.
+    right2left: BTreeMap<Arc<V>, Arc<K>>,
+}
+
+impl<K, V> BiBTreeMap<K, V>
+where
+    K: std::hash::Hash + Eq + Ord,
+    V: std::hash::Hash + Eq + Ord,
+{
+    /// Crate an empty `BiBTreeMap`.
+    pub(crate) fn new() -> Self {
+        BiBTreeMap {
+            left2right: BTreeMap::new(),
+            right2left: BTreeMap::new(),
+        }
+    }
+
+    /// Inserts the given key->value pair into the bimap.
+    pub(crate) fn insert(&mut self, key: K, value: V) {
+        let rc_key = Arc::new(key);
+        let rc_value = Arc::new(value);
+        self.right2left.insert(rc_value.clone(), rc_key.clone());
+        self.left2right.insert(rc_key, rc_value);
+    }
+
+    /// Returns a reference to the value corresponding to the given key.
+    pub(crate) fn get_by_left(&self, key: &K) -> Option<&V> {
+        self.left2right.get(key).map(|x| x.as_ref())
+    }
+
+    /// Returns a reference to the key corresponding to the given value.
+    pub(crate) fn get_by_right(&self, key: &V) -> Option<&K> {
+        self.right2left.get(key).map(|x| x.as_ref())
+    }
+
+    /// Creates an iterator over the key->value pairs within a range of keys in the bimap in
+    /// ascending order.
+    pub(crate) fn range_by_left<'a, R>(&'a self, target: &'a R) -> KeyRange<'_, K, V>
+    where
+        Arc<K>: Borrow<K>,
+        R: RangeBounds<K>,
+    {
+        let start = target.start_bound();
+        let end = target.end_bound();
+        KeyRange::new(&self.left2right, (start, end))
+    }
+
+    /// Returns the number of key->value pairs in the bimap.
+    pub(crate) fn len(&self) -> usize {
+        self.left2right.len()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::BiBTreeMap;
+    use std::ops::Bound::{Included, Unbounded};
+
+    fn bimap_init() -> BiBTreeMap<i32, i32> {
+        let mut bimap = BiBTreeMap::new();
+        bimap.insert(1, 2);
+        bimap.insert(3, 4);
+        bimap.insert(5, 6);
+        bimap.insert(9, 10);
+
+        bimap
+    }
+
+    #[test]
+    fn insert() {
+        let mut bimap = BiBTreeMap::new();
+        assert!(bimap.len() == 0);
+
+        bimap.insert(1, 2);
+        assert!(bimap.len() == 1);
+    }
+
+    #[test]
+    fn get_left() {
+        let bimap = bimap_init();
+
+        assert!(bimap.get_by_left(&1).unwrap() == &2);
+        assert!(bimap.get_by_left(&3).unwrap() == &4);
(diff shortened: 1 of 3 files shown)


ffmancera:
Working on the fmt check :)

ffmancera:
> Working on the fmt check :)

Done!

atenart (commented):
LGTM, we could merge this as-is.

If you want to go further and have some Rust fun I do have a suggestion: we could implement our own type (easier later reuse) and take the opportunity to save some memory by using referenced-counting pointers (`Rc`) instead of having twice the data/key in memory.

I can give some pointers if needed. But really, feel free to say no :)

ffmancera:
> LGTM, we could merge this as-is.
> 
> If you want to go further and have some Rust fun I do have a suggestion: we could implement our own type (easier later reuse) and take the opportunity to save some memory by using referenced-counting pointers (`Rc`) instead of having twice the data/key in memory.
> 
> I can give some pointers if needed. But really, feel free to say no :)

Count me in. Let me add this to my TODO list.

atenart:
> > LGTM, we could merge this as-is.
> > If you want to go further and have some Rust fun I do have a suggestion: we could implement our own type (easier later reuse) and take the opportunity to save some memory by using referenced-counting pointers (`Rc`) instead of having twice the data/key in memory.
> > I can give some pointers if needed. But really, feel free to say no :)
> 
> Count me in. Let me add this to my TODO list.

So I guess we can merge this now and then you can work on the above at some point. Does that sound good?

ffmancera:
> > > LGTM, we could merge this as-is.
> > > If you want to go further and have some Rust fun I do have a suggestion: we could implement our own type (easier later reuse) and take the opportunity to save some memory by using referenced-counting pointers (`Rc`) instead of having twice the data/key in memory.
> > > I can give some pointers if needed. But really, feel free to say no :)
> > 
> > 
> > Count me in. Let me add this to my TODO list.
> 
> Do you prefer to get this merged first and at some point work on the above; or to keep this open and push the above here? Both work for me.

I do not have a preference to be honest. Take the decision  you prefer, I am already working on the proposed solution we discussed above. But if you want to speed up the milestone, feel free to merge this as it is.

atenart:
> > Do you prefer to get this merged first and at some point work on the above; or to keep this open and push the above here? Both work for me.
> 
> I do not have a preference to be honest. Take the decision you prefer, I am already working on the proposed solution we discussed above. But if you want to speed up the milestone, feel free to merge this as it is.

All right, will keep this open for a while and only merge it if we quickly need it at some point. Thanks!

ffmancera:
> LGTM, we could merge this as-is.
> 
> If you want to go further and have some Rust fun I do have a suggestion: we could implement our own type (easier later reuse) and take the opportunity to save some memory by using referenced-counting pointers (`Rc`) instead of having twice the data/key in memory.
> 
> I can give some pointers if needed. But really, feel free to say no :)

So just to provide some update. I have been playing around with this. I have implemented a custom lightweight version of BiBTreeMap using `Rc` and the main problem is that `Rc` is not thread-safe therefore it cannot be shared between threads. That is a problem because we want to use it from KernelInspector struct which is used in a different thread (`src/core/inspect/inspect.rs`)

Instead, I need to use `Arc` so it can perform reference counting properly. Working on it!

ffmancera:
So what do you think about this approach? Suggestions welcomed!

src/core/inspect/bimap.rs
```diff
@@ -0,0 +1,52 @@
+use std::borrow::Borrow;
+use std::collections::{btree_map::Range, BTreeMap};
+use std::ops::RangeBounds;
+use std::sync::Arc;
```
amorenoz, line 4: nit: not really sure if it's an extended practice, but we  tend to group these kind of lines together as in:
```rust
use std::{
   borrow::Borrow,
   collections::{btree_map::Range, BTreeMap};
   ops::RangeBounds;
   sync::Arc,
};
```

src/core/inspect/bimap.rs
amorenoz, line 1: nit: I know I'm saying this after fighting against inter-module dependencies in #388 but, given this is fairly independent struct, maybe it should live in some more generic module such as "helpers".

src/core/inspect/kernel.rs
```diff
@@ -4,25 +4,27 @@ use std::{
...
-use super::{btf::BtfInfo, kernel_version::KernelVersion};
+use super::{bimap::BiBTreeMap, btf::BtfInfo, kernel_version::KernelVersion};
 use crate::core::kernel::Symbol;
 
 /// Provides helpers to inspect probe related information in the kernel.
 pub(crate) struct KernelInspector {
     /// Btf information.
     pub(crate) btf: BtfInfo,
-    /// Symbols bi-directional map (addr<>name).
+    /// Symbols bi-directional map represented by a tuple of maps. (addr<>name).
```
amorenoz, line 27: You can keep the comment saying it's a bi-directional map :-).

amorenoz (commented):
Thanks for the update!
I've added some nit comments.

Also, CI is failing because tests fail to build: `make test` should reproduce them.

src/core/inspect/bimap.rs
ffmancera: Sure thing!

src/core/inspect/bimap.rs
ffmancera: So, if there is consensus about it, I am fine doing that :)

ffmancera:
I just created a trivial PR fixing the clippy warnings not related to this one. #389

src/core/inspect/bimap.rs
ffmancera: Done!

amorenoz (commented):
Thanks  @ffmancera, the code looks good to me. My only nit would be missing documentation on the BiBTreeMap (with triple "///"). I'll leave @atenart some time to take a look as well as he did the initial review.

src/core/helpers/bimap.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub(crate) fn insert(&mut self, key: K, value: V) {
+        let rc_key = Arc::new(key);
+        let rc_value = Arc::new(value);
+        self.right2left.insert(rc_value.clone(), rc_key.clone());
+        self.left2right.insert(rc_key, rc_value);
+    }
+
+    pub(crate) fn get_by_left(&self, key: &K) -> Option<&Arc<V>> {
+        self.left2right.get(key)
+    }
```
atenart, line 34: You can probably use the following to directly return a reference on `V` (no `Arc` involved):

```rust
        self.left2right.get(key).map(|x| x.as_ref())
```

Same for `get_by_right`.

src/core/helpers/bimap.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    pub(crate) fn get_by_right(&self, key: &V) -> Option<&Arc<K>> {
+        self.right2left.get(key)
+    }
+
+    pub(crate) fn range_by_left<T, R>(&self, target: R) -> Range<'_, Arc<K>, Arc<V>>
```
atenart, line 40: Two suggestions, feel free to push back:

- We could implement out own `Range` to avoid exposing `Arc` to the users.
- We could implement `get_by_left` too even though it's not used atm.

atenart (commented):
LGTM. Two additional suggestions (feel free to ignore):

- Adding doc comments (`///`) on `BiBTreeMap`.
- Adding some unit tests.

src/core/helpers/bimap.rs
ffmancera: I am fine with implementing `range_by_left` but isn't implementing our Range too much? If we do not want to expose Arc<> to users another way is to perform here the `next_back()`.. maybe what it makes sense is to implement our own `next_back` for `Arc<K>, Arc<v>`?

This way the user won't need to call `deref`. What do you think?

atenart: > isn't implementing our Range too much? If we do not want to expose Arc<> to users another way is to perform here the `next_back()`.. maybe what it makes sense is to implement our own `next_back` for `Arc<K>, Arc<v>`?
> 
> This way the user won't need to call `deref`. What do you think?

That's what I had in mind as I don't think implementing `DoubleEndedIterator` (`next_back`) without having your own `Range` is possible. Having said that it should be quite simple: keeping a reference of the map in the custom range and implementing `DoubleEndedIterator` (and `Iterator` maybe). Both only require one function to be implemented, which I think should be a one-liner.

Or did you had something else in mind / am I missing something?

src/core/helpers/bimap.rs
ffmancera: Thanks, done!

ffmancera:
I am gonna add some unit tests and documentation, with that it should be done. Pushed to check that CI passes.

atenart:
c8s CI failure is unrelated and broken in `main`, please ignore it.

ffmancera:
Ready to review I think.

retis/src/helpers/bimap.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// relationship.
+    left2right: BTreeMap<Arc<K>, Arc<V>>,
+    /// A `BTreeMap` of `Arc` wrapped keys and values representing the value->key
+    /// relationship.
+    right2left: BTreeMap<Arc<V>, Arc<K>>,
+}
+
+impl<K, V> BiBTreeMap<K, V>
+where
+    K: std::hash::Hash + Eq + Ord + std::fmt::Display,
```
atenart, line 49: Nit: no need to require `std::fmt::Display` here.

retis/src/helpers/bimap.rs
```diff
@@ -0,0 +1,148 @@
...
+#[cfg(test)]
+mod tests {
+    use super::BiBTreeMap;
+    use std::ops::Bound::{Included, Unbounded};
+
+    fn bimap_init() -> BiBTreeMap<i32, i32> {
+        let mut bimap = BiBTreeMap::new();
+        bimap.insert(1, 2);
+        bimap.insert(3, 4);
+        bimap.insert(5, 6);
```
atenart, line 105: Nit: can we add a key not in the range of the `range_left` test here? Eg. `(9, 8)`.

atenart (changes requested):
Two minors nits, otherwise looks really nice. Will merge next version.

Thanks for the effort and for going the extra mile and adding tests!

retis/src/helpers/bimap.rs
ffmancera: Ah right. I added them for some manual testing while coding. Sorry it is a leftover. Dropping it.

atenart (approved):
LGTM, thanks!!

PR #387, opened 2024-04-26
add neq to string comparison

vlrpl:
The series makes possible matches like the following:

`retis collect -m 'sk_buff.dev.name != "lo"' -f "icmp" -p "tp:net:netif_receive_skb" -p "kprobe:__ip_local_out" -c skb --skb-sections all --cmd "ping -c 3 8.8.8.8"`


atenart (approved):
LGTM, thanks!

PR #392, opened 2024-05-29
Meta adv matches

vlrpl:
Implements  #317 plus lhs-only expressions.
- Needs unit tests.
- The doc needs some improvements
  - pseudo-EBNF requires some improvements (e.g. split numeric and string-only ops)
  - alternatively, pseudo-EBNF could be replaced by simple generic examples (which could be easier for the user to use as reference)
  
Turned into RFC.
Everything is open to discussion including:
- syntax for mask and cast (slight preference to keep it simple :)
- doc (see above)
- have a local version with some tests (I will update as soon as I find some time)

### DIFF
--- a/docs/filtering.md
+++ b/docs/filtering.md
@@ -70,10 +70,26 @@ INFO  L2 packet filter(s) loaded
 
 Metadata filtering instead allows to write filters that match packets based
 on their metadata.
-These filters can match against any subfield of the `sk_buff` and subsequent
-inner data structures.
+These filters can match against any subfield of the `sk_buff` and
+subsequent inner data structures.
 Meta filtering also automatically follows struct pointers, so indirect access to
 structures pointed by an `sk_buff` field is possible.
+A filter expression is represented by the pseudo EBNF grammar below:
+
+```none
+EXPR ::= LHS ' ' OP_RHS | LHS
+OP_RHS ::= OP ' ' RHS_NUM | EQ_NE ' ' RHS_STR
+LHS ::= 'sk_buff' MEMBER
+MEMBER ::= IDENTIFIER MEMBER | IDENTIFIER
+IDENTIFIER ::= '.' #'[a-zA-Z_][a-zA-Z0-9_]*'
+OP ::= EQ_NE | '<' | '<=' | '>' | '>='
+EQ_NE ::= '==' | '!='
+RHS_STR ::= '"' ASCII '"' | '\'' ASCII '\''
+ASCII ::= #'[:ascii:]*'
+RHS_NUM ::= #'0x[a-fA-F0-9]+' | ('-')? #'[0-9]+'
+```
+
+An example of filter that respect a previous definition is:
 
 ```none
 $ retis collect -m 'sk_buff.dev.nd_net.net.ns.inum == 4026531840'
@@ -86,6 +102,7 @@ The comparison operators are:
 2. "!=" for *not equal to*
 3. "<" and "<=" for *less than* and *less than or equal to*
 4. ">" and ">=" for *greater than* and *greater than or equal to*
+5. if OP and RHS are omitted, a *not equal to* zero numeric comparison is assumed
 
 At the moment, only number and string comparisons are supported.
 The right-hand side (rhs) of numeric matches must be expressed as
--- a/docs/filtering.md
+++ b/docs/filtering.md
@@ -70,8 +70,8 @@ INFO  L2 packet filter(s) loaded
 
 Metadata filtering instead allows to write filters that match packets based
 on their metadata.
-These filters can match against any subfield of the `sk_buff` and
-subsequent inner data structures.
+These filters can match against any subfield of the `sk_buff` and subsequent
+inner data structures.
 Meta filtering also automatically follows struct pointers, so indirect access to
 structures pointed by an `sk_buff` field is possible.
 A filter expression is represented by the pseudo EBNF grammar below:
@@ -81,12 +81,17 @@ EXPR ::= LHS ' ' OP_RHS | LHS
 OP_RHS ::= OP ' ' RHS_NUM | EQ_NE ' ' RHS_STR
 LHS ::= 'sk_buff' MEMBER
 MEMBER ::= IDENTIFIER MEMBER | IDENTIFIER
-IDENTIFIER ::= '.' #'[a-zA-Z_][a-zA-Z0-9_]*'
+IDENTIFIER ::= '.' #'[a-zA-Z_][a-zA-Z0-9_]*' (':' MASK)?
 OP ::= EQ_NE | '<' | '<=' | '>' | '>='
 EQ_NE ::= '==' | '!='
+MASK ::= ('~')? MASK_NUM
+MASK_NUM ::= HEX | DEC | BIN
 RHS_STR ::= '"' ASCII '"' | '\'' ASCII '\''
 ASCII ::= #'[:ascii:]*'
-RHS_NUM ::= #'0x[a-fA-F0-9]+' | ('-')? #'[0-9]+'
+RHS_NUM ::= HEX | ('-')? DEC
+HEX ::= #'0x[a-fA-F0-9]+'
+DEC ::= #'[0-9]+'
+BIN ::= #'0b[0-1]+'
 ```
 
 An example of filter that respect a previous definition is:
@@ -111,6 +116,25 @@ latter starting with `0x` prefix.
 All the comparison operators support numbers (both signed and unsigned).
 Bitfields are supported as well (both signed and unsigned) and they
 are treated as regular numbers.
+For numeric comparisons, an additional bitwise AND operation can be
+performed by specifying a *mask*.
+A *mask* can be expressed as a hexadecimal number (e.g. *0xdeaf*), a
+binary number (e.g. *0b01010101*), and a regular decimal number.
+The filtering engine allows you to specify masks up to **u64::MAX**
+with any target. While this approach is safe, ensuring consistency is
+the user's responsibility.
+The following example demonstrates this approach:
+
+```
+$ retis collect -m 'sk_buff._nfct:0x7 == 0x2'
+...
+```
+
+which is equivalent to the following:
+
+```none
+(sk_buff->_nfct & NFCT_INFOMASK) == IP_CT_NEW
+```
 
 For strings only the operators *equal to* and *not equal to* are supported,
 furthermore, the string (rhs) must be enclosed between *quotes*.
--- a/docs/filtering.md
+++ b/docs/filtering.md
@@ -80,8 +80,9 @@ A filter expression is represented by the pseudo EBNF grammar below:
 EXPR ::= LHS ' ' OP_RHS | LHS
 OP_RHS ::= OP ' ' RHS_NUM | EQ_NE ' ' RHS_STR
 LHS ::= 'sk_buff' MEMBER
-MEMBER ::= IDENTIFIER MEMBER | IDENTIFIER
-IDENTIFIER ::= '.' #'[a-zA-Z_][a-zA-Z0-9_]*' (':' MASK)?
+MEMBER ::= NEXTIDENT MEMBER | NEXTIDENT
+NEXTIDENT ::= '.' IDENT (':' MASK (':' IDENT)?)?
+IDENT ::= #'[a-zA-Z_][a-zA-Z0-9_]*'
 OP ::= EQ_NE | '<' | '<=' | '>' | '>='
 EQ_NE ::= '==' | '!='
 MASK ::= ('~')? MASK_NUM
@@ -146,6 +147,21 @@ $ retis collect -m 'sk_buff.dev.name == "eth0"'
 
 The example above shows how strings can be matched and how they are
 required to be quoted.
+
+Another useful feature meta filtering expose is the ability to follow
+pointers embedded in members with a different defined type.
+For example, the filter below:
+
+```none
+$ retis collect -m sk_buff._nfct:~0x7:nf_conn.mark
+...
+```
+
+is equivalent to the following:
+
+```none
+(nf_conn *)(skb->_nfct & NFCT_PTRMASK)->mark != 0
+```
 
 Metadata filtering, being a BTF-based way of filtering, is theoretically
 not limited to `sk_buff`, so from a generic point of view it can support
--- a/docs/filtering.md
+++ b/docs/filtering.md
@@ -1,0 +1,129 @@
+# Filtering
+
+Retis offers two distinct methods for filtering packets, both of which
+can operate simultaneously:
+
+- packet-based filtering which filters packets based on their content
+  (headers).
+- metadata-based filtering which filters packets based on their
+  associated metadata.
+
+These filtering mechanisms ensure that only relevant packets are
+reported, so reducing the volume of uninteresting events and
+improving the efficiency of packet tracing.
+
+## Packet
+
+Packet filtering uses a pcap-filter syntax. See `man pcap-filter` for an
+overview on the syntax. Upon execution the pcap-filter gets compiled in cBPF
+and subsequently translated into an eBPF program which in turn gets consumed by
+the probes.
+
+```none
+$ retis collect -f 'tcp port 443'
+...
+```
+
+Packet filtering can be of two types: L2 and L3. Retis automatically detects and
+generates L2/L3 filters based on the expression. This allows to match both packets
+fully formed and packets not having a valid L2 header yet (`sk_buff` having invalid
+`mac_header` but valid and set `network_header`). One advantage of this approach is
+the ability to match locally generated packets while still allowing matches based
+on L2 criteria.
+
+For example, the following filter:
+
+```none
+$ retis collect -f 'arp or tcp port 443'
+L2+L3 packet filter(s) loaded
+...
+```
+
+Internally generates two filters. For probes where only the `network_header`
+is valid and set in the `sk_buff` the filter would match packets with tcp
+source or destination port 443. For `sk_buff` with valid `mac_header` both arp
+and tcp packets would be matched.
+Please note that some limitations exist and they are a direct consequence of
+libpcap capabilities.
+For example filters like:
+
+```none
+$ retis collect -f 'ether broadcast or tcp port 443'
+L2 packet filter(s) loaded
+...
+```
+
+Will only generate L2 filters, that is, packets will be matched only if
+`mac_header` is set.
+For further information about the reason an L3 filter gets skipped, please
+use the `--log-level debug` option, i.e.:
+
+```none
+$ retis --log-level debug collect -f 'ether broadcast or tcp port 443'
+...
+DEBUG Skipping L3 filter generation (Could not compile the filter: libpcap error: not a broadcast link).
+INFO  L2 packet filter(s) loaded
+...
+```
+
+## Metadata
+
+Metadata filtering instead allows to write filters that match packets based
+on their metadata.
+These filters can match against any subfield of the `sk_buff` and subsequent
+inner data structures.
+Meta filtering also automatically follows struct pointers, so indirect access to
+structures pointed by an `sk_buff` field is possible.
+
+```none
+$ retis collect -m 'sk_buff.dev.nd_net.net.ns.inum == 4026531840'
+...
+```
+
+The comparison operators are:
+
+1. "==" for *equal to*
+2. "!=" for *not equal to*
+3. "<" and "<=" for *less than* and *less than or equal to*
+4. ">" and ">=" for *greater than* and *greater than or equal to*
+
+At the moment, only number and string comparisons are supported.
+The right-hand side (rhs) of numeric matches must be expressed as
+literal and can be represented in either base 10 or base 16, with the
+latter starting with `0x` prefix.
+All the comparison operators support numbers (both signed and unsigned).
+Bitfields are supported as well (both signed and unsigned) and they
+are treated as regular numbers.
+
+For strings only the operators *equal to* and *not equal to* are supported,
+furthermore, the string (rhs) must be enclosed between *quotes*.
+
+```none
+$ retis collect -m 'sk_buff.dev.name == "eth0"'
+...
+```
+
+The example above shows how strings can be matched and how they are
+required to be quoted.
+
+Metadata filtering, being a BTF-based way of filtering, is theoretically
+not limited to `sk_buff`, so from a generic point of view it can support
+all filters under the form *struct_type_name.field1.field2.field3* with
+the above constraints, but for the time being only `struct sk_buff` is
+supported.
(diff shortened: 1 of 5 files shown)


atenart:
Overall the changes look good.

>     * The doc needs some improvements
>       * pseudo-EBNF requires some improvements (e.g. split numeric and string-only ops)
>       * alternatively, pseudo-EBNF could be replaced by simple generic examples (which could be easier for the user to use as reference)

IMO both are useful. Doc could start with a variety of examples to show common practices and have a combination of simple and complex filters. Then the EBNF format is nice for advanced people to really understand how it works.

Additionally IMO it's time to move the filtering doc into its own page. We can keep one or two simple examples in the main doc so people get an idea about what it does, but not more.

>     * syntax for mask and cast (slight preference to keep it simple :)

I'm OK with the current proposal, so this is just to explore other possibilities:

- I guess not using `&` for masks is to avoid confusion in the shell? Other than that `sk_buff._nfct&0x7` would look nice. One way here would be to only allow complex operations in between parenthesis, eg. `sk_buff.(_nfct&0x7)`, as most shell implementations would throw an error if `(` is not in a string. This would be similar to bash bitwise operators.
- I'm wondering if we could not just use the C casting syntax, eg. `sk_buff.(nf_conn)_nfct:~0x7.mark`; or with the above, `sk_buff.(nf_conn)(_nfct&~0x7).mark`. Might not be as easy to parse though.

>     * have a local version with some tests (I will update as soon as I find some time)

Yes, please :)

A few nits too:
- Commit message `meta: parse members into structs` does not seem to match the changes.
- The mask parsing code should make sure it's not `0`. Right now I think this would result in the mask being ignored silently (not that `0` is a valid value).
- Error messages should be improved. I don't think they're easy to understand from an user point of view. A few examples (but there are more):
  - "mask for non-ptr intermediate members is not supported" -> "Intermediate members masking if only supported for pointers" and if possible with "(<show the actual intermediate member + mask>)".
  - "invalid lhs" -> split in two and have "Expression does not point to a member" and "Starting struct isn't supported (!= sk_buff)".
  - "invalid mask format" -> "mask should be in hex starting with 0x".
  - "invalid filter format". This one's not easy to improve. I'm wondering if we should consider having online references for complex errors, eg. "Invalid filter format, see https://.../error/42". (Only if we have a lot of those examples).

And finally, thanks for removing that huge regex definition :)

vlrpl:
removed RFC (of course a respin is required)

vlrpl:
> Overall the changes look good.
> 
> > ```
> > * The doc needs some improvements
> >   * pseudo-EBNF requires some improvements (e.g. split numeric and string-only ops)
> >   * alternatively, pseudo-EBNF could be replaced by simple generic examples (which could be easier for the user to use as reference)
> > ```
> 
> IMO both are useful. Doc could start with a variety of examples to show common practices and have a combination of simple and complex filters. Then the EBNF format is nice for advanced people to really understand how it works.
> 
> Additionally IMO it's time to move the filtering doc into its own page. We can keep one or two simple examples in the index so people get an idea about what it does, but not more.
> 
> > ```
> > * syntax for mask and cast (slight preference to keep it simple :)
> > ```
> 
> I'm OK with the current proposal, so this is just to explore other possibilities:
> 
>     * I guess not using `&` for masks is to avoid confusion in the shell? Other than that `sk_buff._nfct&0x7` would look nice. One way here would be to only allow complex operations in between parenthesis, eg. `sk_buff.(_nfct&0x7)`, as most shell implementations would throw an error if `(` is not in a string. This would be similar to bash bitwise operators.
> 
>     * I'm wondering if we could not just use the C casting syntax, eg. `sk_buff.(nf_conn)_nfct:~0x7.mark`; or with the above, `sk_buff.(nf_conn)(_nfct&~0x7).mark`. Might not be as easy to parse though.

Yes, I tried to avoid symbols that could potentially conflict with the shells (although between single quotes should be fine).

The last proposal is nice, but, yes, as you say the parsing might require some more work.
I thought this feature was "immutable" because there was no real reason to extend it, but tbf there are at least two types of use cases we might want to consider for the future (even though IMHO both require per-probe filters).
- `net_device` privs access: `sk_buff.(vlan_dev_priv)(dev + CONST_VAL).nr_egress_mappings`
- skb cb or in general simple ptr cast (e.g. from `void *`): `sk_buff.(inet_skb_parm)cb.flags`

the current syntax can be extended as well, having:
- `sk_buff.dev:+CONST_VAL:vlan_dev_priv.nr_egress_mappings` (no operator defaults to `&`)
- `sk_buff.cb::inet_skb_parm.flags` (mask is not optional ATM if cast is present, but one can include `~0x0` to obtain the same result)

Admittedly the first is more familiar, but still, I have no strong opinion on it as even the current syntax looks pretty easy.
Maybe we can avoid the potential parsing ugliness and use regex only for this part (and not for the whole expression), but this also depends on whether we plan to drop it or not (we still depend on it).

Not sure. What's your opinion on it?

> 
> 
> > ```
> > * have a local version with some tests (I will update as soon as I find some time)
> > ```
> 
> Yes, please :)
> 
> A few nits too:
> 
>     * Commit message `meta: parse members into structs` does not seem to match the changes.
> 
>     * The mask parsing code should make sure it's not `0`. Right now I think this would result in the mask being ignored silently (not that `0` is a valid value).
> 
>     * Error messages should be improved. I don't think they're easy to understand from an user point of view. A few examples (but there are more):
>       
>       * "mask for non-ptr intermediate members is not supported" -> "Intermediate members masking if only supported for pointers" and if possible with "(<show the actual intermediate member + mask>)".
>       * "invalid lhs" -> split in two and have "Expression does not point to a member" and "Starting struct isn't supported (!= sk_buff)".
>       * "invalid mask format" -> "mask should be in hex starting with 0x".
>       * "invalid filter format". This one's not easy to improve. I'm wondering if we should consider having online references for complex errors, eg. "Invalid filter format, see https://.../error/42". (Only if we have a lot of those examples).
> 
> 
> And finally, thanks for removing that huge regex definition :)

atenart:
> > > ```
> > > * syntax for mask and cast (slight preference to keep it simple :)
> > > ```
> > 
> > I'm OK with the current proposal, so this is just to explore other possibilities:
> > ```
> > * I guess not using `&` for masks is to avoid confusion in the shell? Other than that `sk_buff._nfct&0x7` would look nice. One way here would be to only allow complex operations in between parenthesis, eg. `sk_buff.(_nfct&0x7)`, as most shell implementations would throw an error if `(` is not in a string. This would be similar to bash bitwise operators.
> > 
> > * I'm wondering if we could not just use the C casting syntax, eg. `sk_buff.(nf_conn)_nfct:~0x7.mark`; or with the above, `sk_buff.(nf_conn)(_nfct&~0x7).mark`. Might not be as easy to parse though.
> > ```
> 
> Yes, I tried to avoid symbols that could potentially conflict with the shells (although between single quotes should be fine).
> 
> The last proposal is nice, but, yes, as you say the parsing might require some more work. I thought this feature was "immutable" because there was no real reason to extend it, but tbf there are at least two types of use cases we might want to consider for the future (even though IMHO both require per-probe filters).
> 
>     * `net_device` privs access: `sk_buff.(vlan_dev_priv)(dev + CONST_VAL).nr_egress_mappings`
> 
>     * skb cb or in general simple ptr cast (e.g. from `void *`): `sk_buff.(inet_skb_parm)(cb).flags`
> 
> 
> the current syntax can be extended as well, having:
> 
>     * `sk_buff.dev:+CONST_VAL:vlan_dev_priv.nr_egress_mappings` (no operator defaults to `&`)
> 
>     * `sk_buff.cb::inet_skb_parm.flags` (mask is not optional ATM if cast is present, but one can include `~0x0` to obtain the same result)
> 
> 
> Admittedly the first is more familiar to C dev (but things are different anyway), but still, I have no strong opinion on it as even the current syntax looks pretty easy. Maybe we can avoid the potential parsing ugliness and use regex only for this part (and not for the whole expression), but this also depends on whether we plan to drop it or not (we still depend on it).
> 
> Not sure. What's your opinion on it?

I'm not against using regex in general and I don't think it will go away; what I did not like in the way it was used was it was a very complex one which made reading and maintaining the code quite hard. Other than that, as long as it's not in the hot path, IMO it's fine to use regex.

Having said that, the C syntax version of the filter would actually be very hard to get right, as things like `(cast)(field + val)` and `(cast)(val + field)` should yield the same result; and it could be combined with masking.

Having a stricter syntax actually makes things more sane and easier to work with and verify. I find the `field:expr:expr` formatting nice looking and I think we could rely on the order of the expression and not on any special char, something like `field:mask:offset:cast`. Eg. `sk_buff.field:0xff:42:some_struct`. I also find it OK to leave some parts blank if not used, eg. `sk_buff.cb:::inet_skb_parm.flags` (that only works if we support 3-4 expressions max, for readability[1]).

[1] We can always support something like `field[:[mask:offset:]cast]` to support the easy case where only a cast is needed.

vlrpl:
> Overall the changes look good.
> 
> > ```
> > * The doc needs some improvements
> >   * pseudo-EBNF requires some improvements (e.g. split numeric and string-only ops)
> >   * alternatively, pseudo-EBNF could be replaced by simple generic examples (which could be easier for the user to use as reference)
> > ```
> 
> IMO both are useful. Doc could start with a variety of examples to show common practices and have a combination of simple and complex filters. Then the EBNF format is nice for advanced people to really understand how it works.
> 
> Additionally IMO it's time to move the filtering doc into its own page. We can keep one or two simple examples in the index so people get an idea about what it does, but not more.
> 
> > ```
> > * syntax for mask and cast (slight preference to keep it simple :)
> > ```
> 
> I'm OK with the current proposal, so this is just to explore other possibilities:
> 
>     * I guess not using `&` for masks is to avoid confusion in the shell? Other than that `sk_buff._nfct&0x7` would look nice. One way here would be to only allow complex operations in between parenthesis, eg. `sk_buff.(_nfct&0x7)`, as most shell implementations would throw an error if `(` is not in a string. This would be similar to bash bitwise operators.
> 
>     * I'm wondering if we could not just use the C casting syntax, eg. `sk_buff.(nf_conn)_nfct:~0x7.mark`; or with the above, `sk_buff.(nf_conn)(_nfct&~0x7).mark`. Might not be as easy to parse though.
> 
> 
> > ```
> > * have a local version with some tests (I will update as soon as I find some time)
> > ```
> 
> Yes, please :)
> 
> A few nits too:
> 
>     * Commit message `meta: parse members into structs` does not seem to match the changes.
> 
>     * The mask parsing code should make sure it's not `0`. Right now I think this would result in the mask being ignored silently (not that `0` is a valid value).
> 
>     * Error messages should be improved. I don't think they're easy to understand from an user point of view. A few examples (but there are more):
>       
>       * "mask for non-ptr intermediate members is not supported" -> "Intermediate members masking if only supported for pointers" and if possible with "(<show the actual intermediate member + mask>)".
>       * "invalid lhs" -> split in two and have "Expression does not point to a member" and "Starting struct isn't supported (!= sk_buff)".
>       * "invalid mask format" -> "mask should be in hex starting with 0x".
>       * "invalid filter format". This one's not easy to improve. I'm wondering if we should consider having online references for complex errors, eg. "Invalid filter format, see https://.../error/42". (Only if we have a lot of those examples).
> 

This particular instance gets updated/removed by the next patches, but I see your point.
Having an online reference would be nice, and maybe we could consider creating it regardless (to expand, maybe with an example, terse messages that could seem a bit obscure to the casual user)

> 
> And finally, thanks for removing that huge regex definition :)

retis/src/core/filters/meta/filter.rs
```diff
@@ -391,75 +413,131 @@ impl FilterMeta {
...
-            if Self::check_one_walkable(&x, &mut ind)? {
+            if Self::check_one_walkable(&x, &mut ind, casted)? {
                 return Ok((ind, x));
             }
         }
 
         bail!("failed to retrieve next walkable object.")
     }
 
+    fn parse_mask(el: &str) -> Result<u64> {
```
atenart, line 451: We should check `mask != 0` at the end of this function, otherwise setting a mask of 0 actually means setting a mask of `~0` (as mask is not used).

retis/src/core/filters/meta/filter.rs
```diff
@@ -391,75 +413,131 @@ impl FilterMeta {
...
-        let init_sym = fields.remove(0);
-
-        if !init_sym.eq("sk_buff") {
-            bail!("unsupported data structure {init_sym}. sk_buff must be used.")
-        }
-
-        let mut types = btf
+        let mut types = btf_info
             .resolve_types_by_name(init_sym)
             .map_err(|e| anyhow!("unable to resolve sk_buff data type {e}"))?;
```
atenart, line 531: nit: Not part of this MR, but technically this could be another struct. The check for `sk_buff` is done elsewhere and could be extended.

retis/src/core/filters/meta/filter.rs
```diff
@@ -478,32 +556,52 @@ impl FilterMeta {
...
+                                op.l.mask = field.mask;
                                 ops.push(op);
                             }
                             std::cmp::Ordering::Greater => {
                                 bail!("pointers of pointers are not supported")
                             }
-                            _ => offt = offset,
+                            _ => {
+                                if field.mask != 0 {
+                                    bail!("intermediate members masking is only supported for pointers");
```
atenart, line 567: This not strictly true, some unsigned numbers can be walkable when casted.

retis/src/core/filters/meta/filter.rs
```diff
@@ -478,32 +556,52 @@ impl FilterMeta {
...
+                        mask = field.mask;
                     }
 
                     stored_offset = offset;
                     if let Some(bfs) = bfs {
                         stored_bf_size = bfs;
                     }
                 }
-                None => bail!("{field} not found or is a bitfield!"),
+                None => bail!("{} not found!", field.member),
```
atenart, line 598: nit: "field {} not found in type {}"

retis/src/core/filters/meta/filter.rs
```diff
@@ -391,75 +413,131 @@ impl FilterMeta {
...
+                .try_into()
+                .map_err(|_| anyhow!("cannot split filter ({filter})"))?,
+            1 => [expr[0], "!=", "0"],
+            _ => bail!("invalid filter ({filter})"),
+        };
+
+        let lhs: Vec<_> = lhs
+            .split('.')
+            .map(|x| {
+                let mut elem = x.split(':');
```
atenart, line 482: nit: We should make sure len is not > 2, or throw an error.

atenart (commented):
- We should restrict masking in intermediate fields only if casting is enabled. Otherwise this can lead to weird things.
- Masking and casting the initial struct is currently allowed (from what I can tell, in the filter syntax only, not in practice); this should trigger an error.
- Casting the leaf field is allowed in the syntax, this should trigger an error.

retis/src/core/filters/meta/filter.rs
vlrpl: yes, this fell through the cracks. Harmless but not needed.

retis/src/core/filters/meta/filter.rs
```diff
@@ -379,8 +394,15 @@ impl Rval {
...
-    fn check_one_walkable(t: &Type, ind: &mut u8) -> Result<bool> {
+    fn check_one_walkable(t: &Type, ind: &mut u8, casted: bool) -> Result<bool> {
         match t {
+            Type::Int(i) => {
+                if casted && i.size() == 8 {
+                    *ind += 1;
+                } else {
+                    bail!("found {} while walking types", t.name())
+                }
+            }
```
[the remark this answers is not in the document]
vlrpl: although we don't support 32bits, I think this could be converted into something like [sizeof(void *)](https://godbolt.org/z/objMxvzG9)

In that case `>=` would make sense but honestly I would not add that support (e.g. for cases like `(struct foo *)(num >> 4)`)

vlrpl:
> * We should restrict masking in intermediate fields only if casting is enabled. Otherwise this can lead to weird things.
> 

IMHO, weird request weird behavior (or UB :)
Sometimes pointers (aligned) embed other data in the lowest bits and the user might want to mask them out (`~0x3`).
Not sure about this.

>     * Masking and casting the initial struct is currently allowed (from what I can tell, in the filter syntax only, not in practice); this should trigger an error.
> 
>     * Casting the leaf field is allowed in the syntax, this should trigger an error.

ack for the other two

docs/filtering.md
```diff
@@ -0,0 +1,186 @@
...
+EXPR ::= LHS ' ' OP_RHS | LHS
+OP_RHS ::= OP ' ' NUMBER | EQ_NE ' ' STRING
+LHS ::= 'sk_buff' MEMBER
+MEMBER ::= NEXTIDENT MEMBER | NEXTIDENT
+NEXTIDENT ::= '.' IDENT (':' MASK (':' IDENT)?)?
+IDENT ::= #'[a-zA-Z_][a-zA-Z0-9_]*'
+OP ::= EQ_NE | '<' | '<=' | '>' | '>='
+EQ_NE ::= '==' | '!='
+MASK ::= ('~')? MASK_NUM
+MASK_NUM ::= NUMBER | BIN
```
atenart, line 89: Isn't BIN a number (HEX is)?

retis/src/core/filters/meta/filter.rs
```diff
@@ -391,75 +411,153 @@ impl FilterMeta {
...
+                    0x0
+                };
 
-        // The captures ensure at least two elements are present
-        let init_sym = fields.remove(0);
+                // tgt_type is optional.
+                let tgt_type = elem.next();
+
+                if elem.next().is_some() {
+                    bail!("unexpected field expression (must be under the form field:mask:type)");
```
atenart, line 518: nit: `field[:mask[:type]]`?

atenart (commented):
Not sure why but looks like filter like this work `sk_buff.dev:0xf.ifindex == 2` while I would expect the load of `dev:0xf` to fail.

atenart:
> > * We should restrict masking in intermediate fields only if casting is enabled. Otherwise this can lead to weird things.
> 
> IMHO, weird request weird behavior (or UB :) Sometimes pointers (aligned) embed other data in the lowest bits and the user might want to mask them out (`~0x3`). Not sure about this.

Right; I agree with the above, let's keep mask w/o casting working on intermediate fields.

docs/filtering.md
vlrpl: yes :)
the problem is that NUMBER is used for right-hand side definition and binary nums are not supported.
Do you prefer to s/NUMBER/DEC_HEX/ ?

atenart: Maybe we could define an helper to parse such numbers and support all variants for all usages?

vlrpl:
> Not sure why but looks like filter like this work `sk_buff.dev:0xf.ifindex == 2` while I would expect the load of `dev:0xf` to fail.

masking is allowed for (intermediate) pointers. Admittedly, the expression doesn't make any sense, but we cannot check what's a valid address. Consistency is intentionally up to the user.

atenart:
> > Not sure why but looks like filter like this work `sk_buff.dev:0xf.ifindex == 2` while I would expect the load of `dev:0xf` to fail.
> 
> masking is allowed for (intermediate) pointers. Admittedly, the expression doesn't make any sense, but we cannot check what's a valid address. Consistency is intentionally up to the user.

Right, but I meant, the filter does work as if no masking was applied. Let's use a more complex value (hard to find by chance in memory): `'sk_buff.dev:0x1.name == "enp44s0u1u1c2"'`. This does work as expected (I tested with `!=` too). Or I am missing something?

atenart:
> > > * We should restrict masking in intermediate fields only if casting is enabled. Otherwise this can lead to weird things.
> > 
> > 
> > IMHO, weird request weird behavior (or UB :) Sometimes pointers (aligned) embed other data in the lowest bits and the user might want to mask them out (`~0x3`). Not sure about this.
> 
> Right; I agree with the above, let's keep mask w/o casting working on intermediate fields.

Might be because of this:

```diff
--- a/retis/src/core/filters/meta/bpf/include/meta_filter.h
+++ b/retis/src/core/filters/meta/bpf/include/meta_filter.h
@@ -107,7 +107,7 @@ static __always_inline long meta_process_ops(struct retis_meta_ctx *ctx)
                                                  (char *)ctx->base + (val->l.offt)))
                                return -1;
 
-                       ctx->base = ctx->mask ? (void *)(ptr & ctx->mask)
+                       ctx->base = val->l.mask ? (void *)(ptr & val->l.mask)
                                              : (void *)ptr;
 
                        continue;
```

`ctx->mask` is not populated yet above.

docs/filtering.md
vlrpl: problem is that on RHS negative decimals are allowed.
I guess this would require a little bit more than simply adding a shared helper.
Not a huge deal, but maybe can be done as a follow-up. 
Does this sound ok for you?

While we speak (negative dec), I realize that the doc requires an update 
to match the syntax:

```diff
diff --git a/docs/filtering.md b/docs/filtering.md
index 59522fd9..fbd902f0 100644
--- a/docs/filtering.md
+++ b/docs/filtering.md
@@ -78,7 +78,7 @@ A filter expression is represented by the pseudo EBNF grammar below:
 
 ```none
 EXPR ::= LHS ' ' OP_RHS | LHS
-OP_RHS ::= OP ' ' NUMBER | EQ_NE ' ' STRING
+OP_RHS ::= OP ' ' RHS_NUM | EQ_NE ' ' RHS_STR
 LHS ::= 'sk_buff' MEMBER
 MEMBER ::= NEXTIDENT MEMBER | NEXTIDENT
 NEXTIDENT ::= '.' IDENT (':' MASK (':' IDENT)?)?
@@ -86,10 +86,10 @@ IDENT ::= #'[a-zA-Z_][a-zA-Z0-9_]*'
 OP ::= EQ_NE | '<' | '<=' | '>' | '>='
 EQ_NE ::= '==' | '!='
 MASK ::= ('~')? MASK_NUM
-MASK_NUM ::= NUMBER | BIN
-STRING ::= '"' ASCII '"' | '\'' ASCII '\''
+MASK_NUM ::= BIN | DEC | HEX
+RHS_STR ::= '"' ASCII '"' | '\'' ASCII '\''
 ASCII ::= #'[:ascii:]*'
-NUMBER ::= HEX | DEC
+RHS_NUM ::= HEX | ('-')? DEC
 HEX ::= #'0x[a-fA-F0-9]+'
 DEC ::= #'[0-9]+'
 BIN ::= #'0b[0-1]+'
```

atenart: > problem is that on RHS negative decimals are allowed. I guess this would require a little bit more than simply adding a shared helper. Not a huge deal, but maybe can be done as a follow-up. Does this sound ok for you?

IMO it's important for the UX to always unify behaviors as much as possible (here allowing the same set of bases for numbers), it should be possible to disallow unsigned numbers after calling that helpers or using a flag as a parameter.

Having said that I don't want to block this only for this reason, if a follow-up is coming.

atenart (approved):
LGTM, thanks!

PR #393, opened 2024-06-06
Meta remove unions

vlrpl:
Some clean-ups that serve as a preparation for multiple-field matching.
Submitted separately as they make sense regardless.


atenart (approved):
LGTM, thanks!

PR #400, opened 2024-06-20
core: events: bpf: do not exit polling thread on EINTR

### LINKED ISSUE #397: retis collect doesn't survive pause+resume
```
# retis collect
Collector(s) started: skb-tracking, ct, ovs, skb-drop, skb
^Z
[1]+  Stopped                 retis collect
[root@rhel9 ovs]# fg
retis collect
WARN  libbpf: prog 'probe_kprobe': BPF program load failed: Invalid argument
WARN  libbpf: prog 'probe_kprobe': -- BEGIN PROG LOAD LOG --
fd 20 is not pointing to valid bpf_map
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
-- END PROG LOAD LOG --
WARN  libbpf: prog 'probe_kprobe': failed to load: -22
WARN  libbpf: failed to load object 'kprobe_bpf'
Error: Invalid argument (os error 22)
# 
# echo $?
1
```

atenart:
While polling the ringbuffer -EINTR can be returned on various occasions. Some are valid reasons to exit the polling thread (^C, running state info from rb handler, etc) but some are not (^Z). We can safely omit explicitly breaking the polling loop as it already depends on Running.

Fixes #397.

Fixes: 95a690e99fdd ("allow to gracefully interrupt the tool.")

### DIFF
--- a/retis/src/core/events/bpf.rs
+++ b/retis/src/core/events/bpf.rs
@@ -153,11 +153,10 @@ impl BpfEventsFactory {
                         // ringbuffer. This could normally be
                         // triggered by an actual interruption
                         // (signal) or artificially from the
-                        // callback. Exit without printing any error.
+                        // callback. Do not print any error.
                         libbpf_rs::ErrorKind::Interrupted => (),
                         _ => error!("Unexpected error while polling ({e})"),
                     }
-                    break;
                 }
             }
         }))


vlrpl (approved):
It's definitely too narrow given other signals e.g. SIGTSTP may trigger this.
the change lgtm.

PR #410, opened 2024-07-02
docs: contributing: add a conscious language point

atenart:


ffmancera:
Hi, it looks good to me. Just a small nit: you could add a link as a source of what conscious language is. Like https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language but is not a must. I think this message is good enough :)

atenart:
> Hi, it looks good to me. Just a small nit: you could add a link as a source of what conscious language is. Like https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language but is not a must. I think this message is good enough :)

Thanks for looking at this PR! I agree having a reference page on what is conscious language would be a plus, however I feel the above link is more about the who and why than the what. Do you know of any other web page describing the what and the why of conscious language?

ffmancera:
> > Hi, it looks good to me. Just a small nit: you could add a link as a source of what conscious language is. Like https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language but is not a must. I think this message is good enough :)
> 
> Thanks for looking at this PR! I agree having a reference page on what is conscious language would be a plus, however I feel the above link is more about the who and why than the what. Do you know of any other web page describing the what and the why of conscious language?

Yes, there are some:

- If you want something more "technical": https://datatracker.ietf.org/doc/html/draft-knodel-terminology-02
- If you want something more for general public: https://buffer.com/resources/inclusive-language-tech/

I personally prefer the first one because the "general" one is in my opinion, talking about some other stuff not related at all.

amorenoz:
> > > Hi, it looks good to me. Just a small nit: you could add a link as a source of what conscious language is. Like https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language but is not a must. I think this message is good enough :)
> > 
> > 
> > Thanks for looking at this PR! I agree having a reference page on what is conscious language would be a plus, however I feel the above link is more about the who and why than the what. Do you know of any other web page describing the what and the why of conscious language?
> 
> Yes, there are some:
> 
>     * If you want something more "technical": https://datatracker.ietf.org/doc/html/draft-knodel-terminology-02
> 
>     * If you want something more for general public: https://buffer.com/resources/inclusive-language-tech/
> 
> 
> I personally prefer the first one because the "general" one is in my opinion, talking about some other stuff not related at all.

Just for the sake of completeness, in OVS we have  used this link as a reference:

https://inclusivenaming.org/word-lists/

atenart:
I read all the above links, and to be fair I'm not 100% happy with any of them for various reasons. If that's OK let's not include a link for now and if we find a good one we can add it later. Not having such a link is not an issue IMO.

vlrpl:
> I read all the above links, and to be fair I'm not 100% happy with any of them for various reasons. If that's OK let's not include a link for now and if we find a good one we can add it later. Not having such a link is not an issue IMO.

I agree, not having a specific link is fine for the time being.
I'll wait a few more hours to see if any objections come up before merging this PR.
Thanks to everyone for pointing out some alternatives.

vlrpl (approved):
lgtm, thanks!

PR #412, opened 2024-07-04
Benchmark: build the raw event instead of using a static slice

atenart:
Instead of using a statically defined slice containing a valid raw event, build one at runtime. This has the immense advantage of not having to fight that static definition whenever we modify raw event sections. As a side effect it's easier to control the raw event and the raw section content, opening the way for fine grained benchmarks.


atenart:
Rebased on top of main and fixed conflicts. Some modifications in the ct raw event modifications due to the recent changes in the ct collector (parent info support).

amorenoz (approved):
LGTM. Thanks

PR #413, opened 2024-07-16
docs: install: fix copr link

atenart:
The doc then used the right instructions, pointing to the right copr; that's probably why we didn't caught it before.


vlrpl (approved):
lgtm, thanks!

PR #414, opened 2024-07-16
core: inspect: kernel: looks for exact matches on unset kconfig options

atenart:
Unset kconfig options are commented out and follow a specific syntax. We were only looking for commented CONFIG_ option but this turned out to be an issue if a kconfig option is in a real comment and not unset. Eg,

```
  # The kernel was built with
  # CONFIG_SYSTEM_TRUSTED_KEYS=""
  # We are resetting this value to facilitate local builds
  CONFIG_SYSTEM_TRUSTED_KEYS=""
```

To support the above, be more precise on how we match unset kconfig options, by not only looking for the '# CONFIG_" prefix but also for the " is not set" suffix.
 
While at it improve slightly the error messages to know which kind of kconfig option cannot be parsed in case we have an issue.

Also add a test for unset kconfig option in the unit tests and add a test for the above option and the corresponding snippet in our kconfig file.


vlrpl (approved):
Tested this succesfully. LGTM, thanks!

PR #416, opened 2024-07-17
Fix c8s runtime CI

atenart:
Main goal was to fix the c8s pipeline, which is done by fixing the c8s repo urls. While at it also transitioned to use boxes provided by the upstream projects (Fedora, CentOS Stream); this allows to have more recent images.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -36,7 +36,8 @@ Vagrant.configure("2") do |config|
   config.vm.box_check_update = false
 
   config.vm.define "x86_64-f40" do |fedora|
-    fedora.vm.box = "fedora/40-cloud-base"
+    fedora.vm.box = "fedora-40-cloud"
+    fedora.vm.box_url = get_box("https://dl.fedoraproject.org/pub/fedora/linux/releases/40/Cloud/x86_64/images/", /.*vagrant\.libvirt\.box$/)
 
     fedora.vm.provision "common", type: "shell", inline: $bootstrap_rhel_common
     fedora.vm.provision "shell", inline: <<-SHELL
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -36,7 +36,8 @@ Vagrant.configure("2") do |config|
   config.vm.box_check_update = false
 
   config.vm.define "x86_64-f40" do |fedora|
-    fedora.vm.box = "fedora/40-cloud-base"
+    fedora.vm.box = "fedora-40-cloud"
```
vlrpl, line 39: are we ok with the current cache handling?
So far we have:

```yaml
  vagrant_cache: &vagrant_cache
    reupload_on_changes: true
    fingerprint_script: |
      # for rawhide invalidate once a week
      if [ x${DISTRO} = xrawhide ]; then
          date -d 'next monday - 1 week' +%Y%m%d
      fi
      echo ${DISTRO}; cat Vagrantfile
    folder: /root/.vagrant.d/boxes
```

I guess we have to update this.

atenart: Ah right, good catch!

I'm wondering if we should not just remove the `DISTRO` check and have that fingerprint expiring for all boxes regardless of their origin. Should not be huge difference for the non-custom ones and would simplify maintenance here. WDYT?

vlrpl: TL;DR: I don't see any downside with the proposal, so I'm ok with it (especially because we will not special case anymore :)

The ones retrieved from the repository depend on the checksum to determine their validity. Consequently, these images will be re-downloaded every week (plus every image update), similar to those not from the repository, if the date is added unconditionally. IMO, this approach should not significantly affect us unless there is a traffic quota in place, which I am not aware of.

vlrpl (approved):
lgtm, thanks!

PR #418, opened 2024-07-18
docs: requirement: refine page

atenart:
- Update the Kconfig list.
- Remove the mandatory vs option listing. Retis does require all of them and can work in degraded mode if some of them are missing, but that is not advised.
- Remove the obsolete /etc/os-release requirement.
- Add CentOS Stream 8 (>= 8.6) as being supported.


vlrpl (approved):
lgtm

PR #419, opened 2024-08-01
Use bindgen for common structures and types

vlrpl:
So far we manage common definitions by keeping them aligned manually.
This is inconvenient other than error-prone.

This PR is far from being complete and only aims to be a proposal to improve this aspect.
The current strategy consists of annotating types we need to redefine in Rust and generating the Rust types through the new `gen-bindings` target.
The general idea is to commit those generated bindings, but it's perfectly possible to refresh them at every fresh build (in this case they should not live under `src/` but under something like `env!("OUT_DIR")` or any other `.gitignore`d dir.

Notes: 

- `bindgen` recursively generates bindings for `--allowlist-item`, so if another include defines a `__binding` that is referenced in another header file, the type will be generated twice.
- `bindgen` cannot be fed with multiple C header files (a wrapping header that includes multiple ones is required for this), nor can it append new definitions into an already existing file.
- `Default` trait could be automatically derived, but that would prevent potential consumers from redefining their custom implementation, so no `Default` trait is derived.


atenart:
> This PR is far from being complete and only aims to be a proposal to improve this aspect.

nit: Could you mark it as "RFC" so it doesn't get merged before being complete?

> The current strategy consists of annotating types we need to redefine in Rust and generating the Rust types through the new `gen-bindings` target. The general idea is to commit those generated bindings, but it's perfectly possible to refresh them at every fresh build (in this case they should not live under `src/` but under something like `env!("OUT_DIR")` or any other `.gitignore`d dir.

I think having a separate make target and including the generated bindings in the sources is a good approach, as it makes the build to be 100% reproducible and allows to review the generated bindings before merging.

>     * `bindgen` recursively generates bindings for `--allowlist-item`, so if another include defines a `__binding` that is referenced in another header file, the type will be generated twice.

Does that means if types A and B are `__binding` and B is also a field in A, the binding for B will be generated twice?

>     * `bindgen` cannot be fed with multiple C header files (a wrapping header that includes multiple ones is required for this), nor can it append new definitions into an already existing file.

I guess we can come up with a expectation on where should be defined types used for the BPF <> Rust communication. Like what we do for BPF hooks and BPF includes. Eg. not a single header but expected paths/files that we can either use directly or include somewhere automatically when generating bindings.

Some high level comments:

- Having annotations and a script to get the list of what types should be generated is nice. I would also be OK with an explicitly defined list somewhere (could be in the `Makefile`).
- I don't think that will be an issue, but could you check with #412 to see if there is any incompatibility between the two efforts?
- Not really related, but IMO this show why we should get rid of `ebpf.mk` and have everything in the top-level `Makefile`.

Overall I think this will be a nice improvement.

vlrpl:
> > This PR is far from being complete and only aims to be a proposal to improve this aspect.
> 
> nit: Could you mark it as "RFC" so it doesn't get merged before being complete?
> 
> > The current strategy consists of annotating types we need to redefine in Rust and generating the Rust types through the new `gen-bindings` target. The general idea is to commit those generated bindings, but it's perfectly possible to refresh them at every fresh build (in this case they should not live under `src/` but under something like `env!("OUT_DIR")` or any other `.gitignore`d dir.
> 
> I think having a separate make target and including the generated bindings in the sources is a good approach, as it makes the build to be 100% reproducible and allows to review the generated bindings before merging.

That's the reason I was considering this approach. The only downside is that the clang args passed to `bindgen` and `vmlinux.h` include the arch type, so we might end up generating arch-specific bindings which by all means will be identical. I'm not sure this is really a bad thing though.

> 
> > ```
> > * `bindgen` recursively generates bindings for `--allowlist-item`, so if another include defines a `__binding` that is referenced in another header file, the type will be generated twice.
> > ```
> 
> Does that means if types A and B are `__binding` and B is also a field in A, the binding for B will be generated twice?
> 

yes, twice but in two separate files. E.g. given

`/foo/path/bpf/fileA.c::BindingA` and `/foobar/path/bpf/fileB.c::BindingB` we'll end up having:

- `/foo/path/fileA_uapi.rs` with both `BindingA` and `BindingB`
- `/foo/path/fileB_uapi.rs` with `BindingB` only

If we don't annotate `BindingB`, only `fileA_uapi.rs` should be generated with both types.

> > ```
> > * `bindgen` cannot be fed with multiple C header files (a wrapping header that includes multiple ones is required for this), nor can it append new definitions into an already existing file.
> > ```
> 
> I guess we can come up with a expectation on where should be defined types used for the BPF <> Rust communication. Like what we do for BPF hooks and BPF includes. Eg. not a single header but expected paths/files that we can either use directly or include somewhere automatically when generating bindings.
> 
> Some high level comments:
> 
>     * Having annotations and a script to get the list of what types should be generated is nice. I would also be OK with an explicitly defined list somewhere (could be in the `Makefile`).
> 

I thought about a list, but if the script doesn't show significant drawbacks, it's probably handier from the dev standpoint.
One limitation the approach has is that we might probably need to turn `#define` into `const ...` as of course we cannot annotate macros.

>     * I don't think that will be an issue, but could you check with [Benchmark: build the raw event instead of using a static slice #412](https://github.com/retis-org/retis/pull/412) to see if there is any incompatibility between the two efforts?
> 

yep, makes sense to have a deeper look at it

>     * Not really related, but IMO this show why we should get rid of `ebpf.mk` and have everything in the top-level `Makefile`.
> 
> 
> Overall I think this will be a nice improvement.

atenart:
> > > The current strategy consists of annotating types we need to redefine in Rust and generating the Rust types through the new `gen-bindings` target. The general idea is to commit those generated bindings, but it's perfectly possible to refresh them at every fresh build (in this case they should not live under `src/` but under something like `env!("OUT_DIR")` or any other `.gitignore`d dir.
> > 
> > I think having a separate make target and including the generated bindings in the sources is a good approach, as it makes the build to be 100% reproducible and allows to review the generated bindings before merging.
> 
> That's the reason I was considering this approach. The only downside is that the clang args passed to `bindgen` include the arch type (for our `vmlinux.h`), so we might end up generating arch-specific bindings which by all means will be identical. I'm not sure this is really a bad thing though.

IMO we should try to avoid having architecture-specific bindings. The reason we have different `vmlinux.h` files is given the amount of things defined in there, there are some architecture-specific ones. But for our raw events, I think we can try to only use sized types and then only generate the bindings once. I'm not sure if we can automatically make bindgen fail non problematic types, but in any way we should be able to see this at review time / see a failure in the CI.

> > > ```
> > > * `bindgen` recursively generates bindings for `--allowlist-item`, so if another include defines a `__binding` that is referenced in another header file, the type will be generated twice.
> > > ```
> > 
> > Does that means if types A and B are `__binding` and B is also a field in A, the binding for B will be generated twice?
> 
> yes, twice but in two separate files. E.g. given
> 
> `/foo/path/bpf/fileA.c::BindingA` and `/foobar/path/bpf/fileB.c::BindingB` we'll end up having:
> 
>     * `/foo/path/fileA_uapi.rs` with both `BindingA` and `BindingB`
> 
>     * `/foo/path/fileB_uapi.rs` with `BindingB` only
> 
> If we don't annotate `BindingB`, only `fileA_uapi.rs` should be generated with both types.

OK. I guess including a `__binding` into another one is not something we'll often see. We can live with that.

vlrpl:
An older PR  manipulates the data structures and also introduces a macro to unify raw events.
The macro has to be essentially reverted, but the number of conflicts will make me very happy when I'm going to fix all the problems :)

That aside, this is half of an RFC, but there are a couple of things I would like to clarify before declaring it official.
ATM, given a file called `bpf/a.{bpf.c,h}` we generate `a_uapi.rs` essentially creating one level above `bpf/`. We then include those bindings with `include_binding!()` and consume the content.
This leads to many scattered files around. I'm unsure whether this is the best approach or not.
Probably, having some directory under `src/bindings/` would help to make things a little cleaner, but we should play well while exporting them (not a big deal anyway, even if we export everything crate-wide).

Any preference/opinion is welcome.

vlrpl:
The `bindings/` approach was preferred and proved to be good.
Most of the common code got converted in this PR, but things are getting unmanageable so it's easy enough for anyone to migrate the remaining code in a follow-up.

The bindings do not depend on endianness, so they can be generated from any arch at the moment.
Please be careful potentially requesting changes that can give birth to a chain of conflicts (unless, of course, it's an important thing :)

This turned out to be needed, but it's extremely boring (just wanted to reiterate it :D)

vlrpl:
> > > > The current strategy consists of annotating types we need to redefine in Rust and generating the Rust types through the new `gen-bindings` target. The general idea is to commit those generated bindings, but it's perfectly possible to refresh them at every fresh build (in this case they should not live under `src/` but under something like `env!("OUT_DIR")` or any other `.gitignore`d dir.
> > > 
> > > 
> > > I think having a separate make target and including the generated bindings in the sources is a good approach, as it makes the build to be 100% reproducible and allows to review the generated bindings before merging.
> > 
> > 
> > That's the reason I was considering this approach. The only downside is that the clang args passed to `bindgen` include the arch type (for our `vmlinux.h`), so we might end up generating arch-specific bindings which by all means will be identical. I'm not sure this is really a bad thing though.
> 
> IMO we should try to avoid having architecture-specific bindings. The reason we have different `vmlinux.h` files is given the amount of things defined in there, there are some architecture-specific ones. But for our raw events, I think we can try to only use sized types and then only generate the bindings once. I'm not sure if we can automatically make bindgen fail non problematic types, but in any way we should be able to see this at review time / see a failure in the CI.
> 

of course, parameters and a bunch of other stuff.
As long as we don't use architecture-specific flavors of data structures we should be fine here. We just need to generate them and commit regardless of the host's arch.

> > > > ```
> > > > * `bindgen` recursively generates bindings for `--allowlist-item`, so if another include defines a `__binding` that is referenced in another header file, the type will be generated twice.
> > > > ```
> > > 
> > > 
> > > Does that means if types A and B are `__binding` and B is also a field in A, the binding for B will be generated twice?
> > 
> > 
> > yes, twice but in two separate files. E.g. given
> > `/foo/path/bpf/fileA.c::BindingA` and `/foobar/path/bpf/fileB.c::BindingB` we'll end up having:
> > ```
> > * `/foo/path/fileA_uapi.rs` with both `BindingA` and `BindingB`
> > 
> > * `/foo/path/fileB_uapi.rs` with `BindingB` only
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If we don't annotate `BindingB`, only `fileA_uapi.rs` should be generated with both types.
> 
> OK. I guess including a `__binding` into another one is not something we'll often see. We can live with that.

vlrpl:
rebased on top of current main

vlrpl:
yet another rebase-on-top-of-current-main fun :)

retis/src/module/nft/bpf/nft.bpf.c
```diff
@@ -17,28 +18,33 @@
 #define retis_get_nft_type(ctx, cfg)		\
 	RETIS_HOOK_GET(ctx, cfg->offsets, nft_type, enum nft_trace_types)
 
-/* Nft hook configuration.
```
atenart, line 20: nit: this could stay.

retis/src/module/skb/bpf/skb_hook.bpf.c
```diff
@@ -10,43 +10,40 @@
...
-#define COLLECT_NS		3
-#define COLLECT_META		4
-#define COLLECT_DATA_REF	5
-#define COLLECT_GSO		6
-
-/* Skb hook configuration. A map is used to set the config from userspace.
- *
- * Please keep in sync with its Rust counterpart in module::skb::bpf.
+/* Skb raw event sections. */
+enum skb_opts {
```
atenart, line 14: nit: s/skb_opts/skb_sections/

retis/src/module/skb/bpf/skb_hook.bpf.c
```diff
@@ -10,43 +10,40 @@
...
 	__uint(type, BPF_MAP_TYPE_ARRAY);
 	__uint(max_entries, 1);
 	__type(key, u32);
 	__type(value, struct skb_config);
 } skb_config_map SEC(".maps");
 
-/* Please keep the following structs in sync with its Rust counterpart in
- * module::skb::bpf.
- */
+const u8 __binding ifnamesz = 16;
```
atenart, line 36: Please add a comment on why we need to define this as a const variable instead of only using a define. Also we can link the two:

```c
#define IFNAMSIZ	16
const u8 __binding ifnamesz = IFNAMSIZ;
```

retis/src/bindings/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+pub(crate) mod kernel_upcall_ret_uapi;
+pub(crate) mod kernel_upcall_tp_uapi;
+use kernel_exec_tp_uapi::exec_ip;
+
+impl Default for exec_ip {
+    fn default() -> Self {
+        unsafe { mem::zeroed() }
+    }
+}
+
```
atenart, line 89: It seems bindgen can implement the `Default` trait by itself (see `nf_conn_ip`), so we probably do not need a special implementation anymore here.

retis/src/module/ovs/bpf/include/ovs_common.h
```diff
@@ -50,10 +49,10 @@ struct {
 
 /* Context saved between the begining and end of ovs_execute_actions calls. */
 struct execute_actions_ctx {
-	struct sk_buff *skb;
+	void *skb;
```
atenart, line 52: Please add a comment about this.

I'm not sure but doing the above might not help in catching casting issues. Not sure either if that is very likely. If we still want to support it, we could use `ifdef __BINDGEN__` for such cases. (Note that I haven't checked if such a thing is defined by bindgen or if we need to manually do it).

retis/src/bindings/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+use events_uapi::*;
+
+impl Default for common_task_event {
+    fn default() -> Self {
+        Self {
+            pid: 0,
+            comm: [0; retis_max_comm as usize],
+        }
+    }
+}
```
atenart, line 105: My understanding is bindgen can handle initializing long arrays; no need for a special implementation.

retis/src/core/events/bpf.rs
```diff
@@ -26,42 +28,49 @@ pub(super) const COMMON_SECTION_TASK: u64 = 1;
...
+}
 
-            fn to_string_opt(&self) -> Result<Option<String>> {
-                let res = self.to_string()?;
+/// A macro for converting a `c_char` array into a `String`.  It
+/// returns an error in case of failure. Upon successful conversion,
+/// it checks if the resulting `String` is empty, returning `Ok(None)`
+/// if it is.
+#[macro_export]
+macro_rules! raw_to_string_opt {
```
atenart, line 58: nit: as we have a single user, maybe we can open code this one? Feel free to ignore :)

retis/src/bindings/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+
+impl Default for retis_log_event {
```
atenart, line 107: Same here, bindgen can probably handle this.

atenart (changes requested):
Very nice!

I think we now should add a step in the CI to make sure no UAPI change in the BPF side got unnoticed and not propagated to bindgen Rust definitions.

nit: commit title "nft: declare and generate bindingsk" has a small typo.

Makefile
```diff
@@ -1,16 +1,36 @@
...
+
+BPF_CFLAGS := -target bpf \
+              -Wall \
+              -Wno-unused-value \
+              -Wno-pointer-sign \
+              -Wno-compare-distinct-pointer-types \
+              -fno-stack-protector \
+              -Werror \
+              -D__TARGET_ARCH_$(BPF_ARCH) \
+              -O2
```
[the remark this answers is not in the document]
vlrpl: given modules might face some changes, it is worth tackling this after (or in that) change.
All the `bpf/` dirs may be moved possibly changing/simplifying a bit the build logic.

Makefile
```diff
@@ -97,8 +143,12 @@ $(EBPF_PROBES) $(EBPF_HOOKS): $(LIBBPF_INCLUDES)
 	CFLAGS_INCLUDES="$(INCLUDES)" \
 	$(MAKE) -r -f $(ROOT_DIR)/ebpf.mk -C $@
 
+clean-bindings:
```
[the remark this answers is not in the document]
vlrpl: I intended this and `gen-bindings` as maintenance-only targets, but it still makes sense to clean before generating. Doing so though what's the value of a standalone `clean-bindings` target?

retis/src/module/skb/bpf/skb_hook.bpf.c
vlrpl: This slipped out. I didn't check the other patches, but I guess they follow the pattern you suggest, or at least they are supposed to.

Makefile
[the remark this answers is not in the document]
atenart: You're right!

Makefile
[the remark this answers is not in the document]
atenart: In this case I don't see a value for a standalone `clean-bindings`. But then if it's not in the help, that's fine, and having a dedicated target can help to split things a bit in the Makefile. Your choice :)

retis/src/module/skb/bpf/skb_hook.bpf.c
atenart: Yes, I saw later patches do follow the above.

tools/annotations.py
```diff
@@ -0,0 +1,42 @@
...
+            find_annotations(child, file_name)
+
+    find_annotations(translation_unit.cursor, file_path)
+    return annotated
+
+if __name__ == "__main__":
+    if len(sys.argv) < 3:
+        print(f"Usage: {sys.argv[0]} <header_file> <annotation> [<clang_args> ...]")
+        sys.exit(1)
+
```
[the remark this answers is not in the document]
vlrpl: not a python expert, but I'm not sure if I get it here on both.

If less than 3, it means only `header_file` is present and that is supposed to show the help. With `< 2` that will proceed and that is not what we want.
Also, slicing `sys.argv[3:]` returns an empty array in that case, which is fine as this will feed `index.parse()` with no options. Am I missing something?

Makefile
```diff
@@ -1,16 +1,36 @@
-ROOT_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+ROOT_DIR := $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
```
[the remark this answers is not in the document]
vlrpl: this one removes the `[...]/retis//retis/src/bindings` in the path (cosmetic). I put it here, but can be part of any PR/patch. Maybe I can move it in the outstanding Makefile fixes series?

The `out_console` is needed here because otherwise `gen-bindings V=1` will fail. 
Maybe there's a better way, but my Makefile-fu didn't suggest anything better :)

Makefile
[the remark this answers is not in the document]
vlrpl: ack, let's go with the undocumented target

retis/src/module/ovs/bpf/include/ovs_common.h
vlrpl: added `-D__BINDGEN__` to the invocation and kept `struct sk_buff` for the other case

vlrpl:
> Very nice!
> 

thanks for taking a look.

> I think we now should add a step in the CI to make sure no UAPI change in the BPF side got unnoticed and not propagated to bindgen Rust definitions.
> 

The reason I didn't include it (at least so far) was mostly because different versions of bindgen may result in [slightly] different code generation. This would force us to generate the bindings following the same version the CI uses (or check with the same version we used to generate them). The latter is probably better, but I'm unsure how straightforward/convenient this is.

> nit: commit title "nft: declare and generate bindingsk" has a small typo.

retis/src/module/ovs/bpf/include/ovs_common.h
atenart: Maybe an even better way would be to hide the `__BINDGEN__` use into a macro. Eg.

```c
BINDGEN_PTR(struct sk_buff *, skb)

// Without __BINDGEN__ defined,
struct sk_buff *skb

// With __BINDGEN__ defined,
void *skb
```

That makes the switch automatic for all pointers and adds a clear API when dealing with pointers in the events.

atenart:
> > I think we now should add a step in the CI to make sure no UAPI change in the BPF side got unnoticed and not propagated to bindgen Rust definitions.
> 
> The reason I didn't include it (at least so far) was mostly because different versions of bindgen may result in [slightly] different code generation. This would force us to generate the bindings following the same version the CI uses (or check with the same version we used to generate them). The latter is probably better, but I'm unsure how straightforward/convenient this is.

I see. Maybe forcing to update the bindings in such case would be OK though. It's easy to do when the above happens and helps preventing issues in the UAPI?

retis/src/module/ovs/bpf/include/ovs_common.h
vlrpl: what about a more descriptive name? E.g. `BINDING_SKIP_PTR()` ?

vlrpl:
> > > I think we now should add a step in the CI to make sure no UAPI change in the BPF side got unnoticed and not propagated to bindgen Rust definitions.
> > 
> > 
> > The reason I didn't include it (at least so far) was mostly because different versions of bindgen may result in [slightly] different code generation. This would force us to generate the bindings following the same version the CI uses (or check with the same version we used to generate them). The latter is probably better, but I'm unsure how straightforward/convenient this is.
> 
> I see. Maybe forcing to update the bindings in such case would be OK though. It's easy to do when the above happens and helps preventing issues in the UAPI?

I'm not sure. Our bindings are fairly simple and should be stable. Updating once in a while should be ok (maybe using some stable version?).

If the container (current or future) ships a different version (and no real changes are present) we'd end up failing because of:

```diff
-/* automatically generated by rust-bindgen 0.69.4 */
+/* automatically generated by rust-bindgen 0.70.1 */

```

and every time the version changes this will happen. Of course, we can exclude such a false positive, but I would avoid chasing this kind of CI breakage because of potential false positive (like a typo, or different wording in the comment emitted by bindgen :) unless this is really beneficial (actual bugs are prevented/fixed).

I guess we could use the `+stable` toolchain (assuming `bindgen-cli` version follows `+stable`), but `+stable` has to be part of `rust:bookworm` (I don't know if it's the case currently), otherwise we should install it manually and then use it for our `gen-binding` target.

retis/src/module/ovs/bpf/include/ovs_common.h
atenart: Yes, the name can be changed. However I'm not sure about "skip", that's describing the internals of how this trick works, not what is provided (here, the ability to define pointers in the UAPI).

atenart:
OK, let's no do this for now. The version strings is not ideal as you noted.

retis/src/module/ovs/bpf/include/ovs_common.h
vlrpl: `BINDING_PTR()` sounds better

vlrpl:
no further updated planned for this PR

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -49,7 +49,7 @@ struct tracking_info {
 	u64 last_seen;
 	/* Original head address; useful when the head is invalidated */
 	u64 orig_head;
-};
+} __binding;
```
atenart, line 52: The comment above the definition "Please keep in sync with its Rust counterpart..." can be removed. There are many other examples.

retis/src/module/nft/bpf/nft.bpf.c
```diff
@@ -17,28 +18,30 @@
...
+/** 
+ * Nft hook configuration.
+ *
```
atenart, line 23: There are extra spaces at EOL in this file.

retis/src/module/nft/bpf/nft.bpf.c
```diff
@@ -7,6 +7,7 @@
 #define VERD_SCALE (NFT_RETURN * -1)
 #define ALLOWED_VERDICTS(verd, mask) (1 << (verd + VERD_SCALE) & mask)
 #define NFT_NAME_SIZE 128
+__binding const u32 nft_name_size = NFT_NAME_SIZE;
```
atenart, line 10: Please add a comment here on why we need this construction, like the other. Alternatively we could use a `BINDING_DEF` macro that does the trick somehow.

Also, it's not used, do we need it in the UAPI? Only `log_events_max` is used (the macro could still be nice here). It seems array sizes are not useful to be shared in the UAPI.

retis/src/core/probe/kernel/kernel.rs
```diff
@@ -158,18 +151,18 @@ impl RawEventSectionFactory for KernelEventFactory {
...
-    impl RawSectionBuilder for RawKernelEvent {
+    impl RawSectionBuilder for kernel_event {
         fn build_raw(out: &mut Vec<u8>) -> Result<()> {
-            let data = RawKernelEvent {
+            let data = Self {
                 symbol: Symbol::from_name("openvswitch:ovs_do_execute_action")?.addr()?,
-                r#type: 2, // Raw tracepoint.
-                stack_id: u64::MAX,
+                type_: 2, // Raw tracepoint.
+                stack_id: u64::MAX as i64,
```
atenart, line 165: I think we should just use `-1` here as this is the "no stack available" value. This doesn't change runtime behavior as the stack handling is gated for tests.

retis/src/core/events/bpf/include/events.h
```diff
@@ -10,14 +10,19 @@
 #define RAW_EVENT_DATA_SIZE	1024 - 2 /* Remove the size field */
 #define RETIS_MAX_COMM		64
 
+__binding const int retis_max_comm = RETIS_MAX_COMM;
```
atenart, line 13: Ditto.

retis/src/core/events/bpf/include/events.h
```diff
@@ -10,14 +10,19 @@
...
+
 /* Please keep the below in sync with its Rust counterpart. */
 #define LOG_MAX			127
 #define LOG_EVENTS_MAX		32
 
+__binding const int log_max = LOG_MAX;
+__binding const int log_events_max = LOG_EVENTS_MAX;
```
atenart, line 20: Ditto.

atenart (changes requested):
This looks really nice. Only a few minor comments; once resolved IMO this can be merged.

Looks like some parts are not converted yet (eg. meta-filtering). I guess that will come at a later time? (This is fine).

retis/src/module/nft/bpf/nft.bpf.c
atenart: >  Alternatively we could use a BINDING_DEF macro that does the trick somehow.

I'm thinking something like (highly untested):

```c
#define BINDING_DEF(def, val) enum enum_ ## def { def = val } __binding
```

atenart:
Actually just found an issue which is not in `main`. Loading all (but ovs) collectors, I get the following error now:

```
ERROR [eBPF] Failed to get event section: no space left
```

retis/src/module/nft/bpf/nft.bpf.c
vlrpl: > Please add a comment here on why we need this construction, like the other. Alternatively we could use a `BINDING_DEF` macro that does the trick somehow.
> 
> Also, it's not used, do we need it in the UAPI? Only `log_events_max` is used (the macro could still be nice here). It seems array sizes are not useful to be shared in the UAPI.

I guess this was used when there was the default trait implementation.
If it's not used anymore, it's better to remove it.
No need to carry unneeded things

retis/src/module/nft/bpf/nft.bpf.c
vlrpl: this was needed to instruct bindgen with an "annotating" comment/doc.
Must be converted back to C style.

vlrpl:
> Actually just found an issue which is not in `main`. Loading all (but ovs) collectors, I get the following error now:
> 
> ```
> ERROR [eBPF] Failed to get event section: no space left
> ```

the message is correct, and this happens because we hit event data exhaustion.
There's a way to solve this without increasing the data size, but this will be unavoidable if we add more sections.

retis/src/module/nft/bpf/nft.bpf.c
vlrpl: ok, the trailing whitespaces need to be removed. The comment has to stay instead. Plain C-style doesn't make bindgen pick this annotation up.

vlrpl:
Pushed for a run of CI.
Some final work is required for the new alignment part.

retis/src/module/nft/bpf/nft.bpf.c
atenart: The above could still be used for `log_events_max`.

retis/src/core/events/bpf.rs
```diff
@@ -487,39 +484,20 @@ impl BpfEventsFactory {
...
-unsafe impl Plain for LogEvent {}
-
 /// Max number of events we can store at once in the shared map. Please keep in
 /// sync with its BPF counterpart.
 pub(super) const BPF_EVENTS_MAX: u32 = 8 * 1024;
 
 /// Size of the raw data buffer of a BPF event. Please keep synced with its BPF
 /// counterpart.
-pub(crate) const BPF_RAW_EVENT_DATA_SIZE: usize = 1024 - 2 /* remove the size field */;
+pub(crate) const BPF_RAW_EVENT_DATA_SIZE: usize = 1024 - 8 /* remove the size and pad fields */;
```
atenart, line 493: Maybe we can use `mem::offset_of!` instead to improve maintenance?

retis/src/core/events/bpf/include/events.h
```diff
@@ -7,17 +7,23 @@
 
 /* Please keep the below in sync with its Rust counterpart. */
 #define EVENTS_MAX		8 * 1024
-#define RAW_EVENT_DATA_SIZE	1024 - 2 /* Remove the size field */
+#define RAW_EVENT_DATA_SIZE	1024 - 8 /* Remove the size and pad fields */
```
atenart, line 10: We could introduce an `offsetof` macro and use it here for better maintenance?

retis/src/module/nft/bpf/nft.bpf.c
vlrpl: my plan was to find a nice way to avoid the usage of types bigger than 8 bytes for alignment requirements.
This should happen through static assertion at compile time, meaning we might end up wrapping event-binding declarations as well.
If you agree, we can add this wrapper and the above in a follow-up as they are both not blockers for the time being.

Makefile
```diff
@@ -168,6 +220,8 @@ help:
 	$(call help_once,NOVENDOR            --  Avoid to self detect and consume the vendored headers)
 	$(call help_once,                        shipped with libbpf-sys.)
 	$(call help_once,RA                  --  Applies to check and clippy and runs those targets with the options needed)
-	$(call help_once,                        for rust-analyzer. When $$(RA) is used, $$(V) becomes ineffective.)
+	$(call help_once,                        for rust-analyzer. When $$(RA) is used $$(V) becomes ineffective.)
```
atenart, line 223: `rust-analyzer` is still here, but not sure if this is intended. (it's probably OK but just want to make sure).

retis/src/core/events/bpf.rs
```diff
@@ -333,7 +342,8 @@ pub(crate) fn parse_raw_event<'a>(
         cursor += mem::size_of_val(&raw_section.header);
 
         // Compute where the current section ends.
-        let raw_section_end = cursor + raw_section.header.size as usize;
+        let raw_section_end =
+            cursor + raw_section.header.size as usize - raw_section.header.pad_size as usize;
```
atenart, line 346: Removing the pad size if fine when setting up the raw section Rust representation, but the below checks against the real size of the event (pad included):

```rust
        } else if raw_section_end > raw_event_size {                            
            error!("Section goes past the buffer: {raw_section_end} > {raw_event_size}");
            break;                                                              
        }
```

retis/src/core/events/bpf/include/events.h
```diff
@@ -35,15 +41,22 @@ enum retis_event_owners {
...
 
 /* Please keep synced with its Rust counterpart. */
 struct retis_raw_event_section_header {
 	u8 owner;
 	u8 data_type;
 	u16 size;
-} __attribute__((packed));
+	u8 pad_size;
+	u8 pad[3];
+} __binding;
```
atenart, line 57: Why not also converting the other event structs? This makes things to be in between 2 states which is not great for maintenance.

atenart (changes requested):
The bindgen part looks very nice and ready. The padding one is a great addition but here are some comments and I'd like to have another look at it. If you want we can split them into two PRs to ease things.

Makefile
vlrpl: yes, this is intended (not a target)

retis/src/core/events/bpf/include/events.h
vlrpl: admittedly, yes, I had the same thought, but this PR has the potential to become huge (more than it is already), that's why I left it out to be converted in a follow-up PR.  Let's just convert RawEvent as well.

retis/src/module/ct/bpf.rs
```diff
@@ -192,18 +135,18 @@ impl CtEventFactory {
             };
             (orig, reply)
         } else if flags & RETIS_CT_IPV6 != 0 {
-            let s = unsafe { raw.orig.src.addr.ipv6 };
-            let d = unsafe { raw.orig.dst.addr.ipv6 };
+            let s = unsafe { std::mem::transmute::<[u32; 4], [u8; 16]>(raw.orig.src.addr.ipv6) };
+            let d = unsafe { std::mem::transmute::<[u32; 4], [u8; 16]>(raw.orig.dst.addr.ipv6) };
```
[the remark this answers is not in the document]
vlrpl: normally there are unions with 4, 2, 1B (I guess ipv6 is just that). I simply picked 4.

retis/src/core/events/bpf/include/events.h
vlrpl: not sure `offsetof` can be used here as we need to define this macro getting the offset of a member using the macro in the definition :)

retis/src/core/events/bpf.rs
vlrpl: let's make this a binding and zero this maintenance :)

retis/src/module/ct/bpf.rs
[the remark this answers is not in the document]
atenart: Not sure if I got your comment right, so to be sure: my point is if the IPv6 address is stored directly in a `u8 v6[16]` in the event section, there's no need for transmutation.

(I guess moving the IPv4 to an array would help doing the same thing too).

retis/src/core/events/bpf/include/events.h
atenart: This should also help in not having to explicitly populate the `pad` fields in the Rust side.

retis/src/module/ct/bpf.rs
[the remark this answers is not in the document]
vlrpl: I was simply confirming that I picked one of the layouts of the C source union.
Converted to u8, though.

vlrpl:
I went for the split, the risk of conflicts is fairly high with this.

retis/src/core/events/bpf/include/events.h
vlrpl: a quick check confirm this is a problem for the reason I explained above

retis/src/core/events/bpf.rs
vlrpl: good catch. I overlooked this for some reason. 
It's fairly easy to fix this (also most probably this could happen in synthetic tests only), but regardless, we're very close to the event limit and the padding strategy (admittedly conservative on the correctness side) may lead to overflow. We should consider increasing the event size, reducing the modules' requirement (e.g. reducing the nft chain buffer size), optimizing the alignment strategy, or completely changing the raw events (the best option long term, but I don't consider this a priority now).

retis/src/module/ct/bpf.rs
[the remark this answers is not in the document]
atenart: Ah got it now :)

retis/src/core/events/bpf.rs
vlrpl: Either way, this got split because of the overflow risk, so it's no longer part of this series.

retis/src/core/events/bpf/include/events.h
atenart: Thanks for checking!

retis/src/core/events/bpf.rs
atenart: Right, let's tackle the alignment in a separate PR. I believe it's close to being ready (we can increase the buffer size for the above issue that's fine).

atenart (approved):
Very happy to see this, thanks for the hard work!

PR #422, opened 2024-08-13
Fix loading issue with meta filtering on older kernels

vlrpl:


retis/src/core/filters/meta/bpf/include/meta_filter.h
```diff
@@ -176,11 +177,8 @@ bool cmp_bytes(struct retis_meta_ctx *ctx)
 	if (bpf_probe_read_kernel_str(val, sz, (char *)ctx->base + ctx->offset) < 0)
 		return 0;
 
-	if (!sz)
-		return false;
-
 	const char *sp1 = ctx->data, *sp2 = val;
-	while (sz-- > 0 && *sp1 && *sp2 && !(*sp1 - *sp2))
+	while (sz-- > 0 && (ret = !(*sp1 - *sp2)) && *sp1 && *sp2)
```
atenart, line 181: I don't get why using an extra variable (`ret`) here and only here makes the verifier happy. Is that intended?

vlrpl: looks misleading because a hunk was missing (`return ret` remained out from my refresh).
Still that was not enough, but this time with recent verifiers (`-O1` fixed this new issue).
I rearranged a bit the code in a simpler way and it now seems to work properly on both older and recent verifiers.

atenart (approved):
Looks good, thanks for fixing this!

PR #424, opened 2024-08-29
Add tests for skb device name and skb inum

sundaram123krishnan:
Implemented tests for skb device name and skb inum


vlrpl (commented):
@sundaram123krishnan, the patch looks good.
I have only a couple of minor suggestions (see inline).

tests/test_filter.py
```diff
@@ -0,0 +1,97 @@
+from testlib import Retis, assert_events_present
+
+
+def test_meta_filter_string(two_ns_simple):
+    ns = two_ns_simple
+    retis = Retis()
+    retis.collect(
+        "-c", "skb", "-m", "sk_buff.dev.name == 'veth01'", "-p", "tp:net:netif_rx"
+    )
+    print(ns.run("ns0", "ping", "-c", "1", "10.0.42.2"))
```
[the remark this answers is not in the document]
sundaram123krishnan: Will prefer to open a separate PR for this task

tests/test_filter.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+    retis = Retis()
+    retis.collect(
+        "-c",
+        "skb",
+        "-f",
+        "icmp and host 10.0.42.2",
+        "-m",
+        f"sk_buff.dev.nd_net.net.ns.inum == {inum}",
```
atenart, line 60: The leading `f` should be removed.

atenart:
In addition to the above comment, the "Signed-off-by" line in the commit descriptions should be set after an empty newline, otherwise it is shown as part of the commit title. (You can also squash both commits if you want).

tests/test_filter.py
```diff
@@ -0,0 +1,79 @@
...
+        "-p",
+        "tp:skb:kfree_skb",
```
atenart, line 62: Can we use `tp:net:netif_rx` here as well? The ICMP packet is not expected to be dropped.

atenart (approved):
I manually tested this, changes LGTM.

Thanks for your contribution, taking into account comments and for your patience!

sundaram123krishnan:
Thx @atenart 
Will open another PR regarding adding time constraints to the `ping` in the tests.

PR #427, opened 2024-09-26
[RFC] Introduce a common types factory

atenart:
The goal is to provide a common way of reporting common type values as part of the events, to help introspection. The drop reasons are converted over this and the reset reason are added. Other types can be added later. Based on #421.

Copying the ongoing discussion from #421:

> > > I think this might be problematic. Not all functions (and therefore not all probes) that have `struct skb_event_reason` as argument are actually dropping the packet.
> > 
> > Right.
> > > We could very easily end up having duplicate drop events that would be terribly confusing. I do think `DropEvent` must come _only_ from `kfree_skb`. Am I missing a use-case?
> > > I can see some kind of generic "tracking type" or "type-driven events" for some interesting types but we should be very clear that it just means "this probe happened to have this type and value in its arguments". IMO, A DropEvent has a very different meaning.
> > 
> > Drops are quite complex. Bottom line is drop reasons are not linked to a packet effectively being dropped: eg. the kernel could be too old (in case we emulate a reason, but I think we should stop doing so) or the reason might be `SKB_COMSUMED` or `SKB_NOT_DROPPED_YET`. Detecting a packet is being dropped is really about seeing it in the right tracepoint: `skb:kfree_skb`; that's all[1]. The `skb-drop` section was just here to hold drop-related information, not that the packet was effectively dropped.
> > Now, I agree this is confusing and not a great UX. The naming of `skb-drop` was pretty bad.
> > Then on why retrieving those two "common types". More and more functions in the kernel accept those "reason" enums as parameters and are even returning them. Retrieving those values gives very valuable information, as we can understand better the flow of a packet, and how a function was behaving.
> 
> Can you give a concrete example?

[tcp_rcv_state_process](https://elixir.bootlin.com/linux/v6.11/C/ident/tcp_rcv_state_process), [tcp_sequence](https://elixir.bootlin.com/linux/v6.11/C/ident/tcp_sequence) and [tcp_child_process](https://elixir.bootlin.com/linux/v6.11/C/ident/tcp_child_process) return a reason (there are others). You can also get to `kfree_skb_reason` (the function, not the `skb:kfree_skb` tp) with a `SKB_CONSUMED` reason (in which case it's not a drop).

But you're also right, the reason (being reset or drop) will end up in the tracepoint where the action is actually performed. That's maybe the real value here.

> > Finally (I think that's important for finding a solution) I wasn't sure about how to model those "common types" in the events. I chose a single section per type (and reused `skb-drop` for this), but that might not be the perfect solution.
> > I think we should improve both situations.
> > One possibility: remove the `skb-drop` section and provide a `drop()` helper for events. A bit like the `tracking_id()` helper in the tracking section. That new helper could be used to flag a packet being dropped in the output, and later in Python. It would just be hidden from the raw JSON event (IMO that is fine). In addition the drop reason should be stored in a new section. The `skb-drop` collector can stay, it has value in probing at the right place.
> 
> Maybe that's an option but my feeling is that "packets being dropped" is way more important than "functions happen to pass or return some drop reason". I think the UX should strongly favor the former and try by all means not to get them mixed.
> 
> How about:
> 
>     * We keep the `DropEvent` linked uniquely to kfree_skb. We make it clear in the string representation that there was a packet drop.
> 
>     * We add a `SkResetEvent` linked uniquely to trace. We make it clear in the string representation that a rst was actually sent.
> 
>     * We have a `ProbeArgsEvent` or `ProbeMetadata` or `ProbeInfo` (something like that) that gives information about the probe itself, its arguments and return code for known types. We could support native C types and some well-known structs or enums. Here we can add skb_drop_reason but the string representation should be super-clear this is an argument of the function, e.g:
> 
> 
> ```
>  [k] [tcp_rcv_state_process] skb ffff8e8ec1ab3300
>    probe info: arg0(struct sock*) = 0x123445, arg1(struct skb*) = 0xffff8e8ec1ab3300, ret(enum skb_drop_reason) = SKB_DROP_REASON_TCP_INVALID_SEQUENCE
>    192.168.1.137 > 192.168.1.138 ttl 64 tos 0x0 id 49968 off ... rest of skb section ...
>    [... other sections ...]
> 
>  [k] [kfree_skb_reason] skb ffff8e8ec1ab3300
>    192.168.1.137 > 192.168.1.138 ttl 64 tos 0x0 id 49968 off ... rest of skb section
>     drop (reason TCP_OLD_SEQUENCE) 
> ```

I'll think a bit more about it, but I think I like that idea. By default get sections for drop / reset linked to the actual action that happened and convert the "common type" logic to retrieve all parameters and return values. That would be off by default. We can probably remove the factory from `skb-drop` and reuse the same (but producing an `skb-drop` section, so we can reuse the enum parsing code).


amorenoz:
> I'll think a bit more about it, but I think I like that idea. By default get sections for drop / reset linked to the actual action that happened and convert the "common type" logic to retrieve all parameters and return values. That would be off by default. We can probably remove the factory from skb-drop and reuse the same (but producing an skb-drop section, so we can reuse the enum parsing code).

+1 to having a common, shared enum value extractor.

atenart:
> > I'll think a bit more about it, but I think I like that idea. By default get sections for drop / reset linked to the actual action that happened and convert the "common type" logic to retrieve all parameters and return values. That would be off by default. We can probably remove the factory from skb-drop and reuse the same (but producing an skb-drop section, so we can reuse the enum parsing code).
> 
> +1 to having a common, shared enum value extractor (not sure if reusing the word "Factory" is to confusing lol)

Yeah, I initially thought of keeping the common factory introduced here; but that does not make sense. Common helpers do.

atenart:
Closing this for now.

PR #429, opened 2024-09-27
Better handle packets we don't recognize

### LINKED ISSUE #428: Properly handle  non-Ethernet frames and not-supported ethertypes
We've seen non-Ethernet packets in retis:

e.g:
```
48678778241154 (0) [irq/175-iwlwifi] 1669 [tp] skb:kfree_skb drop (reason mac80211_monitor/RX_DROP_MONITOR)                                                                                                         
    bpf_prog_be31ae23198a0378_sd_devices+0x3f744                                                                                                                                                                     
    bpf_prog_be31ae23198a0378_sd_devices+0x3f744                                                                                                                                                                     
    bpf_trace_run3+0x8b                                                                                                                                                                                              
    kfree_skb_reason+0x8b                                                                                                                                                                                            
    ieee80211_rx_handlers+0x2b8                                                                                                                                                                                      
    ieee80211_prepare_and_rx_handle+0x862                                                                                                                                                                            
    ieee80211_rx_list+0x9b3                                                                                                                                                                                          
    ieee80211_rx_napi+0x51                                                                                                                                                                                           
    iwl_mvm_rx_mpdu_mq+0xb08                                                                                                                                                                                         
    iwl_pcie_rx_handle+0x260                                                                                                                                                                                         
    iwl_pcie_napi_poll_msix+0x39                                                                                                                                                                                     
    __napi_poll+0x28                                                                                                                                                                                                 
    net_rx_action+0x2c6                                                                                                                                                                                              
    handle_softirqs+0xf4                                                                                                                                                                                             
    do_softirq.part.0+0x3b                                                                                                                                                                                           
    __local_bh_enable_ip+0x60                                                                                                                                                                                        
    iwl_pcie_irq_rx_msix_handler+0xd1                                                                                                                                                                                
    irq_thread_fn+0x20                                                                                                                                                                                               
    irq_thread+0x1b0                                                                                                                                                                                                 
    kthread+0xcf                                                                                                                                                                                                     
    ret_from_fork+0x31                                                                                                                                                                                               
    ret_from_fork_asm+0x1a                                 
```

Also, it's possible  that the packet has a valid Ethernet header but  the `ethertype is not one of the few we support`.

On both cases, the output  is  weird because the skb module will not print anything as reported [here](https://github.com/retis-org/retis/pull/411#issuecomment-2372280724).

We should:
- try better to detect  non-Ethernet frames in the kernel
-  handle  both cases better in rust  when printing.

atenart:
Fixes #428.

### DIFF
--- a/retis-events/src/display.rs
+++ b/retis-events/src/display.rs
@@ -313,4 +313,9 @@ impl DelimWriter {
         }
         Ok(())
     }
+
+    /// Was the DelimWriter used?
+    pub fn used(&self) -> bool {
+        !self.first
+    }
 }
--- a/retis-events/src/net.rs
+++ b/retis-events/src/net.rs
@@ -6,7 +6,7 @@ use base64::{
 };
 
 /// Returns a translation of some ethertypes into a readable format.
-pub(crate) fn etype_str(etype: u16) -> Option<&'static str> {
+pub fn etype_str(etype: u16) -> Option<&'static str> {
     Some(match etype {
         0x0800 => "IPv4",
         0x0806 => "ARP",
--- a/retis-events/src/skb.rs
+++ b/retis-events/src/skb.rs
@@ -285,6 +285,13 @@ impl EventFmt for SkbEvent {
             write!(f, "size {}]", gso.size)?;
         }
 
+        // If we didn't print any section, it means the section has raw packet
+        // data but we were unable to decode it. Print something.
+        if !space.used() {
+            space.write(f)?;
+            write!(f, "unknown packet (see raw data)")?;
+        }
+
         Ok(())
     }
 }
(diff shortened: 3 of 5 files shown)


vlrpl:
the change looks good to me. Just a tiny nit on patch 3/4 commit message.
I think `$subject ` is a placeholder, and while at it, the last line `(only looking like a valid ethertype, but what can we do?` has unmatching parentheses.

atenart:
> the change looks good to me. Just a tiny nit on patch 3/4 commit message. I think `$subject ` is a placeholder, and while at it, the last line `(only looking like a valid ethertype, but what can we do?` has unmatching parentheses.

`$subject` was intentional because I was a bit lazy I guess to repeat the title :) But you're right; I fixed the two above issues.

vlrpl (approved):
lgtm, thanks!

PR #430, opened 2024-10-04
Makefile fixes

vlrpl:
- fixes $(eval)
- introduces `rust-analyzer` target
- fixes `make help` target


Makefile
```diff
@@ -97,6 +97,9 @@ $(EBPF_PROBES) $(EBPF_HOOKS): $(LIBBPF_INCLUDES)
 	CFLAGS_INCLUDES="$(INCLUDES)" \
 	$(MAKE) -r -f $(ROOT_DIR)/ebpf.mk -C $@
 
+rust-analyzer:
+	$(CARGO) check --quiet --message-format=json --all-targets --keep-going
```
amorenoz, line 105: `rust-analyzer` supports running `clippy` instead of `check` for more precise diagnostics, could we support it as well?

We could have another target `rust-analyzer-clippy` but that's ugly....
Maybe a single `cargo` target that takes the command and arguments from env and a wrapper script in `tools` that adjusts arguments?

Or maybe a wrapper target in pure Makefile:

```makefile
rust-analyzer:
	$(eval PRINT=echo >/dev/null)
	$(eval CARGO_CMD_OPTS=--quiet --message-format=json --all-targets --keep-going)
	[ $$(echo $(MAKECMDGOALS) | wc -w) -eq 1 ] && echo "Specify a command to run: check | clippy" || true

check:
	$(call build, check, checking)

clippy:
	$(call build, clippy, running clippy)

```

This has the benefit of giving us `make check` and `make clippy` for day-to-day development.

amorenoz: Tried this and it works nicely with my IDE. LGTM

amorenoz:
Can we also remove CFLAGS from the exported  variables, since we don't use it right now and it triggers a rebuild of `vprintf`, `libbpf-rs`, et.al?

amorenoz (approved):
Tested and confirmed that: rust analyzer (using both `check` and `clippy` targets with `RA=1`) does not trigger additional rebuilds even if `make` is run in the meantime.

PR #431, opened 2024-10-16
Containerfile and workflow improvements

### LINKED ISSUE #291: Provide a next version of the image
Automatically build a `next` version of the container image based on the `main` branch whenever a change is merged. This could be done in quay.io directly, or using GH actions.

Note that for this to work properly, `retis-derive` should be released in its own crate, which is one of the goals of #282.

vlrpl:
- Modifies `Containerfile` keeping in mind the workspace structure
  - while at it, included some clean-ups
- Introduces the creation of a new image based on the current main
  - scheduled every 1st of the month
  - can be triggered manually
- Simplifies the release process
  - RELEASE.md was updated accordingly
- Tokens have been added to the repo settings

Workflow complete run:
- [manual](https://github.com/vlrpl/retis/actions/runs/11386275978/job/31678089186)
  - nexttest latest-nexttest
- [scheduled](https://github.com/vlrpl/retis/actions/runs/11384264586/job/31671597581) 
  - next

New images generated with the workflow (will be deleted):
https://quay.io/repository/retis/retis?tab=tags

Fixes #291

### DIFF
--- a/.github/workflows/build_push_image.yaml
+++ b/.github/workflows/build_push_image.yaml
@@ -1,0 +1,66 @@
+name: Build and push container image
+
+on:
+  # Manual trigger for the workflow.
+  workflow_dispatch:
+    inputs:
+      release_tags:
+        required: false
+        default: 'next'
+
+  # Every day at 01:32 AM UTC.
+  schedule:
+    - cron: "32 01 * * *"
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      # This is kept intentionally separate from "Set environment".
+      - name: Check scheduled run
+        if: github.event_name == 'schedule'
+        run: |
+          last_commit=$(git log -n1 --since=yesterday --oneline)
+          if [ -z "${last_commit}" ]; then
+              echo "No new commits found, cancelling ..."
+              exit 1
+          else
+              echo "Proceeding with the current tip:"
+              echo "  ${last_commit}"
+          fi
+
+      - name: Set environment
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+              echo "RELEASE_TAGS=${{ github.event.inputs.release_tags }}" >> $GITHUB_ENV
+          elif [ "${{ github.event_name }}" = "schedule" ]; then
+              echo "RELEASE_TAGS=next" >> $GITHUB_ENV
+          elif [ "${{ github.event_name }}" = "release" ]; then
+              :
+          fi
+
+      - name: Build container image
+        id: build_image
+        uses: redhat-actions/buildah-build@v2
+        with:
+          containerfiles: ./Containerfile
+          image: retis
+          tags: ${{ env.RELEASE_TAGS }}
+
+      - name: Push container image
+        id: push_image
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ${{ steps.build_image.outputs.image }}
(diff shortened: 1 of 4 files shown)


vlrpl:
I performed some tests for scheduled runs that required some push. Now it's ready for review.

atenart (changes requested):

RELEASE.md
```diff
@@ -18,10 +18,8 @@
          not that can be set in the project settings.
       1. `Packages > Rebuild`.
    1. Build and push the container image.
-      1. `$ buildah build -t quay.io/retis/retis:x.y.z`
-      1. `$ buildah push quay.io/retis/retis:x.y.z`
-      1. Manually tag on the web UI the image pushed with `latest`, if
-         applicable.
+      1. Manually run the workflow `build-and-push-image` selecting the branch and setting the
+         `release_tags` with the space separated list of tags (i.e. release_tags="x.y.z latest").
```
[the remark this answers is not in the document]
vlrpl: > The current process is to build an image tagged `x.y.z` and then in the web UI to make a symlink for the `latest` tag. Wouldn't this trigger two identical builds?
> 

AFAIU, it builds once and tags it multiple times:
```
Successfully tagged localhost/retis:next
771af2b19e6e100c8ea7c6debfe675485d18aa8b63e969cc7f405b87f763b1dc
Tagging the built image with tags next,latest-next
```

Unfortunately, the scheduled run overwrote the next image, so the alias is not visible, but the history on the registry normally shows the same pattern we have with latest <-> 1.4.0

> On the other hand it makes updating the `latest` tag automatic.
> 

That was the idea explicitly supporting multiple tags.

> Also can we be a little bit more explicit on where to launch the workflow? (eg adding "in the Actions tab on the Retis Github page").

I think this will be the link, I'll add that as well
[https://github.com/retis-org/retis/actions/workflows/build_push_image.yaml](https://github.com/retis-org/retis/actions/workflows/build_push_image.yaml)

.github/workflows/build_push_image.yaml
```diff
@@ -0,0 +1,71 @@
...
+          elif [ "${{ github.event_name }}" = "release" ]; then
+              :
+          fi
+
+      - name: Log in to the registry
+        uses: redhat-actions/podman-login@v1
+        with:
+          registry: quay.io/retis
+          username: ${{ secrets.REGISTRY_USERNAME }}
+          password: ${{ secrets.REGISTRY_TOKEN }}
```
atenart, line 51: Sorry I only see this now, I think this can be done at the `redhat-actions/push-to-registry@v2` stage by adding `username` and `password` under `with:`.

vlrpl: yes, it could, but I split to make it bail it early (without building) in case of issues (network failures and so forth).
Do you prefer to squash this in `push-to-registry`

atenart: > yes, it could, but I split it to make it bail it early (without building) in case of issues (network failures and so forth).

Makes sense.

> Do you prefer to squash this in `push-to-registry`?

The advantage is building stages are not logged in; not sure if that is really better. So up to you :)

vlrpl: had no strong opinion on it, eventually, I went with your suggestion.
I triggered a manual build on my repo to check the upload goes fine, just in case.
I'll put the link here once the job's done.

vlrpl: Job successfully ran to completion:

https://github.com/vlrpl/retis/actions/runs/11387554527/job/31682439658

atenart: Thanks!

atenart (approved):
Nice additions!

PR #434, opened 2024-10-18
ct: add support for mark and labels

### LINKED ISSUE #407: [feature] Print out ct mark and labels
`conntrack -L` prints out mark and labels, but `retis` doesn't.  This information is important for debugging OVN issues, since OVN is storing a lot of information in there like MAC addresses and ports for ECMP symmetric replies, flags for connections blocked by ACLs and so on.

atenart:
RFC as I don't want this to be merged before the bindgen PR (#419); this will need to be rebased.

A test image is available and can be used as follow:

```
$ RETIS_TAG=ct-mark-label ./retis_in_container.sh collect -c ct ...
```

Labels can be set for testing, eg:

```
$ iptables -A INPUT -m connlabel --label 42 --set
```

A few questions / remarks:
- I took the opportunity to make the parent ct information to be printed on a newline. Is that understandable?
- Both the mark and labels are retrieved for the base *and* the parent `nf_conn`.
- Labels are a bitmask; I chose to print them as an hex value to save some space. Is that fine?

@igsilya can you help testing to make sure this looks good? Thanks!

Close #407.

### DIFF
--- a/retis-events/src/ct.rs
+++ b/retis-events/src/ct.rs
@@ -1,6 +1,6 @@
 use std::fmt;
 
-use super::*;
+use super::{helpers::U128, *};
 use crate::{event_section, event_type, Formatter};
 
 #[event_type]
@@ -127,6 +127,8 @@ pub struct CtConnEvent {
     pub tcp_state: Option<String>,
     /// Connection mark tracking label
     pub mark: Option<u32>,
+    /// Connection tracking labels.
+    pub labels: Option<U128>,
 }
 
 impl EventFmt for CtEvent {
@@ -210,6 +212,10 @@ impl CtEvent {
 
         if let Some(mark) = conn.mark {
             write!(f, " mark {}", mark)?;
+        }
+
+        if let Some(labels) = &conn.labels {
+            write!(f, " labels {:#x}", labels.bits())?;
         }
 
         Ok(())
--- a/retis-events/src/ct.rs
+++ b/retis-events/src/ct.rs
@@ -125,6 +125,8 @@ pub struct CtConnEvent {
     pub reply: CtTuple,
     /// TCP state; if any
     pub tcp_state: Option<String>,
+    /// Connection mark tracking label
+    pub mark: Option<u32>,
 }
 
 impl EventFmt for CtEvent {
@@ -204,6 +206,10 @@ impl CtEvent {
             ZoneDir::Reply => write!(f, "reply-zone {}", conn.zone_id)?,
             ZoneDir::Default => write!(f, "zone {}", conn.zone_id)?,
             ZoneDir::None => (),
+        }
+
+        if let Some(mark) = conn.mark {
+            write!(f, " mark {}", mark)?;
         }
 
         Ok(())
--- a/retis-events/src/ct.rs
+++ b/retis-events/src/ct.rs
@@ -142,7 +142,7 @@ impl EventFmt for CtEvent {
         Self::format_conn(&self.base, f)?;
 
         if let Some(parent) = &self.parent {
-            write!(f, " parent [")?;
+            write!(f, "\n\\ parent [")?;
             Self::format_conn(parent, f)?;
             write!(f, "]")?;
         }
--- a/retis-events/src/helpers.rs
+++ b/retis-events/src/helpers.rs
@@ -54,6 +54,27 @@ pub(crate) fn protocol_str(protocol: u8) -> Option<&'static str> {
     })
 }
 
+/// u128 representation in the events. We can't use the Rust primitive as serde
+/// does not handle the type well.
+#[event_type]
+pub struct U128 {
+    hi: u64,
+    lo: u64,
+}
+
+impl U128 {
+    pub fn from_u128(from: u128) -> Self {
+        Self {
+            hi: (from >> 64) as u64,
+            lo: from as u64,
+        }
+    }
+
+    pub fn bits(&self) -> u128 {
+        (self.hi as u128) << 64 | self.lo as u128
+    }
+}
+
 /// Represents a raw packet. Stored internally as a `Vec<u8>`.
 /// We don't use #[event_type] as we're implementing serde::Serialize and
 /// serde::Deserialize manually.
--- a/retis-events/src/helpers.rs
+++ b/retis-events/src/helpers.rs
@@ -5,6 +5,8 @@ use base64::{
     display::Base64Display, engine::general_purpose::STANDARD, prelude::BASE64_STANDARD, Engine,
 };
 
+use crate::event_type;
+
 /// Returns a translation of some ethertypes into a readable format.
 pub fn etype_str(etype: u16) -> Option<&'static str> {
     Some(match etype {
--- a/retis/src/module/ct/bpf/ct.bpf.c
+++ b/retis/src/module/ct/bpf/ct.bpf.c
@@ -59,6 +59,7 @@ struct ct_event {
 	struct nf_conn_tuple orig;
 	struct nf_conn_tuple reply;
 	u32 flags;
+	u32 mark;
 	u16 zone_id;
 	u8 tcp_state;
 } __binding;
@@ -100,6 +101,9 @@ static __always_inline int process_nf_conn(struct ct_event *e,
 
 		e->zone_id = (u8) BPF_CORE_READ(ct, zone.id);
 	}
+
+	if (bpf_core_field_exists(ct->mark))
+		e->mark = BPF_CORE_READ(ct, mark);
 
 	switch (l3num) {
 	case NFPROTO_IPV4:
@@ -212,16 +216,16 @@ DEFINE_HOOK(F_AND, RETIS_ALL_FILTERS,
 	if (!ct_protocol_is_supported(l3num, protonum))
 		return 0;
 
-	e = get_event_section(event, COLLECTOR_CT, SECTION_BASE_CONN,
-			      sizeof(*e));
+	e = get_event_zsection(event, COLLECTOR_CT, SECTION_BASE_CONN,
+			       sizeof(*e));
 	if (!e)
 		return 0;
 	process_nf_conn(e, nf_conn, l3num, protonum);
 
 	nf_conn = BPF_CORE_READ(nf_conn, master);
 	if (nf_conn) {
-		e = get_event_section(event, COLLECTOR_CT, SECTION_PARENT_CONN,
-				      sizeof(*e));
+		e = get_event_zsection(event, COLLECTOR_CT, SECTION_PARENT_CONN,
+				       sizeof(*e));
 		if (!e)
 			return 0;
 		process_nf_conn(e, nf_conn,
--- a/retis/src/module/ct/bpf/ct.bpf.c
+++ b/retis/src/module/ct/bpf/ct.bpf.c
@@ -60,6 +60,7 @@ struct ct_event {
 	struct nf_conn_tuple reply;
 	u32 flags;
 	u32 mark;
+	u8 labels[16];
 	u16 zone_id;
 	u8 tcp_state;
 } __binding;
@@ -84,6 +85,34 @@ static __always_inline bool ct_protocol_is_supported(u16 l3num, u8 protonum)
 	}
 
 	return true;
+}
+
+/* See ctnetlink_dump_labels(). */
+static __always_inline void get_nf_ct_labels(struct ct_event *e,
+					     struct nf_conn *ct)
+{
+	struct nf_ct_ext *ext = BPF_CORE_READ(ct, ext);
+	struct nf_conn_labels *labels;
+	int offset, nf_ct_ext_labels;
+
+	/* Conntrack labels depend on CONFIG_NF_CONNTRACK_LABELS, the following
+	 * enum variant is only defined if enabled.
+	 */
+	if (!bpf_core_enum_value_exists(enum nf_ct_ext_id, NF_CT_EXT_LABELS))
+		return;
+
+	if (!ext)
+		return;
+
+	nf_ct_ext_labels = bpf_core_enum_value(enum nf_ct_ext_id, NF_CT_EXT_LABELS);
+	offset = BPF_CORE_READ(ext, offset)[nf_ct_ext_labels];
+	if (!offset)
+		return;
+
+	labels = (void *)ext + offset;
+
+	BUILD_BUG_ON(sizeof(labels->bits) != sizeof(e->labels));
+	bpf_core_read(&e->labels, sizeof(labels->bits), &labels->bits);
 }
 
 static __always_inline int process_nf_conn(struct ct_event *e,
@@ -183,6 +212,8 @@ static __always_inline int process_nf_conn(struct ct_event *e,
 		break;
 	}
 
+	get_nf_ct_labels(e, ct);
+
 	return 0;
 }
 
@@ -231,6 +262,7 @@ DEFINE_HOOK(F_AND, RETIS_ALL_FILTERS,
 		process_nf_conn(e, nf_conn,
 				(u16)BPF_CORE_READ(nf_conn, ORIG.src.l3num),
 				(u8)BPF_CORE_READ(nf_conn, ORIG.dst.protonum));
+
 	}
 
 	m = get_event_section(event, COLLECTOR_CT, SECTION_META, sizeof(*m));
--- a/retis-events/src/lib.rs
+++ b/retis-events/src/lib.rs
@@ -13,7 +13,7 @@ pub mod display;
 pub use display::*;
 
 pub mod file;
-pub mod net;
+pub mod helpers;
 #[cfg(feature = "python")]
 pub mod python;
 #[cfg(feature = "python-embed")]
--- a/retis-events/src/skb.rs
+++ b/retis-events/src/skb.rs
@@ -1,7 +1,7 @@
 use std::fmt;
 
 use super::{
-    net::{etype_str, protocol_str, RawPacket},
+    helpers::{etype_str, protocol_str, RawPacket},
     *,
 };
 use crate::{event_section, event_type, Formatter};
--- a/retis/src/bindings/ct_uapi.rs
+++ b/retis/src/bindings/ct_uapi.rs
@@ -74,6 +74,7 @@ pub struct ct_event {
     pub orig: nf_conn_tuple,
     pub reply: nf_conn_tuple,
     pub flags: u32_,
+    pub mark: u32_,
     pub zone_id: u16_,
     pub tcp_state: u8_,
 }
--- a/retis/src/bindings/ct_uapi.rs
+++ b/retis/src/bindings/ct_uapi.rs
@@ -75,6 +75,7 @@ pub struct ct_event {
     pub reply: nf_conn_tuple,
     pub flags: u32_,
     pub mark: u32_,
+    pub labels: [u8_; 16usize],
     pub zone_id: u16_,
     pub tcp_state: u8_,
 }
--- a/retis/src/core/probe/kernel/bpf/include/helpers.h
+++ b/retis/src/core/probe/kernel/bpf/include/helpers.h
@@ -5,6 +5,12 @@
 
 #define MIN(a, b)	(((a) < (b)) ? (a) : (b))
 
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
+#define BUILD_BUG_ON(cond)	_Static_assert(!(cond), "BUILD_BUG_ON failed " #cond)
+#else
+#define BUILD_BUG_ON(cond)
+#endif
+
 enum bpf_func_id___x { BPF_FUNC_get_func_ip___5_15_0 = 42 };
 
 /* The following helper retrieves the function IP in kprobes.
--- a/retis/src/module/ct/bpf.rs
+++ b/retis/src/module/ct/bpf.rs
@@ -16,7 +16,7 @@ use crate::{
         inspect::inspector,
     },
     event_section_factory,
-    events::*,
+    events::{helpers::U128, *},
     helpers,
 };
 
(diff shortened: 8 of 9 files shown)


atenart:
Documentation of the new fields is missing, this should be done before merging.

retis-events/src/ct.rs
```diff
@@ -125,6 +125,10 @@ pub struct CtConnEvent {
     pub reply: CtTuple,
     /// TCP state; if any
     pub tcp_state: Option<String>,
+    /// Connection mark tracking label
+    pub mark: Option<u32>,
+    /// Connection tracking labels.
+    pub labels: Option<u128>,
```
atenart, line 131: The documentation of those new fields is missing.

retis/src/module/ct/bpf/ct.bpf.c
```diff
@@ -227,6 +260,7 @@ DEFINE_HOOK(F_AND, RETIS_ALL_FILTERS,
 		process_nf_conn(e, nf_conn,
 				(u16)BPF_CORE_READ(nf_conn, ORIG.src.l3num),
 				(u8)BPF_CORE_READ(nf_conn, ORIG.dst.protonum));
+
```
atenart, line 263: Extra new line.

retis-events/src/helpers.rs
```diff
@@ -52,6 +54,27 @@ pub(crate) fn protocol_str(protocol: u8) -> Option<&'static str> {
     })
 }
 
+/// u128 representation in the events. We can't use the Rust primitive as serde
+/// does not handle the type well.
+#[event_type]
+pub struct U128 {
+    hi: u64,
+    lo: u64,
+}
```
atenart, line 63: Due to serde limitations I chose to represent `u128` as the above. Another way would be to use something like `U128(u128)` and represent it as a string in JSON.

igsilya:
Feels like I'm getting some random garbage instead of labels.  Every time I run the test they are different or not show up at all.  For example:
```
1326209075534 (12) [swapper/12] 0 [tp] openvswitch:ovs_do_execute_action #134c83864ebffff98c3d24a6400 (skb ffff98c3c2e21200) n 0
  if 97 (ovs-p3) rxif 97 10.1.1.4.80 > 10.1.1.3.46922 ttl 64 tos 0x0 id 0 off 0 [DF] len 60 proto TCP (6) flags [S.] seq 769399365 ack 3766449066 win 65160
  exec ct zone 0
  ↳ 1326209079076 (12) [swapper/12] 0 [tp] openvswitch:ovs_do_execute_action #134c83864ebffff98c3d24a6400 (skb ffff98c3c2e21200) n 1
      if 97 (ovs-p3) rxif 97 10.1.1.4.80 > 10.1.1.3.46922 ttl 64 tos 0x0 id 0 off 0 [DF] len 60 proto TCP (6) flags [S.] seq 769399365 ack 3766449066 win 65160
      exec recirc 0x2
      ct_state REPLY tcp (SYN_RECV) orig [10.1.1.3.46922 > 10.1.1.4.80] reply [10.1.1.4.80 > 10.1.1.3.46922] zone 0 mark 0
  ↳ 1326209086216 (12) [swapper/12] 0 [tp] skb:kfree_skb #134c83864ebffff98c3d24a6400 (skb ffff98c3c2e21200) n 2 drop (reason openvswitch/OVS_DROP_EXPLICIT)
      if 97 (ovs-p3) rxif 97 10.1.1.4.80 > 10.1.1.3.46922 ttl 64 tos 0x0 id 0 off 0 [DF] len 60 proto TCP (6) flags [S.] seq 769399365 ack 3766449066 win 65160
      ct_state REPLY tcp (SYN_RECV) orig [10.1.1.3.46922 > 10.1.1.4.80] reply [10.1.1.4.80 > 10.1.1.3.46922] zone 0 mark 0 labels 0x64766564752d646d65
```
It supposed to be the same ct entry, but the labels are not shown the first time.  And the `0x64766564752d646d65` looks suspiciously like an ASCII text. :)

The test that produced this is using either `0x2` or `0xa000d000005000001` as a label and the test passes, so these values are there.

igsilya:
For the formatting of the parent entry, it looks like this:
```
1974207113815 (0) [python3] 15686 [tp] openvswitch:ovs_do_execute_action #1cba7e9c346ffff98c3d19d0f00 (skb ffff98c3f246eae8) n 0
  if 106 (ovs-p1) rxif 106 10.1.1.2.45333 > 10.1.1.9.39971 ttl 64 tos 0x0 id 29108 off 0 [DF] len 60 proto TCP (6) flags [S] seq 359608549 win 64240
  exec ct zone 1 nat
  ↳ 1974207122891 (0) [python3] 15686 [tp] openvswitch:ovs_do_execute_action #1cba7e9c346ffff98c3d19d0f00 (skb ffff98c3f246eae8) n 1
      if 106 (ovs-p1) rxif 106 10.1.1.2.45333 > 10.1.1.9.39971 ttl 64 tos 0x0 id 29108 off 0 [DF] len 60 proto TCP (6) flags [S] seq 359608549 win 64240
      exec recirc 0x2
      ct_state RELATED tcp (SYN_SENT) orig [10.1.1.2.45333 > 10.1.1.9.39971] reply [10.1.1.1.39971 > 10.1.1.2.45333] zone 1 mark 0
      parent [tcp (ESTABLISHED) orig [10.1.1.1.56222 > 10.1.1.2.21] reply [10.1.1.2.21 > 10.1.1.9.56222] zone 1 mark 0]
  ↳ 1974207131313 (0) [python3] 15686 [tp] openvswitch:ovs_dp_upcall #1cba7e9c346ffff98c3d19d0f00 (skb ffff98c3f246eae8) n 2
      if 106 (ovs-p1) rxif 106 10.1.1.2.45333 > 10.1.1.9.39971 ttl 64 tos 0x0 id 29108 off 0 [DF] len 60 proto TCP (6) flags [S] seq 359608549 win 64240
      upcall (miss) port 2548630282 cpu 0
      ct_state RELATED tcp (SYN_SENT) orig [10.1.1.2.45333 > 10.1.1.9.39971] reply [10.1.1.1.39971 > 10.1.1.2.45333] zone 1 mark 0
      parent [tcp (ESTABLISHED) orig [10.1.1.1.56222 > 10.1.1.2.21] reply [10.1.1.2.21 > 10.1.1.9.56222] zone 1 mark 0]
  ↳ 1974207149639 (0) [python3] 15686 [kr] ovs_dp_upcall #1cba7e9c346ffff98c3d19d0f00 (skb ffff98c3f246eae8) n 3
      if 106 (ovs-p1) rxif 106 10.1.1.2.45333 > 10.1.1.9.39971 ttl 64 tos 0x0 id 29108 off 0 [DF] len 60 proto TCP (6) flags [S] seq 359608549 win 64240
      upcall_ret (0/1974207131313) ret 0
      ct_state RELATED tcp (SYN_SENT) orig [10.1.1.2.45333 > 10.1.1.9.39971] reply [10.1.1.1.39971 > 10.1.1.2.45333] zone 1 mark 0
      parent [tcp (ESTABLISHED) orig [10.1.1.1.56222 > 10.1.1.2.21] reply [10.1.1.2.21 > 10.1.1.9.56222] zone 1 mark 0]
```

It might be better if the parent was indented further to the right, so it better aligns with the current connection and that it looks more like part of the `ct_state` and not something separate from it.  It doesn't have to align perfectly, and it is not possible unless we align parts of the message internally, but it might make sense to at least shift the whole thing `len("ct_state ")` characters to the right.  WDYT?

```
  ↳ 1974207149639 (0) [python3] 15686 [kr] ovs_dp_upcall #1cba7e9c346ffff98c3d19d0f00 (skb ffff98c3f246eae8) n 3
      if 106 (ovs-p1) rxif 106 10.1.1.2.45333 > 10.1.1.9.39971 ttl 64 tos 0x0 id 29108 off 0 [DF] len 60 proto TCP (6) flags [S] seq 359608549 win 64240
      upcall_ret (0/1974207131313) ret 0
      ct_state RELATED tcp (SYN_SENT) orig [10.1.1.2.45333 > 10.1.1.9.39971] reply [10.1.1.1.39971 > 10.1.1.2.45333] zone 1 mark 0
               parent [tcp (ESTABLISHED) orig [10.1.1.1.56222 > 10.1.1.2.21] reply [10.1.1.2.21 > 10.1.1.9.56222] zone 1 mark 0]
```

atenart:
> Feels like I'm getting some random garbage instead of labels. Every time I run the test they are different or not show up at all. For example:
> 
> ```
> 1326209075534 (12) [swapper/12] 0 [tp] openvswitch:ovs_do_execute_action #134c83864ebffff98c3d24a6400 (skb ffff98c3c2e21200) n 0
>   if 97 (ovs-p3) rxif 97 10.1.1.4.80 > 10.1.1.3.46922 ttl 64 tos 0x0 id 0 off 0 [DF] len 60 proto TCP (6) flags [S.] seq 769399365 ack 3766449066 win 65160
>   exec ct zone 0
>   ↳ 1326209079076 (12) [swapper/12] 0 [tp] openvswitch:ovs_do_execute_action #134c83864ebffff98c3d24a6400 (skb ffff98c3c2e21200) n 1
>       if 97 (ovs-p3) rxif 97 10.1.1.4.80 > 10.1.1.3.46922 ttl 64 tos 0x0 id 0 off 0 [DF] len 60 proto TCP (6) flags [S.] seq 769399365 ack 3766449066 win 65160
>       exec recirc 0x2
>       ct_state REPLY tcp (SYN_RECV) orig [10.1.1.3.46922 > 10.1.1.4.80] reply [10.1.1.4.80 > 10.1.1.3.46922] zone 0 mark 0
>   ↳ 1326209086216 (12) [swapper/12] 0 [tp] skb:kfree_skb #134c83864ebffff98c3d24a6400 (skb ffff98c3c2e21200) n 2 drop (reason openvswitch/OVS_DROP_EXPLICIT)
>       if 97 (ovs-p3) rxif 97 10.1.1.4.80 > 10.1.1.3.46922 ttl 64 tos 0x0 id 0 off 0 [DF] len 60 proto TCP (6) flags [S.] seq 769399365 ack 3766449066 win 65160
>       ct_state REPLY tcp (SYN_RECV) orig [10.1.1.3.46922 > 10.1.1.4.80] reply [10.1.1.4.80 > 10.1.1.3.46922] zone 0 mark 0 labels 0x64766564752d646d65
> ```
> 
> It supposed to be the same ct entry, but the labels are not shown the first time. And the `0x64766564752d646d65` looks suspiciously like an ASCII text. :)
> 
> The test that produced this is using either `0x2` or `0xa000d000005000001` as a label and the test passes, so these values are there.

Thanks for testing! I believe there are two issues:

- The garbage value being reported is an issue in how we report labels from BPF to userspace: if no label needs to be reported the data is not initialized to 0 and garbage is reported and parsed as valid labels. This is easy to fix.
- It seems the BPF side fails to identify when labels are available and only report them in some cases. I'll investigate.

(The above is probably a combination of the two issues).

Also there is the issue reported in #445.

atenart:
> It might be better if the parent was indented further to the right, so it better aligns with the current connection and that it looks more like part of the `ct_state` and not something separate from it. It doesn't have to align perfectly, and it is not possible unless we align parts of the message internally, but it might make sense to at least shift the whole thing `len("ct_state ")` characters to the right. WDYT?
> 
> ```
>   ↳ 1974207149639 (0) [python3] 15686 [kr] ovs_dp_upcall #1cba7e9c346ffff98c3d19d0f00 (skb ffff98c3f246eae8) n 3
>       if 106 (ovs-p1) rxif 106 10.1.1.2.45333 > 10.1.1.9.39971 ttl 64 tos 0x0 id 29108 off 0 [DF] len 60 proto TCP (6) flags [S] seq 359608549 win 64240
>       upcall_ret (0/1974207131313) ret 0
>       ct_state RELATED tcp (SYN_SENT) orig [10.1.1.2.45333 > 10.1.1.9.39971] reply [10.1.1.1.39971 > 10.1.1.2.45333] zone 1 mark 0
>                parent [tcp (ESTABLISHED) orig [10.1.1.1.56222 > 10.1.1.2.21] reply [10.1.1.2.21 > 10.1.1.9.56222] zone 1 mark 0]
> ```

IMO we should avoid to shift things more, we already have events (in series) and their sections. Maybe instead we can use a special char to start the parent line? `↳` is already used but we can find something else. If no better idea `\` could do.

```
  ↳ 1974207149639 (0) [python3] 15686 [kr] ovs_dp_upcall #1cba7e9c346ffff98c3d19d0f00 (skb ffff98c3f246eae8) n 3
      if 106 (ovs-p1) rxif 106 10.1.1.2.45333 > 10.1.1.9.39971 ttl 64 tos 0x0 id 29108 off 0 [DF] len 60 proto TCP (6) flags [S] seq 359608549 win 64240
      upcall_ret (0/1974207131313) ret 0
      ct_state RELATED tcp (SYN_SENT) orig [10.1.1.2.45333 > 10.1.1.9.39971] reply [10.1.1.1.39971 > 10.1.1.2.45333] zone 1 mark 0
      \ parent [tcp (ESTABLISHED) orig [10.1.1.1.56222 > 10.1.1.2.21] reply [10.1.1.2.21 > 10.1.1.9.56222] zone 1 mark 0]
```

WDYT?

igsilya:
> IMO we should avoid to shift things more, we already have events (in series) and their sections. Maybe instead we can use a special char to start the parent line?

I like how the protocol and reply align in the shifted version, and it's easier to compare IPs this way.  But I won't insist.  Having a special character seems better than not having it.

atenart:
@igsilya the various issues should be fixed now. I refreshed the container image.

igsilya:
Updated version seems to work:
```
25320889831 (24) [python3] 2994 [tp] openvswitch:ovs_do_execute_action #5e53dfd42ffff90be14a28280 (skb ffff90be0ac7e0e8) n 0
  if 10 (ovs-p1) rxif 10 10.1.1.2.59825 > 10.1.1.9.44731 ttl 64 tos 0x0 id 29302 off 0 [DF] len 1439 proto TCP (6) flags [P.] seq 2695371793:2695373180 ack 848702181 win 502
  exec ct zone 1 nat
  ↳ 25320894171 (24) [python3] 2994 [tp] openvswitch:ovs_do_execute_action #5e53dfd42ffff90be14a28280 (skb ffff90be0ac7e0e8) n 1
      if 10 (ovs-p1) rxif 10 10.1.1.2.59825 > 10.1.1.1.44731 ttl 64 tos 0x0 id 29302 off 0 [DF] len 1439 proto TCP (6) flags [P.] seq 2695371793:2695373180 ack 848702181 win 502
      exec recirc 0x2
      ct_state ESTABLISHED tcp (ESTABLISHED) orig [10.1.1.2.59825 > 10.1.1.9.44731] reply [10.1.1.1.44731 > 10.1.1.2.59825] zone 1 mark 0 labels 0x4d2000000000000000000000001
      \ parent [tcp (ESTABLISHED) orig [10.1.1.1.39854 > 10.1.1.2.21] reply [10.1.1.2.21 > 10.1.1.9.39854] zone 1 mark 0 labels 0x4d2000000000000000000000001]
  ↳ 25320900071 (24) [python3] 2994 [tp] openvswitch:ovs_do_execute_action #5e53dfd42ffff90be14a28280 (skb ffff90be0ac7e0e8) n 2
      if 10 (ovs-p1) rxif 10 10.1.1.2.59825 > 10.1.1.1.44731 ttl 64 tos 0x0 id 29302 off 0 [DF] len 1439 proto TCP (6) flags [P.] seq 2695371793:2695373180 ack 848702181 win 502
      [recirc_id 0x2] exec oport 2
      ct_state ESTABLISHED tcp (ESTABLISHED) orig [10.1.1.2.59825 > 10.1.1.9.44731] reply [10.1.1.1.44731 > 10.1.1.2.59825] zone 1 mark 0 labels 0x4d2000000000000000000000001
      \ parent [tcp (ESTABLISHED) orig [10.1.1.1.39854 > 10.1.1.2.21] reply [10.1.1.2.21 > 10.1.1.9.39854] zone 1 mark 0 labels 0x4d2000000000000000000000001]
```
Tested with RHEL 9.4 stock kernel and 6.10-rc6 upstream kernel.  I no longer see any garbage data and the labels are the labels match what OVS has in OpenFlow rules.  Also works without kenrel config available.

atenart:
> Tested with RHEL 9.4 stock kernel and 6.10-rc6 upstream kernel. I no longer see any garbage data and the labels are the labels match what OVS has in OpenFlow rules. Also works without kenrel config available.

Thanks for testing!

retis/src/module/ct/bpf/ct.bpf.c
```diff
@@ -86,6 +87,34 @@ static __always_inline bool ct_protocol_is_supported(u16 l3num, u8 protonum)
...
+		return;
+
+	nf_ct_ext_labels = bpf_core_enum_value(enum nf_ct_ext_id, NF_CT_EXT_LABELS);
+	offset = BPF_CORE_READ(ext, offset)[nf_ct_ext_labels];
+	if (!offset)
+		return;
+
+	labels = (void *)ext + offset;
+
+	BUILD_BUG_ON(sizeof(labels->bits) != sizeof(e->labels));
```
vlrpl, line 114: this should statically check at compile time against `vmlinux.h` definition.
```c
#define member_sz(type, member) sizeof(((type *)0)->member)
BUILD_BUG_ON(member_sz(struct nf_conn_labels, bits) != member_sz(struct ct_event, labels));
```
another approach can be the above putting `BUILD_BUG_ON()` below the `__binding` definition.
I slightly prefer having the static assertion close to the definition, but it's up to you.

That being said, I wonder if `bpf_core_field_size()` could be helpful here (although I think you're intentionally checking against `vmlinux.h`), so probably it's not the case.

atenart: > this should statically check at compile time against `vmlinux.h` definition.

Isn't that what's already done here?

> ```c
> #define member_sz(type, member) sizeof(((type *)0)->member)
> BUILD_BUG_ON(member_sz(struct nf_conn_labels, bits) != member_sz(struct ct_event, labels));
> ```
> 
> another approach can be the above putting `BUILD_BUG_ON()` below the `__binding` definition. I slightly prefer having the static assertion close to the definition, but it's up to you.

I guess both make sense, close to the definition vs close to where the actual copy is made. The second option has the advantage of not requiring an additional (`member_sz`) definition.

> That being said, I wonder if `bpf_core_field_size()` could be helpful here (although I think you're intentionally checking against `vmlinux.h`), so probably it's not the case.

Yes, that's on purpose as there's no kernel version having different definitions so far. This is to prevent incompatibility when upgrading `vmlinux.h` in case this changes at some point. Catching runtime issues is not in the scope and many other places would be problematic as we can't check for every variable being read at runtime.

vlrpl: > > this should statically check at compile time against `vmlinux.h` definition.
> 
> Isn't that what's already done here?
> 

Of course, it referred to your change. I rephrased the comment backward before posting and this probably was a bit unclear/redundant :)

> > ```c
> > #define member_sz(type, member) sizeof(((type *)0)->member)
> > BUILD_BUG_ON(member_sz(struct nf_conn_labels, bits) != member_sz(struct ct_event, labels));
> > ```
> >   
> > another approach can be the above putting `BUILD_BUG_ON()` below the `__binding` definition. I slightly prefer having the static assertion close to the definition, but it's up to you.
> 
> I guess both make sense, close to the definition vs close to where the actual copy is made. The second option has the advantage of not requiring an additional (`member_sz`) definition.
> 

There are of course advantages in the other way, but as I said, it's up to you. 

> > That being said, I wonder if `bpf_core_field_size()` could be helpful here (although I think you're intentionally checking against `vmlinux.h`), so probably it's not the case.
> 
> Yes, that's on purpose as there's no kernel version having different definitions so far. This is to prevent incompatibility when upgrading `vmlinux.h` in case this changes at some point. Catching runtime issues is not in the scope and many other places would be problematic as we can't check for every variable being read at runtime.

many other places would have been my follow-up :)
Just checking.

atenart: > > > this should statically check at compile time against `vmlinux.h` definition.
> > 
> > Isn't that what's already done here?
> 
> Of course, it referred to your change. I rephrased the comment backward before posting and this probably was a bit unclear/redundant :)

Thanks for clarifying, I was not sure if you meant this or if I was missing something :)

> > > ```c
> > > #define member_sz(type, member) sizeof(((type *)0)->member)
> > > BUILD_BUG_ON(member_sz(struct nf_conn_labels, bits) != member_sz(struct ct_event, labels));
> > > ```
> > >   
> > > another approach can be the above putting `BUILD_BUG_ON()` below the `__binding` definition. I slightly prefer having the static assertion close to the definition, but it's up to you.
> > 
> > 
> > I guess both make sense, close to the definition vs close to where the actual copy is made. The second option has the advantage of not requiring an additional (`member_sz`) definition.
> 
> There are of course advantages in the other way, but as I said, it's up to you.

Let's not introduce another helper then, but overall I don't have a strong preference.

> > > That being said, I wonder if `bpf_core_field_size()` could be helpful here (although I think you're intentionally checking against `vmlinux.h`), so probably it's not the case.
> > 
> > Yes, that's on purpose as there's no kernel version having different definitions so far. This is to prevent incompatibility when upgrading `vmlinux.h` in case this changes at some point. Catching runtime issues is not in the scope and many other places would be problematic as we can't check for every variable being read at runtime.
> 
> many other places would have been my follow-up :) Just checking.

Every bit copied from one place to an event can be checked, although not all copies can be problematic. This specific case could benefit from a runtime check, but I guess it would just flag one of the many possible incompatibilities we could see with newer kernels. In such cases all those things are just bugs we'll have to fix anyway. No strong opinion on this too.

I did check other places when doing this patch and the other ones where:
- IP src/dst
- Ports
- ICMP id

All those are protocol-related, I don't think a check is needed.

vlrpl: yep, I agree (especially with the "every place read at runtime" part).

vlrpl (approved):
thanks, lgtm!

PR #435, opened 2024-10-23
User defined probe UX improvements

atenart:
A few improvements in how we parse `--probe` cli arguments, to make them easier to use and more aligned to other tools with similar arguments.


vlrpl (approved):
lgtm, thanks!

PR #436, opened 2024-10-25
Minor Python support follow-ups

atenart:


retis-events/src/python.rs
```diff
@@ -75,9 +75,10 @@ impl PyEvent {
     /// Allows to use the object as a dictionary, eg. `e['skb']`.
     fn __getitem__<'a>(&'a self, py: Python<'a>, attr: &str) -> PyResult<Py<PyAny>> {
         if let Ok(id) = SectionId::from_str(attr) {
-            if let Some(section) = self.0.get(id) {
-                return Ok(section.to_py(py));
-            }
+            return Ok(match self.0.get(id) {
+                Some(section) => section.to_py(py),
+                None => py.None(),
```
amorenoz, line 80: All the benefits you mention are true. However, this is completely against python "standards". I'm not sure if there is a PEP for this but en every object out there, if it supports `__getitem__` it returns a `KeyError` if the key does not exist. Not throwing but returning `None` would be extremely confusing IMHO.

Given that `__getitem__` has such a well established semantics, I would avoid breaking it here and instead add another helper such as the one in `dict`: https://docs.python.org/3.8/library/stdtypes.html#dict.get

That would allow a cleaner syntax ( `if event.get("skb", None): [...]`) which is both more explicit and does not break existing semantics.

retis/src/process/cli/sort.rs
```diff
@@ -72,7 +72,7 @@ impl SubCommandParserRunner for Sort {
         let mut factory = FileEventsFactory::new(self.input.as_path())?;
 
         if matches!(factory.file_type(), file::FileType::Series) {
-            println!("File already sorted");
+            log::info!("File already sorted");
```
amorenoz, line 75: If we run without logs, or with suppressed stderr we won't see anything. Wouldn't that be confusing?

atenart: In case `-o` is used no, in case stdout is you're right. However all error / info messages are handles by `log` (we even remove the level from the output from `info!` messages are they're part of the expected output), the above was for consistency. If the user explicitly removes those messages then it's expected to have less explanations in regards to what is happening.

No strong feeling though, I can remove it if you prefer.

amorenoz: > Another option is to use `log:info!` in case `-o` is used and `bail!` otherwise.


Why not bail in both cases?

amorenoz: I'm not sure if I like this conditional error. I prefer just using `log` as it is in the patch.

> We could improve things to make `sort` only sorting a file and `print` to have a `--sort` option. Not 100% sure about that though.

Me neither.

amorenoz (approved):
LGMT. Thanks

PR #444, opened 2024-11-08
events: formatting: fix newline at end of buffer handling

atenart:
When a newline is found at the end of the buffer it was printed twice. This is because splitting such string will result in having an empty last split. The logic was then printing a new line for the split and because one was at the end of the buffer.

This also prevents from printing the prefix and fixes cases where a double newline is found in the middle of the buffer.


vlrpl (approved):
lgtm.
@amorenoz, I'll leave the last word on this to you as you may have a better chance of finding any potential problem with your repro.

amorenoz:
Tested on my reproducer and confirmed it fixes it.

PR #446, opened 2024-11-10
doc: introduce code coverage section

vlrpl:
as per title

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -125,7 +125,7 @@ functional_task:
       # Installing it from Hashicorp directly
     - wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor -o /usr/share/keyrings/hashicorp-archive-keyring.gpg
     - echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(lsb_release -cs) main" | tee /etc/apt/sources.list.d/hashicorp.list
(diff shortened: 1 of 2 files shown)


vlrpl:
I snuck into this the vagrant workaround removal as this should be fixed now. For that reason, the PR is labeled with `run-functional-tests`.

atenart (changes requested):
A few comments:
- I tried using it and it seems `cargo-llvm-cov` needs to be installed as well as the rustup `llvm-tools-preview` component.
- Can you use an example for `toolchain` so users know what to look for?
- The `CARGO_OPTS` use should be in its own paragraph, but not in the default example.
- Since the html output is optional, maybe let's not use it in the default example but alongside the last paragraph. Eg. "for redability you can ...".

vlrpl:
> A few comments:
> 
>     * I tried using it and it seems `cargo-llvm-cov` needs to be installed as well as the rustup `llvm-tools-preview` component.
> 

right, these are needed.

>     * Can you use an example for `toolchain` so users know what to look for?
> 

this could be whatever it is installed via rustup, so could be any version (maybe names can be assigned). 
I guess if we want this, we could use +nightly or +stable as those aliases should be safe enough.

>     * The `CARGO_OPTS` use should be in its own paragraph, but not in the default example.
> 
>     * Since the html output is optional, maybe let's not use it in the default example but alongside the last paragraph. Eg. "for redability you can ...".

ok, let's move optional stuff in their paragraph.

atenart:
> > ```
> > * Can you use an example for `toolchain` so users know what to look for?
> > ```
> 
> this could be whatever it is installed via rustup, so could be any version (maybe names can be assigned). I guess if we want this, we could use +nightly or +stable as those aliases should be safe enough.

Could be any version, I think the explanation is good in this regards. I just think using whatever version for the example would gives an idea what is expected.

atenart (approved):
Looking good, thanks!

Issue #437, opened 2024-10-28
Improve python member access of enums


amorenoz:
Follow up after #426
Based on https://github.com/retis-org/retis/pull/426#issuecomment-2389666435, look at what enums can be modified slightly in order to improve how they are accessed from python.

PR #447, opened 2024-11-10
Ovs flows

vlrpl:
extract ufid and other info


vlrpl:
Functional tests are currently broken because of an [issue](https://github.com/hashicorp/vagrant/issues/13534) on Vagrant 2.4.2.

retis-events/src/ovs.rs
```diff
@@ -107,6 +112,59 @@ impl EventFmt for UpcallEvent {
...
+        ufid[1] & 0xffff,
+        ufid[2] >> 16,
+        ufid[2] & 0xffff,
+        ufid[3]
+    )
+}
+
+#[event_type]
+#[derive(Copy, Default, PartialEq)]
+pub struct Ufid(pub [u32; 4]);
```
amorenoz, line 129: Not a full review, just a comment that came up when integrating.

I had this:
```rust
#[event_type]
#[derive(Copy, Default, PartialEq)]
pub struct UFID {
    parts: [u32; 4],
}
```

Looking at the convention taken elsewhere I think your name is probably preferred.

However, whether to use tuple vs named struct took me some thinking. I think tuples are nice if you don't use inner objects too much or else `self.0` is a bit confusing to me when reading. Having it `pub` is definitely a red flag, even more in this case where UFID.0 could be thought to be the first 32bit part.

Why did you need it to be public?
Also, note that internal members of `event_type`s are accessed directly in python and tuples are a bit confusing (https://github.com/retis-org/retis/issues/437).

amorenoz: How about this for a middle ground?
```rust
#[event_type]
#[derive(Copy, Default, PartialEq)]
pub struct UFID(pub u32, pub u32, pub u32, pub u32);

```

vlrpl: > Not a full review, just a comment that came up when integrating.
> 
> I had this:
> 
> ```rust
> #[event_type]
> #[derive(Copy, Default, PartialEq)]
> pub struct UFID {
>     parts: [u32; 4],
> }
> ```
> 
> Looking at the convention taken elsewhere I think your name is probably preferred.
> 
> However, whether to use tuple vs named struct took me some thinking. I think tuples are nice if you don't use inner objects too much or else `self.0` is a bit confusing to me when reading. Having it `pub` is definitely a red flag, even more in this case where UFID.0 could be thought to be the first 32bit part.
> 
> Why did you need it to be public? Also, note that internal members of `event_type`s are accessed directly in python and tuples are a bit confusing (#437).

No real reason. TBF, this was simply quicker to write&share to unblock the other part of the work.
Something like this will have been the final intended form:

```rust
impl Ufid {
    pub fn new_or_from_whatever(ufid: [u32; 4]) -> Self {
        Self(ufid)
    }
}
```

vlrpl: > How about this for a middle ground?
> 
> ```rust
> #[event_type]
> #[derive(Copy, Default, PartialEq)]
> pub struct UFID(pub u32, pub u32, pub u32, pub u32);
> ```

Having parts accessible with this form sgtm.

PR #452, opened 2024-11-14
Some terminal output improvements

atenart:
- Auto detect if we should output colors or not.
- Fix EPIPE handling when writing to stdout.


amorenoz (approved):
LGTM and output is nicer. Thanks

PR #453, opened 2024-11-20
docs: Fix "Edit on GitHub" links

drizzt:
Currently, "Edit on GitHub" links points to the wrong "master" branch so use "main" branch instead.

See https://github.com/mkdocs/mkdocs/issues/1578#issuecomment-410836503


atenart (approved):
LGTM, thanks!

(Note that the default readthedocs page is not from `main` but there's no way to make the edit button work in such case, so the fix does look good).

PR #455, opened 2024-11-29
use binding for packet filter

vlrpl:
Meta filter is going through the dynamic generation rework, so there's no need to migrate that, ATM.


atenart (approved):
LGTM, thanks!

PR #456, opened 2024-11-29
Add codecov support

vlrpl:
Upload coverage data to Codecov and consume the service

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -9,9 +9,11 @@ cargo_task_tempate: &cargo_task_template
       - cat Cargo.lock
   setup_script:
     - apt-get update
-    - apt-get -y install llvm clang libelf-dev libpcap-dev python3-pip make jq
+    - apt-get -y install llvm clang curl libelf-dev libpcap-dev python3-pip make jq
     - rustup component add rustfmt
     - rustup component add clippy
+    - rustup component add llvm-tools-preview
+    - cargo install cargo-llvm-cov
   before_cache_script: rm -rf $CARGO_HOME/registry/index
 
 unittest_task:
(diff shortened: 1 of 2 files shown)


atenart:
Since the actual behavior will only happen for later PRs, I think having a demo on an external GH repo with this included and dummy PRs could help. (Maybe you already have this handy).

vlrpl:
> Since the actual behavior will only happen for later PRs, I think having a demo on an external GH repo with this included and dummy PRs could help. (Maybe you already have this handy).

This is how it will look like on a forked repo:

https://github.com/vlrpl/retis/pull/4

.cirrus.yml
```diff
@@ -9,17 +9,23 @@ cargo_task_tempate: &cargo_task_template
...
   << : *cargo_task_template
   container:
     image: rust:bookworm
   build_script: make V=1 CARGO_OPTS=--verbose
-  test_script: make test V=1 CARGO_OPTS=--verbose
+  test_script:
+    - curl -Os https://uploader.codecov.io/latest/linux/codecov
+    - chmod +x codecov
+    - make test V=1 COV=1 CARGO_CMD_OPTS="--verbose --codecov --output-path retis.codecov"
+    - ./codecov --verbose upload-process --disable-search --fail-on-error --git-service github -F x86_64 -F unittest -f ./retis.codecov
```
atenart, line 28: It seems both coverage tests are using `-F unittest`. On the coverage report we see a line for each arch and a single line for unittest. I'm not sure what this line refers to then, or even if it's one or the other depending on timings. Is that required to set two `-F` args here (and for the other one) or should it be dropped?

.codecov.yml
```diff
@@ -0,0 +1,8 @@
+coverage:
+  status:
+    project:
+      default:
+        informational: true
```
atenart, line 5: Just to understand things better; "informational" means this is just providing the result without gating anything?

atenart (approved):
LGTM, thanks!

PR #470, opened 2025-01-09
Add VLAN ID information to metadata output

andreaskaris:
Add VLAN ID information to metadata output


retis/src/module/skb/bpf/skb_hook.bpf.c
```diff
@@ -1,6 +1,7 @@
 #include <vmlinux.h>
 #include <bpf/bpf_core_read.h>
 #include <bpf/bpf_endian.h>
+#include <if_vlan.h>
```
atenart, line 4: nit: please add this below the inclusion of "common.h", this block is for external ones.

atenart (changes requested):
Thanks for the PR!

Since the TCI is reported as part of the "meta" section, the question of retrieving it from the payload arises. But IMO it makes sense as this improves UX (I would just add a data source flag, see one of the comments).

There's also the question of double-tagging, but this is probably out of scope here. Any idea on how we should proceed (as a follow-up)?

The modifications already looks quite good, so I included nits comments as well.

docs/install.md
```diff
@@ -70,7 +70,7 @@ On Fedora, one can run:
 
 ```none
 $ dnf -y install git cargo clang elfutils-libelf-devel python3-devel \
-        jq libpcap-devel llvm make pkgconf-pkg-config
+        python3-clang jq libpcap-devel llvm make pkgconf-pkg-config
```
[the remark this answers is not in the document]
andreaskaris: I'll revert it with the next push, thanks

retis/src/module/skb/bpf/include/if_vlan.h
```diff
@@ -0,0 +1,107 @@
+#ifndef __MODULE_IF_VLAN__
```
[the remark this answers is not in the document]
andreaskaris: done

retis/src/module/skb/bpf/skb_hook.bpf.c
```diff
@@ -282,6 +284,7 @@ static __always_inline int process_skb(struct retis_raw_event *event,
 		e->csum = BPF_CORE_READ(skb, csum);
 		e->csum_level = (u8)BPF_CORE_READ_BITFIELD_PROBED(skb, csum_level);
 		e->priority = BPF_CORE_READ(skb, priority);
+		vlan_get_tag(skb, dev, &(e->vlan_tci));
```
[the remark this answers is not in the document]
andreaskaris: done

retis/src/module/skb/bpf/include/if_vlan.h
```diff
@@ -0,0 +1,107 @@
...
+static inline int __vlan_get_tag(const struct sk_buff *skb, u16 *vlan_tci)
+{
+	struct vlan_ethhdr *veth = skb_vlan_eth_hdr(skb);
+
+	__be16 h_vlan_proto;
+	bpf_probe_read_kernel(&h_vlan_proto, sizeof(__be16), &veth->h_vlan_proto);
+	if (!eth_type_vlan(h_vlan_proto))
+		return -ENODATA;
+
+    __be16 h_vlan_TCI;
```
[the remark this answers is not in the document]
andreaskaris: done

retis-events/src/skb.rs
```diff
@@ -463,6 +466,7 @@ pub struct SkbMetaEvent {
     pub csum_level: u8,
     /// QoS priority.
     pub priority: u32,
+    pub vlan_tci: u16,
```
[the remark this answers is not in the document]
andreaskaris: I think that's done ... let me know if that's what you had in mind

retis-events/src/skb.rs
```diff
@@ -244,6 +244,9 @@ impl EventFmt for SkbEvent {
                 if meta.data_len != 0 {
                     write!(f, "data_len {} ", meta.data_len)?;
                 }
+                if meta.vlan_tci != 0 {
+                    write!(f, "vlan_tci {:#x} vlan_id {} ", meta.vlan_tci, meta.vlan_tci & 0x0fff)?;
```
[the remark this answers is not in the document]
andreaskaris: > Since TCI includes all other sub-fields I would prefer the above to be something like vlan (id {} prio {} [drop]) with drop being a flag and set if applicable?

done

> In addition it would be nice to add another flag describing the source of the data, eg. accel (vlan (id {} prio {} [drop] [accel]) ). This would require an extra field in the raw and final events. WDYT?

let me look into / think about it

retis/src/module/skb/bpf/include/if_vlan.h
```diff
@@ -0,0 +1,107 @@
...
+ */
+static inline int vlan_get_tag(const struct sk_buff *skb, struct net_device *dev, u16 *vlan_tci)
+{
+	netdev_features_t features;
+	bpf_probe_read_kernel(&features, sizeof(netdev_features_t), &dev->features);
+	if (features & NETIF_F_HW_VLAN_CTAG_TX) {
+		return __vlan_hwaccel_get_tag(skb, vlan_tci);
+	} else {
+		return __vlan_get_tag(skb, vlan_tci);
+	}
```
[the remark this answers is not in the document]
andreaskaris: done

andreaskaris:
Next: looking into how to adjust unit tests

retis/src/module/skb/bpf/include/if_vlan.h
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * Returns error if @skb->vlan_tci is not set correctly
+ */
+static inline int __vlan_hwaccel_get_tag(const struct sk_buff *skb,
+					 u16 *vlan_tci)
+{
+	u16 vlan_proto = BPF_CORE_READ(skb, vlan_proto);
+	u16 temp_vlan_tci = BPF_CORE_READ(skb, vlan_tci);
+
+	if (vlan_tag_present(vlan_proto, vlan_tci)) {
```
atenart, line 35: s/vlan_tci/temp_vlan_tci/

docs/collectors/skb.md
```diff
@@ -104,7 +104,7 @@ Those two sections report metadata and reference counting from the socket buffer
 itself.
 
 ```none
-skb [{csum} hash {skb hash} data_len {skb data lenght} priority {skb priority}
+skb [{csum} hash {skb hash} data_len {skb data lenght} [vlan_tci {vlan_tci} vlan_id {vlan_id}] priority {skb priority}
```
atenart, line 107: \+ `[drop] [accel]`

retis/src/module/skb/bpf/include/if_vlan.h
```diff
@@ -0,0 +1,108 @@
...
+ */
+static inline void vlan_get_tag(const struct sk_buff *skb, u16 *vlan_tci, u8 *vlan_accel)
+{
+		if (!__vlan_hwaccel_get_tag(skb, vlan_tci)) {
+			*vlan_accel = 1;
+			return;
+		}
+
+		*vlan_accel = 0;
+		__vlan_get_tag(skb, vlan_tci);
```
atenart, line 105: nit: indent is one level too high.

atenart (commented):
A few comments but this looks good, thanks for taking the comments into account!

IMO this looks good as-is, but I thought a bit more and have a broader question (see below). Sorry this only comes now, but don't feel obligated to do the changes discussed there; I can do a follow-up before we release the feature. As you prefer :)

I'm wondering if the VLAN data should instead live in its own section in the skb event instead of the meta section, like ARP and others. Using the meta section was logical initially but we're now not only focusing on the skb metadata.
- This would align with the way other protocols are handled.
- VLAN retrieval would be selectable using `--skb-sections` (can be on by default).
- In addition we should parse the header and store its fields in individual fields (id, pcp, dei) for easier processing in Rust, Python or when using the JSON data.

WDYT (+ @vlrpl @amorenoz)?

retis/src/module/skb/bpf/include/if_vlan.h
andreaskaris: done thanks

retis/src/module/skb/bpf/include/if_vlan.h
andreaskaris: done

docs/collectors/skb.md
andreaskaris: done

andreaskaris:
> A few comments but this looks good, thanks for taking the comments into account!
> 
> IMO this looks good as-is, but I thought a bit more and have a broader question (see below). Sorry this only comes now, but don't feel obligated to do the changes discussed there; I can do a follow-up before we release the feature. As you prefer :)
> 
> I'm wondering if the VLAN data should instead live in its own section in the skb event instead of the meta section, like ARP and others. Using the meta section was logical initially but we're now not only focusing on the skb metadata.
> 
>     * This would align with the way other protocols are handled.
> 
>     * VLAN retrieval would be selectable using `--skb-sections` (can be on by default).
> 
>     * In addition we should parse the header and store its fields in individual fields (id, pcp, dei) for easier processing in Rust, Python or when using the JSON data.
> 
> 
> WDYT (+ @vlrpl @amorenoz)?

I think I got it more or less the way that you wanted it to be with its own SkbVlanEvent. Please check the FIXME comments in if_vlan.h and skb_hook.bpf.c. I set `pub vlan: Option<SkbVlanEvent>,` in skb.rs, but  the C part initializes the struct to all 0s (which is logical). So what I end up with is `vlan (id 0 prio 0)` when no VLAN is present. Any idea for how to take advantage of the Option?

Ahh ... I'm seeing now that unmarshal_dev returns Result<Option<SkbDevEvent>> so I can prob simply follow that logic on the Rust side and if all fields are 0 return a None?

atenart:
> I think I got it more or less the way that you wanted it to be with its own SkbVlanEvent.

Yes, that looks great! That was quick :smile: 

> Please check the FIXME comments in if_vlan.h and skb_hook.bpf.c. I set `pub vlan: Option<SkbVlanEvent>,` in skb.rs, but the C part initializes the struct to all 0s (which is logical). So what I end up with is `vlan (id 0 prio 0)` when no VLAN is present. Any idea for how to take advantage of the Option?

IMO the fix should be in the BPF side: you can retrieve the tci using the existing logic and set a flag when either of the method succeed. Then only when that flag is set you can get the VLAN event section and call `set_skb_vlan_event`.

Something like:

```c
if (!__vlan_hwaccel_get_tag(skb, &vlan_tci)) {
        is_vlan = true;
        is_accel = true;
 } else if (!__vlan_get_tag(skb, &vlan_tci)) {
        is_vlan = true;
}

if (is_vlan) {
        e = get_event_section(...);
        if (!e) ...

        set_skb_vlan_event(e, vlan_tci, is_accel);
}
```

Another unrelated suggestion: using CO-RE it's possible to check if a field is valid in a struct, with that we can follow upstream closer when checking if the vlan accel data is there. This has a slight advantage for older kernels (c9s included) to be more aligned with the check. It shouldn't change anything for recent kernels.

```diff
diff --git a/retis/src/core/probe/kernel/bpf/include/compat.h b/retis/src/core/probe/kernel/bpf/include/compat.h
index a839365d8729..8a5beb3019d0 100644
--- a/retis/src/core/probe/kernel/bpf/include/compat.h
+++ b/retis/src/core/probe/kernel/bpf/include/compat.h
@@ -20,4 +20,8 @@ struct nft_rule_dp_last___6_4_0 {
        const struct nft_chain *chain;
 } __attribute__((preserve_access_index));
 
+struct sk_buff___6_1_0 {
+       u8 vlan_present:1;
+} __attribute__((preserve_access_index));
+
 #endif /* __CORE_PROBE_KERNEL_BPF_COMPAT__ */
diff --git a/retis/src/module/skb/bpf/include/if_vlan.h b/retis/src/module/skb/bpf/include/if_vlan.h
index bf3af6e406f7..1c25cead909e 100644
--- a/retis/src/module/skb/bpf/include/if_vlan.h
+++ b/retis/src/module/skb/bpf/include/if_vlan.h
@@ -8,13 +8,12 @@
 #include <bpf/bpf_endian.h>
 
 #include <common.h>
+#include <compat.h>
 
 #define        ENODATA         61
 #define ETH_P_8021Q    0x8100
 #define ETH_P_8021AD   0x88A8
 
-#define vlan_tag_present(vlan_proto, vlan_tci) (vlan_proto != 0 || vlan_tci != 0)
-
 #define set_skb_vlan_event(e, vlan_tci, vlan_accel)    { \
        e->pcp = (vlan_tci & 0xe000) >> 13; \
        e->dei = (vlan_tci & 0x1000) >> 12; \
@@ -29,12 +28,18 @@ struct skb_vlan_event {
        u8 acceleration;
 } __binding;
 
+static __always_inline bool vlan_tag_present(const struct sk_buff *skb)
+{
+       struct sk_buff___6_1_0 *skb_61 = (struct sk_buff___6_1_0 *)skb;
+
+       if (bpf_core_field_exists(skb_61->vlan_present))
+               return BPF_CORE_READ_BITFIELD_PROBED(skb_61, vlan_present);
+
+       return BPF_CORE_READ(skb, vlan_all);
+}
+
 /**
  * __vlan_hwaccel_get_tag - get the VLAN ID that is in @skb->cb[]
- * The kernel uses skb_vlan_tag_present which either relies on vlan_present or
- * on vlan_all depending on the kernel version (see commit 354259fa73e2aac92ae5e19522adb69a92c15b49).
- * Instead, we use vlan_present = skb->vlan_proto != 0 || skb->vlan_tci != 0 to be compatible
- * with all kernel versions.
  * @skb: skbuff to query
  * @vlan_tci: buffer to store value
  *
@@ -43,11 +48,8 @@ struct skb_vlan_event {
 static inline int __vlan_hwaccel_get_tag(const struct sk_buff *skb,
                                         u16 *vlan_tci)
 {
-       u16 vlan_proto = BPF_CORE_READ(skb, vlan_proto);
-       u16 temp_vlan_tci = BPF_CORE_READ(skb, vlan_tci);
-
-       if (vlan_tag_present(vlan_proto, temp_vlan_tci)) {
-               *vlan_tci = temp_vlan_tci;
+       if (vlan_tag_present(skb)) {
+               *vlan_tci = BPF_CORE_READ(skb, vlan_tci);
                return 0;
        } else {
                *vlan_tci = 0;
```

I'm not sure if this has an actual practical advantage so feel free to ignore.

IMO (with the FIXME fixed) the PR looks good, you can squash the fixes into the commits when pushing the next version.

andreaskaris:
O.k. I integrated both of your suggestions, thanks so much! I actually originally had something like that in mind ( checking if field is present with CO-RE) but I had no clue how :)

2 observations: 

i) when I directly run `make gen-bindings` on my laptop against `main`, I get the following error message and `retis/src/bindings/ovs_common_uapi.rs` is not built correctly:
```
[BINDINGS]   processing /home/akaris/development/retis/retis/src/module/ovs/bpf/include/ovs_common.h ...
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:86:90: error: unknown type name '__u64'
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:110:49: error: unknown type name '__u32'
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:122:23: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:122:8: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:122:14: error: function cannot return function type 'int (void)'
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:122:8: error: illegal initializer (only variables can be initialized)
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:185:57: error: unknown type name '__u32'
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:201:23: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:201:8: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:201:14: error: function cannot return function type 'int (void)'
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:201:8: error: illegal initializer (only variables can be initialized)
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:214:38: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:214:23: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:214:29: error: function cannot return function type 'int (void)'
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:214:23: error: illegal initializer (only variables can be initialized)
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:235:66: error: unknown type name '__u32'
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:235:98: error: unknown type name '__u32'
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:235:109: error: unknown type name '__u64'
/home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:264:66: error: unknown type name '__u32'
fatal error: too many errors emitted, stopping now [-ferror-limit=]
```
I hit the same on my branch, so after generating the bindings, I simply stashed the changes to `ovs_common_uapi.rs`. It's probably a dependency issue on my side, but it not, something worth investigating / fixing?

ii) The VLAN priority is not always fetched. I can't explain why, but I just brought up 2 VMs fedora1 and fedora2. I pinged from fedora1 VLAN 123 to fedora2 VLAN 123 after setting:
```
ip link set enp1s0.123 type vlan egress 0:0 1:1 2:2 3:3 4:4 5:5 6:6 7:7
iptables -t mangle -I OUTPUT -p icmp -j CLASSIFY --set-class 0:6
```
And retis correctly showed the VLAN tag, but not the PCP. It's not a code problem, the vlan_tci field simply didn't have it (I printed it out in binary). Then, I added the same mapping to fedora2 and pinged from fedora2 to fedora1 and retis correctly showed the VLAN ID and PCP. On top of this, from that point on, whenever I pinged from fedora1 to fedora2, I could see both VLAN and PCP as well.
I wonder if that's some optimization on the kernel side, etc.... maybe worth investigating

andreaskaris:
Also, I still think this needs unit tests?

retis/src/module/skb/bpf/include/if_vlan.h
```diff
@@ -0,0 +1,111 @@
...
+
+       return BPF_CORE_READ(skb, vlan_all);
+}
+
+/**
+ * __vlan_hwaccel_get_tag - get the VLAN ID that is in @skb->cb[]
+ * The kernel uses skb_vlan_tag_present which either relies on vlan_present or
+ * on vlan_all depending on the kernel version (see commit 354259fa73e2aac92ae5e19522adb69a92c15b49).
+ * Instead, we use vlan_present = skb->vlan_proto != 0 || skb->vlan_tci != 0 to be compatible
+ * with all kernel versions.
```
atenart, line 46: This is not true anymore.

atenart:
> 2 observations:
> 
> i) when I directly run `make gen-bindings` on my laptop against `main`, I get the following error message and `retis/src/bindings/ovs_common_uapi.rs` is not built correctly:
> 
> ```
> [BINDINGS]   processing /home/akaris/development/retis/retis/src/module/ovs/bpf/include/ovs_common.h ...
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:86:90: error: unknown type name '__u64'
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:110:49: error: unknown type name '__u32'
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:122:23: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:122:8: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:122:14: error: function cannot return function type 'int (void)'
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:122:8: error: illegal initializer (only variables can be initialized)
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:185:57: error: unknown type name '__u32'
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:201:23: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:201:8: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:201:14: error: function cannot return function type 'int (void)'
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:201:8: error: illegal initializer (only variables can be initialized)
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:214:38: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:214:23: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:214:29: error: function cannot return function type 'int (void)'
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:214:23: error: illegal initializer (only variables can be initialized)
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:235:66: error: unknown type name '__u32'
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:235:98: error: unknown type name '__u32'
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:235:109: error: unknown type name '__u64'
> /home/akaris/development/retis/retis/src/.out/bpf/bpf_helper_defs.h:264:66: error: unknown type name '__u32'
> fatal error: too many errors emitted, stopping now [-ferror-limit=]
> ```
> 
> I hit the same on my branch, so after generating the bindings, I simply stashed the changes to `ovs_common_uapi.rs`. It's probably a dependency issue on my side, but it not, something worth investigating / fixing?

I saw similar issues today as well but haven't looked closely. Could you open an issue about the above?

> ii) The VLAN priority is not always fetched. I can't explain why, but I just brought up 2 VMs fedora1 and fedora2. I pinged from fedora1 VLAN 123 to fedora2 VLAN 123 after setting:
> 
> ```
> ip link set enp1s0.123 type vlan egress 0:0 1:1 2:2 3:3 4:4 5:5 6:6 7:7
> iptables -t mangle -I OUTPUT -p icmp -j CLASSIFY --set-class 0:6
> ```
> 
> And retis correctly showed the VLAN tag, but not the PCP. It's not a code problem, the vlan_tci field simply didn't have it (I printed it out in binary). Then, I added the same mapping to fedora2 and pinged from fedora2 to fedora1 and retis correctly showed the VLAN ID and PCP. On top of this, from that point on, whenever I pinged from fedora1 to fedora2, I could see both VLAN and PCP as well. I wonder if that's some optimization on the kernel side, etc.... maybe worth investigating

That's a bit weird. I looked at the code added in this PR and things look good. I also made some tests using two netns adding the priority on one side only and that worked as expected. Maybe you encountered some glitch?

> Also, I still think this needs unit tests?

That would be awesome! You can add one in the runtime CI; see `tests/` for their definitions and how those runs in `.cirrus.yml` (see `&functional_test`) but I'm afraid things are not well documented in there. Best thing would be to do this as part of a follow-up series so we can merge this one.

retis/src/module/skb/bpf/include/if_vlan.h
andreaskaris: done

andreaskaris:
Test ticket: https://github.com/retis-org/retis/issues/479
gen-bindings issue: https://github.com/retis-org/retis/issues/478

andreaskaris:
if you're o.k., we can merge this after tests passed so that you can get your other PR in. I'll work on the tests in a separate PR

atenart:
I'm sorry I have an additional suggestion, I should test things in one go to avoid those ping-pong...

Retis is quite particular in the sense that we can see packets almost everywhere in the stack and we can't rely on assumptions, such as what's available in `skb->data` or if a specific offset is valid or not. This PR relies on `skb->data` to get the VLAN information in the non-accel case and this won't work in some situations. We however have an helper to get the mac offset which works in a variety of cases, depending on where is packet is seen. We can use it here to get the VLAN information from more functions:

```diff
diff --git a/retis/src/core/probe/kernel/bpf/include/helpers.h b/retis/src/core/probe/kernel/bpf/include/helpers.h
index f66f0545bf7b..af4369a5a878 100644
--- a/retis/src/core/probe/kernel/bpf/include/helpers.h
+++ b/retis/src/core/probe/kernel/bpf/include/helpers.h
@@ -74,7 +74,7 @@ static __always_inline bool is_network_valid(u16 network)
 }
 #define is_transport_valid is_network_valid
 
-static __always_inline bool is_mac_data_valid(struct sk_buff *skb)
+static __always_inline bool is_mac_data_valid(const struct sk_buff *skb)
 {
        u16 mac, network;
 
@@ -86,7 +86,7 @@ static __always_inline bool is_mac_data_valid(struct sk_buff *skb)
                 BPF_CORE_READ(skb, mac_len) == 0);
 }
 
-static __always_inline bool is_network_data_valid(struct sk_buff *skb)
+static __always_inline bool is_network_data_valid(const struct sk_buff *skb)
 {
        u16 mac, network;
 
diff --git a/retis/src/module/skb/bpf/include/if_vlan.h b/retis/src/module/skb/bpf/include/if_vlan.h
index 3bfc53ec5528..2d5da0bb7d90 100644
--- a/retis/src/module/skb/bpf/include/if_vlan.h
+++ b/retis/src/module/skb/bpf/include/if_vlan.h
@@ -83,6 +83,14 @@ static inline bool eth_type_vlan(__be16 ethertype)
  */
 static inline struct vlan_ethhdr *skb_vlan_eth_hdr(const struct sk_buff *skb)
 {
+       int mac = BPF_CORE_READ(skb, mac_header);
+
+       if (is_mac_data_valid(skb)) {
+               unsigned char *head = BPF_CORE_READ(skb, head);
+
+               return (struct vlan_ethhdr *)(head + mac);
+       }
+
        return (struct vlan_ethhdr *) BPF_CORE_READ(skb, data);
 }
```

Using a simple test with two netns I went from:

```
2227112881080 (3) [ping] 1932 [tp] net:net_dev_start_xmit #2068a45abb8ffff951841cc2580 (skb ffff9518428c2500) n 0
  10.0.42.1 > 10.0.42.2 ttl 64 tos 0x0 id 47962 off 0 [DF] len 84 proto ICMP (1) type 8 code 0
  ↳ 2227112883772 (3) [ping] 1932 [tp] net:net_dev_start_xmit #2068a45abb8ffff951841cc2580 (skb ffff9518428c2500) n 1
      fe:36:e1:2c:4e:04 > b6:77:d7:fc:e2:c9 ethertype 802.1Q (0x8100) vlan (id 42 prio 6)
  ↳ 2227112885321 (3) [ping] 1932 [tp] net:netif_receive_skb #2068a45abb8ffff951841cc2580 (skb ffff9518428c2500) n 2
      fe:36:e1:2c:4e:04 > b6:77:d7:fc:e2:c9 ethertype 802.1Q (0x8100)
```

To:

```
1482252737438 (1) [ping] 1821 [tp] net:net_dev_start_xmit #1591d25db9effff95184260d900 (skb ffff95184d275200) n 0
  10.0.42.1 > 10.0.42.2 ttl 64 tos 0x0 id 53712 off 0 [DF] len 84 proto ICMP (1) type 8 code 0
  ↳ 1482252739859 (1) [ping] 1821 [tp] net:net_dev_start_xmit #1591d25db9effff95184260d900 (skb ffff95184d275200) n 1
      06:86:ba:2f:d5:2a > 3a:cc:77:1d:b6:3f ethertype 802.1Q (0x8100) vlan (id 42 prio 6)
  ↳ 1482252741768 (1) [ping] 1821 [tp] net:netif_receive_skb #1591d25db9effff95184260d900 (skb ffff95184d275200) n 2
      06:86:ba:2f:d5:2a > 3a:cc:77:1d:b6:3f ethertype 802.1Q (0x8100) vlan (id 42 prio 6)
```

See the addition vlan info in `net:netif_receive_skb`.

(There are other things to improve, like correctly parsing the IP data when a VLAN is in the data part; but that has nothing to do with this PR).

atenart (approved):
LGTM, thanks for adding support for retrieving VLAN information and following-up with the different reviews!

PR #471, opened 2025-01-10
collector.rs: count and display processed events

vlrpl:
Small UX enhancement providing feedback on termination showing the number of processed events.

More and fancy counters could be chased (raw events received but not stored and so forth) but that would be out of scope.

### DIFF
--- a/retis/src/collect/collector.rs
+++ b/retis/src/collect/collector.rs
@@ -523,6 +523,7 @@ impl Collectors {
             });
         }
 
+        let (mut iccount, mut eccount) = (0, 0);
         let mut probe_stack = ProbeStack::new(
             collect.stack,
             self.probes.runtime_mut()?.attached_probes(),
@@ -537,6 +538,7 @@ impl Collectors {
                 printers
                     .iter_mut()
                     .try_for_each(|p| p.process_one(&event))?;
+                iccount += 1;
             }
 
             // Then get raw events, if any.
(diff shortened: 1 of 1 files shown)


retis/src/collect/collector.rs
```diff
@@ -549,12 +551,16 @@ impl Collectors {
...
                         .try_for_each(|p| p.process_one(&event))?;
+                    eccount += 1;
                 }
                 Timeout => continue,
             }
         }
 
         printers.iter_mut().try_for_each(|p| p.flush())?;
+        info!("{} raw event(s) processed", eccount);
+        info!("{} internal event(s) processed", iccount);
```
atenart, line 562: Wouldn't just out the total number of events good enough? I'm thinking raw vs internal events is an implementation detail, not a user facing one. If getting the detail is important this could be a debug log level.

vlrpl: It was initially like this (and before it was just raw), but raw events might be more significant from the user's standpoint, at least in terms of pure kernel-side event collection. That's the reason for the separation.
What I mean is that if your collection filter is too narrow, it's more intuitive to see `0 raw event(s) ...` instead of `X event(s) ...`  (whatever x is now or could be in the future) as an aggregate.
They were both at the INFO level because:

- the idea is kind of having the same behavior as the most popular packet capture tools have
- the sum of both should match with what you get

Anyway, I'm not sure if you are suggesting moving the `iccount` to `debug!()` or both :)

atenart: Makes sense. Yes, I think I would move the `iccount` to `debug!` and rename `raw event(s)` to `event(s)`.

vlrpl: > Makes sense. Yes, I think I would move the `iccount` to `debug!` and rename `raw event(s)` to `event(s)`.

I agree this is the best option at this point. Just checking we were on the same page, thanks :)

atenart (approved):
LGTM, thanks!

PR #472, opened 2025-01-15
Added RETIS_IMAGE environment variable for alternate image location

laddp:
Added an environment variable in tools/retis_in_container.sh to point to an alternate location for the retis container image for use by those in disconnected environments.


atenart (approved):
LGTM, thanks!

PR #474, opened 2025-01-17
[RFC] Introduce checkpatch action

vlrpl:
Reusing an existing GH action doesn't seem bad, but Cirrus would be fairly simple.

If Cirrus or in general we reach a consensus around not reusing the action used here, we have to decide whether we have to redistribute `checkpatch.pl` (and `spelling.txt`)

Included a couple of patches that intentionally introduce some coding style issues to show the pros and cons of the action.

Configs have been left intentionally loose, feel free to propose what you think could make sense.

### DIFF
--- a/.checkpatch.conf
+++ b/.checkpatch.conf
@@ -1,0 +1,6 @@
+--no-tree
+
+--ignore EMAIL_SUBJECT
+--ignore FILE_PATH_CHANGES
+--ignore SPDX_LICENSE_TAG
+--ignore VOLATILE
--- a/.github/workflows/checkpatch.yaml
+++ b/.github/workflows/checkpatch.yaml
@@ -1,0 +1,14 @@
+name: eBPF coding style check
+on: [pull_request]
+jobs:
+  review:
+    name: checkpatch run
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: $((${{ github.event.pull_request.commits }} + 1))
+    - name: Run checkpatch
+      uses: webispy/checkpatch-action@v9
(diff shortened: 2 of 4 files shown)


atenart:
While I like not having to distribute our own version of `checkpatch` (as we won't have to maintain and sync it), it has a major drawback: there's no way to check the coding style is correct locally before opening a PR. If we compare to the Rust & Python coding styles, they can be checked locally (although doing so for the Python part is not that straightforward...[1]).

[1] An addition would be to consolidate the various checks into a make target.

vlrpl:
We agreed to vendor `checkpatch.pl` after multiple different attempts.
The PR does not fix all the current style issues. This can be done in a later follow-up (or can be done incrementally when contributors modify affected areas/files).
The PR also includes two patches with style issues; once the PR gets approved, they will be removed.

vlrpl:
> Can we make the checkpatch run a make target so it's easier to run locally before pushing a PR?
> 

Moving the Cirrus Logic in the Makefile has the side effect of causing the build to fail due to style issues:
```
[...]
make: *** [Makefile:183: check-ebpf] Error 1
```
which was the initial reason for leaving it out. However, reconsidering this is not a bad idea, especially since we want to be strict about this.

> Can we put the checkpatch files under `tools/checkpatch/` to group them in a single directory?

done

tools/check.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+make -C../ analyze
```
atenart: Leftover?

Makefile
```diff
@@ -178,6 +178,20 @@ endef
 
 $(foreach tgt,check clippy,$(eval $(call analyzer_tmpl,$(tgt))))
 
+check-ebpf:
+	$(call out_console,CHECKPATCH,checking eBPF coding style ...)
+	base_hash=$$(git merge-base $${BASE_COMMIT:-main} HEAD); \
+	COMMIT_RANGE=$$(git rev-list $${base_hash}..HEAD --reverse); \
```
atenart: \+ `--no-merges` ?

vlrpl: makes sense

tools/check.sh
vlrpl: yes, sorry, unrelated

atenart (approved):
LGTM, nice addition, thanks!

PR #481, opened 2025-01-22
collector: skb: handle VLAN in the payload

atenart:
Collecting packets with one or more VLAN header(s) in their payload resulted in packets we could not parse. Fix this by skipping the VLAN headers if found. The VLAN information is reported in the event but is coming from the BPF part directly (to support h/w acceleration).


atenart:
Note that this does not address some remaining questions about VLAN support (how to report and properly handle QinQ mainly), they will be handled separately.

atenart:
If we go with #485 this can be dropped.

atenart:
As we'll go with #485, this is no longer relevant and can be closed.

PR #483, opened 2025-01-23
collect: use a default set of probes when none are explicitly given

atenart:
Opening this to start the discussion. This could improve UX but there are some caveats.

Pros:
- This makes `retis collect` to be really useful and not only for looking at dropped packets. It can serve as a good starting point.
- We can add even more logic later on, to try being smart by default.

Cons:
- The way to disable the automatic set of probes is to explicitly define probes. This is fine for most usages but there's one drawback: when one does not want probes (at least explicitly defined), eg. `retis collect -c skb-drop`.
- The current solution here is for such uses cases to define a profile, as using a profile disables automatic probing (and regardless of this specific issue doing so makes sense IMO).
- This could also be done with a `--no-default-probes` option, but that isn't great in terms of UX IMO.

IMO adding a default set of probes is better than the current situation, but I'm not 100% happy with the "see drops only" situation. I guess profiles make sense for those use cases (as it shift the need for profiles for a starting point or any useful collection to specific cases), but maybe there's an even better way.

Thoughts?


amorenoz:
> Opening this to start the discussion. This could improve UX but there are some caveats.
> 
> Pros:
> 
>     * This makes `retis collect` to be really useful and not only for looking at dropped packets. It can serve as a good starting point.
> 
>     * We can add even more logic later on, to try being smart by default.
> 
> 
> Cons:
> 
>     * The way to disable the automatic set of probes is to explicitly define probes. This is fine for most usages but there's one drawback: when one does not want probes (at least explicitly defined), eg. `retis collect -c skb-drop`.
> 

I would enable this as an extension of the `auto-mode`, i.e: only enabled when:
- No profiles are given
- No explicit probes are specified
- No explicit collectors are specified

That way, `retis collect -c skb-drop` or `retis collect -c ovs` should only enable the implicit probes from skb-drop and ovs respectively.


>     * The current solution here is for such uses cases to define a profile, as using a profile disables automatic probing (and regardless of this specific issue doing so makes sense IMO).
> 
>     * This could also be done with a `--no-default-probes` option, but that isn't great in terms of UX IMO.
> 
> 
> IMO adding a default set of probes is better than the current situation, but I'm not 100% happy with the "see drops only" situation. I guess profiles make sense for those use cases (as it shift the need for profiles for a starting point or any useful collection to specific cases), but maybe there's an even better way.
> 
> Thoughts?

Other options that have crossed my mind:
- Define a profile that is the "default" one, allowing users to change which one is the default (via a config file?). We could set the current "generic" profile as the default and allow users to change that. The problem here would be that it's not really intuitive that `retis collect` enables a profile but `retis collect -p tp:net:netif_receive_skb` does not... 
- Give retis some interactivity and ask the user to select a profile.

atenart:
> I would enable this as an extension of the `auto-mode`, i.e: only enabled when:
> 
> * No profiles are given

OK.

> * No explicit probes are specified

OK.

> * No explicit collectors are specified

This would only work for collectors installing probes. So enabling `ovs` or `skb-drop` would do fine, but what about `skb,ct`?

Of course we could argue doing manual things requires manual probes and that auto-mode is only on when doing most things automatically.

> Other options that have crossed my mind:
> 
> * Define a profile that is the "default" one, allowing users to change which one is the default (via a config file?). We could set the current "generic" profile as the default and allow users to change that. The problem here would be that it's not really intuitive that `retis collect` enables a profile but `retis collect -p tp:net:netif_receive_skb` does not...

I would prefer not to use profiles for this, and not to use a default one. This is because doing so in Retis directly makes it possible to have some logic and do the right thing based on many factors (current args, kernel version, available probes, etc). And not using a default profile makes the auto-mode useful and allows using Retis w/o a profile.

> * Give retis some interactivity and ask the user to select a profile.

I would also prefer not to require user input. Either the user wants a predefined set of arguments, or we try to do the right thing. This would go against that last part.

atenart:
> > I would enable this as an extension of the `auto-mode`, i.e: only enabled when:
> > 
> > * No profiles are given
> 
> OK.
> 
> > * No explicit probes are specified
> 
> OK.
> 
> > * No explicit collectors are specified
> 
> This would only work for collectors installing probes. So enabling `ovs` or `skb-drop` would do fine, but what about `skb,ct`?
> 
> Of course we could argue doing manual things requires manual probes and that auto-mode is only on when doing most things automatically.

I thought about this again, and I think your proposal makes sense. Either we do things automatically or we don't.

One that front, `auto_mode` is currently only defined by the `--collectors` argument not being set. Should we extend it (and its implications) to the above three conditions?

amorenoz:
> > > I would enable this as an extension of the `auto-mode`, i.e: only enabled when:
> > > 
> > > * No profiles are given
> > 
> > 
> > OK.
> > > * No explicit probes are specified
> > 
> > 
> > OK.
> > > * No explicit collectors are specified
> > 
> > 
> > This would only work for collectors installing probes. So enabling `ovs` or `skb-drop` would do fine, but what about `skb,ct`?
> > Of course we could argue doing manual things requires manual probes and that auto-mode is only on when doing most things automatically.
> 

The existing `auto-mode` already enables all collectors, so the case is: "I (user) want to use the probes retis usually installs in auto mode but _only_ want to enable skb probe, please don't make me type them". My answer would be:
1 ) What you get with auto mode is _more_ than what you want, just ignore the rest.
2 ) If it's _really_ needed, then we can create a profile which contains only these probes and now the user can do `retis -p auto collect -c skb` which is pretty good as well.


> I thought about this again, and I think your proposal makes sense. Either we do things automatically or we don't.
> 
> One that front, `auto_mode` is currently only defined by the `--collectors` argument not being set. Should we extend it (and its implications) to the above three conditions?

What are the implications?

atenart:
> > > > * No explicit collectors are specified
> > > 
> > > 
> > > This would only work for collectors installing probes. So enabling `ovs` or `skb-drop` would do fine, but what about `skb,ct`?
> > > Of course we could argue doing manual things requires manual probes and that auto-mode is only on when doing most things automatically.
> 
> The existing `auto-mode` already enables all collectors, so the case is: "I (user) want to use the probes retis usually installs in auto mode but _only_ want to enable skb probe, please don't make me type them". My answer would be: 1 ) What you get with auto mode is _more_ than what you want, just ignore the rest. 2 ) If it's _really_ needed, then we can create a profile which contains only these probes and now the user can do `retis -p auto collect -c skb` which is pretty good as well.

Makes sense.

> > I thought about this again, and I think your proposal makes sense. Either we do things automatically or we don't.
> > One that front, `auto_mode` is currently only defined by the `--collectors` argument not being set. Should we extend it (and its implications) to the above three conditions?
> 
> What are the implications?

Enabling all collectors by default and letting them opt-out if prerequisites aren't met. Currently this is only controlled by the `--collectors` flag not being set. We could force setting `--collectors` if a profile is used for example. For `--probe` that doesn't make much sense IMO. This would make sure profiles are stable over time.

In any cases we should distinguish `auto_mode` from `auto_collectors` in the code (the later doesn't exist yet) as it's not strictly the same thing (only one implies the other).

My understanding is, except for the small discussion above, we agree on adding probes automatically if no collector, probe and profile is set (that is the current PR behavior). I can thus work on making this PR ready:
- Consolidate `auto_mode` so it's not misunderstood in the code (as explained above).
- Update the CI regarding this.
- Update the doc.
- Include a `--profile` check in the auto collectors mode (iif the above is agreed on, otherwise skip this).

atenart:
I did all I had in mind for this PR, except for the documentation update. But let's see if the current state looks good before I do so.

retis/src/collect/collect.rs
```diff
@@ -317,17 +317,37 @@ impl Collectors {
...
+                // By default dump packets after the device in ingress and
+                // before the device in egress; like AF_PACKET utilities.
+                vec![
+                    Probe::raw_tracepoint(Symbol::from_name("net:netif_receive_skb")?)?,
+                    Probe::raw_tracepoint(Symbol::from_name("net:net_dev_start_xmit")?)?,
+                ]
+            };
+
+            info!(
+                "No probe given: using {}",
```
amorenoz, line 341: nit: probe(s)

retis/src/collect/collect.rs
```diff
@@ -317,17 +317,37 @@ impl Collectors {
...
-                .register_probe(Probe::raw_tracepoint(Symbol::from_name("skb:kfree_skb")?)?)?;
+        // If auto_mode is on and no probe is given, find the right set
+        // automagically.
+        if self.auto_mode && collect.probes.is_empty() {
+            let mut probes = if collect.probe_stack {
+                // If --probe-stack is used, use skb:consume_skb & skb:kfree_skb
+                // as a starting point (these should capture most if not all of
+                // the packets and help moving up the stack).
+                vec![
+                    Probe::raw_tracepoint(Symbol::from_name("skb:consume_skb")?)?,
```
amorenoz, line 328: This is triggered *very* often for non-critical packets. Isn't this going to add a huge amount of noise?

retis/src/collect/collect.rs
```diff
@@ -317,17 +317,37 @@ impl Collectors {
...
+                    Probe::raw_tracepoint(Symbol::from_name("skb:kfree_skb")?)?,
```
amorenoz, line 329: wouldn't `skb_drop` collector enable kfree_skb anyways?

amorenoz (commented):
A couple of comments but in general I think I'm OK with the PR. Let's update the documentation. Also, maybe we should also print the default probes in the `--help`?

retis/src/collect/collect.rs
atenart: (This is not introduced by this PR)

That is if `--probe-stack` is set and no probe is given. Then the idea is to match (a filter is required for `--probe-stack`) packets and rewind their journey. We definitively want to catch non-drop traffic here.

retis/src/collect/collect.rs
atenart: That's right. I think this should still be explicitly set here:
- It makes it intentional as part as the default probe set for `--probe-stack`.
- The two probes (here & `skb-drop` collector) do not have the same intent.
- The two probes will be merged by the manager before being installed.

atenart:
- Updated the help, README, gif demo and the docs.
- Fixed new Clippy warnings.

amorenoz (approved):
LGTM

PR #487, opened 2025-01-30
collector: ct: fix truncation of zone id

igsilya:
Zone id is 16-bit, but for some reason is casted to u8 on read. This makes retis report incorrect values.

Fixes: e883af89e368 ("module: add conntrack module")

### DIFF
--- a/retis/src/collect/collector/ct/bpf/ct.bpf.c
+++ b/retis/src/collect/collector/ct/bpf/ct.bpf.c
@@ -128,7 +128,7 @@ static __always_inline int process_nf_conn(struct ct_event *e,
 		if (zone_dir & NF_CT_ZONE_DIR_REPL)
 			e->flags |= RETIS_CT_DIR_REPLY;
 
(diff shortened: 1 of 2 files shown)


vlrpl:
Thanks, @igsilya  for spotting this, for the fix, and for the test.
Also, the patch lgtm.
There's a problem with (apparently) a previous PR on C8S (ovs tests are not run there).
I'll quickly double-check and be back on this.

vlrpl:
It seems `vlan_all` is missing in older kernels (c8s [0]) and `BPF_CORE_READ()` is not enough.
Didn't dig into it, but this should be protected with `bpf_core_field_exists()` and emit this relo type `BPF_FIELD_EXISTS`.
@andreaskaris, if you have spare cycles, would you like to give it a stab?

[0] https://cirrus-ci.com/task/5509164367085568

vlrpl (approved):
the patch lgtm, thanks!

PR #491, opened 2025-02-07
Extend ct and ovs events 

vlrpl:
Extend ct and ovs events respectively with `ct->status`, and action address while executing.

Edit: I decided to drop the ovs patch for the time being as we have other work pending in that area.
Can be easily resurrected in a later moment.

### DIFF
--- a/retis-events/src/ct.rs
+++ b/retis-events/src/ct.rs
@@ -140,6 +140,8 @@ pub struct CtConnEvent {
     pub mark: Option<u32>,
     /// Connection tracking labels.
     pub labels: Option<U128>,
+    /// nf_conn status (ct->status).
+    pub ct_status: u64,
 }
 
 impl EventFmt for CtEvent {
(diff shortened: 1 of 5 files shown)


retis-events/src/ct.rs
```diff
@@ -168,6 +170,8 @@ impl EventFmt for CtEvent {
 
 impl CtEvent {
     fn format_conn(conn: &CtConnEvent, f: &mut Formatter) -> fmt::Result {
+        write!(f, "status {:x} ", conn.ct_status)?;
```
atenart, line 173: nit: `:#x`

retis-events/src/ovs.rs
```diff
@@ -367,6 +369,8 @@ impl EventFmt for ActionEvent {
             None => write!(f, " unspec")?,
         }
 
+        write!(f, " nla {:x}", self.action_address)?;
```
atenart, line 372: nit: `:#x`

A suggestion: precision like the above are sometimes put between parenthesis in our events, eg. `(nla {})`.

atenart (commented):
The action address seems like an nla object address, which differs a bit from what we usually retrieve. Is that address also available in data coming from external sources and helpful for cross checking things?

Please add the new fields to the documentation.

retis-events/src/ovs.rs
vlrpl: if we want to stay consistent (with e.g. skb) parentheses are wanted, but the prefixed hex is not what we want.

atenart (approved):
LGTM, thanks!

atenart:
Rawhide CI failure is known and unrelated.

PR #492, opened 2025-02-07
move if_vlan.h in the correct path

vlrpl:
`mv retis/src/module/skb/bpf/include/if_vlan.h retis/src/collect/collector/skb/bpf/include/if_vlan.h`


atenart (approved):
Nice finding, thanks for fixing this!

PR #494, opened 2025-02-14
Filtering in subprog

vlrpl:
More info in the commit messages


retis/src/core/filters/filters.rs
```diff
@@ -60,81 +60,161 @@ pub(crate) fn register_filter_handler(
...
-            (
-                libbpf_sys::bpf_program__insns(prog),
-                libbpf_sys::bpf_program__insn_cnt(prog),
-            )
-        };
+        // 0xBAD2310 is used to mark instructions that failed
+        // relocation:
+        // 1: (85) call unknown#195896080
+        // invalid func unknown#195896080
+        // this is skipped as this is intended and handled.
```
atenart, line 133: Can you add this is done specifically by `libbpf` and not us?

retis/src/core/filters/filters.rs
```diff
@@ -60,81 +60,161 @@ pub(crate) fn register_filter_handler(
...
+fn copy_ptr_to_vec<T>(ptr: *mut T, count: usize) -> Vec<T> {
+    let mut vec = Vec::with_capacity(count);
+    unsafe {
+        std::ptr::copy(ptr, vec.as_mut_ptr(), count);
+        vec.set_len(count);
+    }
+    vec
+}
+
+// Rewrites the instructions, indexes all the pseduo calls, fixes
```
atenart, line 108: nit: pseudo

retis/src/core/filters/packets/filter.rs
```diff
@@ -49,6 +49,7 @@ impl FilterPacket {
             bail!("Filter exceeds the maximum allowed size.");
         }
 
+        // ebpf_filter.disasm();
```
atenart, line 52: Still there :sweat_smile:

retis/src/core/filters/filters.rs
```diff
@@ -60,81 +60,161 @@ pub(crate) fn register_filter_handler(
...
+        }
+
+        if insn.src_reg() == bpf_sys::BPF_PSEUDO_CALL {
+            pseudo_calls.push(pos);
+            continue;
+        }
+
+        if filter_types.contains(&imm) {
+            placeholder_calls.push((pos, imm));
+        }
```
atenart, line 145: Not that it's an actual issue, but logically there's nothing preventing us from having and finding multiple times the same placeholder value.

When looping over placeholder_calls later this would be an issue as there would be potential conflicts.

atenart (changes requested):

retis/src/core/filters/packets/filter.rs
vlrpl: :facepalm:

retis/src/core/filters/filters.rs
vlrpl: "failed relocation" implies it (also the log is from libbpf), but I'm ok with being more explicit.

vlrpl: uhm, tbf the log was not the intended one.
While at it, let's replace that and solve both.

retis/src/core/filters/filters.rs
vlrpl: > Not that it's an actual issue, but logically there's nothing preventing us from having and finding multiple times the same placeholder value.
> 
> When looping over placeholder_calls later this would be an issue as there would be potential conflicts.

This requires the same value to be erroneously (or intentionally) inserted, compiled and so forth.
It's an unrecoverable either way, it's not supposed to work (and I think in both cases will at least fail the load).
That being said, we can check for a previously inserted value in the filter map later reused in the actual injection:

```rust
        let filter = retrieve_filter(placeholder.1);
        let fixup_len = filter.len() as i32 - 1;
        if let Some(_) = filters.insert(placeholder.1, filter) {
            error!("error while fixing up instructions: found duplicate {:#x} at {}", placeholder.1, placeholder.0);
            return -1;
        }
```

atenart: Oh you already did it, thanks!

atenart (approved):
Nice improvements, LGTM, thanks!

retis/src/core/filters/filters.rs
vlrpl: > Oh you already did it, thanks!

yeah, I think I cut (unintentionally) the last line  where I said I agreed with adding the check and that I was going to push it (of course, being ready to change it, if needed). Sorry, I realized only now :)

PR #496, opened 2025-02-23
btf/meta filter: introduce boolean expressions

vlrpl:
- Introduced the new grammar
- parsing is now done using pest (doesn't affect rpm as fedora ships the same version) with better error handling
- the filter no longer relies on maps and pre-defined ebpf programs, but the instructions are dynamically generated and then inlined in the same way packet filter does unifying both types of filters (this was ready since a while waiting for the remaining part to land)
- boolean expressions allow matches on multiple fields. It also supports short circuits meaning A || B doesn't process B if A is true (short circuits are applied at any level)
- doc got updated (no plan to go back to EBNF as pest grammar is just as readable)
- tests no longer check the output but just the correctness of the filter (everything but the instructions generated)

Meta filtering is still coupled with the inspector, but that could be easily lifted when it becomes generic and no longer bound to sk_buff only (requires per-probe filtering as well). Also, it ideally should only depend on generic BTF collections to make it reusable in any context.

Based on #494


atenart (commented):
I mostly looked at the first 2 commits (started looking at the third one but will need to come back at it a few times). The general direction is amazing and I'm not seeing any major blocker preventing this from making it out of RFC. I'd still like to test it a bit before you invest more time; just in case.

I'm submitting the comments on the first commits, in case they can be merged separately (I saw but too late that they are in another PR, but I don't think I can move comments between PRs; sorry).

tools/annotations.py
```diff
@@ -6,7 +6,7 @@
 
 def extract_annotated_types(file_path, annotation, clang_args):
     index = clang.cindex.Index.create()
-    translation_unit = index.parse(file_path, args=clang_args)
+    translation_unit = index.parse(file_path, args=clang_args, options=clang.cindex.TranslationUnit.PARSE_INCOMPLETE | clang.cindex.TranslationUnit.PARSE_SKIP_FUNCTION_BODIES)
```
vlrpl, line 9: this is another leftover. It's part of an investigation I was doing with bindings (but turns out bindgen is more sensitive) and macros that don't always seem to play well in all cases.

retis/src/core/filters/packets/filter.rs
```diff
@@ -49,6 +49,7 @@ impl FilterPacket {
             bail!("Filter exceeds the maximum allowed size.");
         }
 
+        // ebpf_filter.disasm();
```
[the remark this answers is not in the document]
vlrpl: definitely

retis/src/core/filters/packets/bpf/include/packet_filter.h
```diff
@@ -32,49 +28,11 @@ BINDING_DEF(SCRATCH_MEM_SIZE, 4)
  */
 BINDING_DEF(SCRATCH_MEM_START, 16 * SCRATCH_MEM_SIZE + STACK_RESERVED)
 
-#define STACK_SIZE		SCRATCH_MEM_START
-
+#define l2 0xdeadbeef
+#define l3 0xdeadc0de
```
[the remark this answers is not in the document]
vlrpl: In this case, I'm not sure as we'd essentially end up saying we need a define because with preprocessor enums cannot be expanded to literals, right?
With `BINDING_DEF()` probably could make sense to get the point without looking at the macro (it hides the fact it is an enum), but here things are pretty explicit. But if you think it adds value,

retis/src/core/filters/filters.rs
```diff
@@ -60,81 +58,149 @@ pub(crate) fn register_filter_handler(
...
+        std::slice::from_raw_parts(
+            f.as_slice().as_ptr() as *const libbpf_sys::bpf_insn,
+            f.len() / std::mem::size_of::<libbpf_sys::bpf_insn>(),
+        )
+    };
+
+    filter.to_vec()
+}
+
+fn copy_btf_info<T>(ptr: *mut T, count: usize) -> Vec<T> {
```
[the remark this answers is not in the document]
vlrpl: sgtm. btf_info was picked mostly because of where the function is placed/scoped (I cannot foresee someone consuming this from here, but more from helpers paths)

retis/src/core/filters/filters.rs
```diff
@@ -60,81 +58,149 @@ pub(crate) fn register_filter_handler(
...
+    let mut placeholder_calls: Vec<(usize, _)> = Vec::new();
+
+    let (insns, insns_cnt) = unsafe {
+        (
+            libbpf_sys::bpf_program__insns(prog),
+            libbpf_sys::bpf_program__insn_cnt(prog),
+        )
+    };
+
+    let insns = unsafe { std::slice::from_raw_parts(insns, insns_cnt as usize) };
```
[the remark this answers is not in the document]
vlrpl: I totally agree

retis/src/core/filters/filters.rs
```diff
@@ -60,81 +58,149 @@ pub(crate) fn register_filter_handler(
...
-        };
+        }
+    }
 
-        prog_ext.splice(inject_pos..inject_pos + filter.len(), filter.to_vec());
+    // Finally inject the filters
+    // placeholder_calls is sorted, so proceed backward to avoid
+    // further adjustment.
+    for placeholder in placeholder_calls.iter().rev() {
+        let filter = retrieve_filter(placeholder.1);
```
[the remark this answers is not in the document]
vlrpl: The loops are kept intentionally separate because we keep the pseudo calls index, and so forth, and injecting while fixing them up requires further adjustments of the index at every injection (based on whether they appear before, or after the injected code).

Splitting made things clearer to read without a real impact (placeholder_calls is k and can be max 3 with metafiltering).
I guess if we don't want to change the logic, we can store this in a HashMap by code and reuse in the injection loop, WDYT?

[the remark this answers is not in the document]
vlrpl: For the time being, I repushed this and the remaining non-pending suggestions to #494

retis/src/core/filters/packets/bpf/include/packet_filter.h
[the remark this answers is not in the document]
atenart: It's explicit if you know about it :) Up to you, I personally like to describe tricks so older me understands quickly what's going on, but not a blocker.

retis/src/core/filters/meta/filter.rs
```diff
@@ -268,160 +569,558 @@ impl MetaOp {
...
-                if rval_len >= md.len() {
-                    bail!("invalid rval size (max {}).", md.len() - 1);
+            if let Rhs::Str(val) = rval {
+                // lenght including '\0'
+                let rval_len = val.len() + 1;
+                let md = &mut ctx.target.md;
+
+                ctx.target.sz = rval_len
+                    .try_into()
+                    .map_err(|_| anyhow!("invalid rval size (max {})", md.len() - 1))?;
```
atenart, line 598: Isn't `md.len()` 32 and errors of casting if rval_len doesn't fit in an u8?

nit: `target.sz` should probably be `usize`?

retis/src/core/filters/meta/filter.rs
```diff
@@ -268,160 +569,558 @@ impl MetaOp {
...
-    fn emit_target(lmo: &MetaLoad, rval: Rval, cmp_op: MetaCmp) -> Result<MetaOp> {
-        let mut op: MetaOp = MetaOp::new();
-        let top = op.target_ref_mut();
-
-        if lmo.is_ptr() || lmo.nmemb > 0 {
-            if cmp_op != MetaCmp::Eq && cmp_op != MetaCmp::Ne {
+        if ctx.load.is_ptr() || ctx.load.nmemb > 0 {
+            if rel_op != RelOp::Eq && rel_op != RelOp::Ne {
                 bail!(
                     "wrong comparison operator. Only '{}' and '{}' are supported for strings.",
```
atenart, line 585: \+ and pointers

retis/src/core/filters/meta/filter.rs
```diff
@@ -268,160 +569,558 @@ impl MetaOp {
...
+            if let Rhs::Str(val) = rval {
+                // lenght including '\0'
+                let rval_len = val.len() + 1;
+                let md = &mut ctx.target.md;
+
+                ctx.target.sz = rval_len
+                    .try_into()
+                    .map_err(|_| anyhow!("invalid rval size (max {})", md.len() - 1))?;
+                md[..val.len()].copy_from_slice(val.as_bytes());
+                if ctx.load.nmemb == 0 {
```
atenart, line 600: Not sure to get when `nmemb` would be 0 and valid? Also couldn't that trigger OOB loads?

retis/src/core/filters/meta/filter.rs
```diff
@@ -136,118 +129,426 @@ impl MetaLoad {
...
+        })
     }
 
-    fn emit_load(btf: &Btf, r#type: &Type, offt: u32, bfs: u32, mask: u64) -> Result<MetaOp> {
-        let mut op: MetaOp = MetaOp::new();
-        let lop = op.load_ref_mut();
-        let mut t = r#type.clone();
-        let mut type_iter = btf.type_iter(
-            r#type
+    fn check_expr(&self, field: &LhsNode, rel_op: RelOp, rval: Rhs) -> Result<XlateCtx> {
```
atenart, line 495: The function name could be more descriptive. My understanding is it is constructing a context, not only validating the expression.

retis/src/core/filters/meta/filter.rs
```diff
@@ -268,160 +569,558 @@ impl MetaOp {
...
+                    bail!("trying to cast a leaf member into {tgt}");
+                }
+
+                self.btf_type = snode;
+                self.offt = offset;
+
+                let mut ctx = self.check_expr(field, relop, rval)?;
+                if let Some(bfs) = bfs {
+                    ctx.load.bf_size = u8::try_from(bfs)?;
+                }
```
atenart, line 1038: I guess that could be moved in `check_expr`.

atenart (changes requested):
One part I didn't reviewed thoroughly is the `emit_*` functions, but nothing looked like a blocker, so we can move forward with this. It's actually looking pretty good! (And is a very nice improvement).

Some high level comments:

- If you can split it into more commits that would make it easier to read, but feel free not to do this if that is too much effort.
- Additional comments on functions, structs and in the code would be nice too!
- You might want to consider splitting `filter.rs`.

retis/src/core/filters/meta/filter.rs
```diff
@@ -136,118 +129,426 @@ impl MetaLoad {
...
+                    mask = u64::from_str_radix(inner.as_str().trim_start_matches("0b"), 2)?
+                }
+                e => bail!("while parsing mask {:#?}", e),
+            }
         }
 
-        Ok(())
+        if not {
+            mask = !mask;
+        }
```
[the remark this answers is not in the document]
vlrpl: IIRC, it's a matter of parsing order. The prefix is available before the mask is.

retis/src/core/filters/meta/filter.rs
```diff
@@ -136,118 +129,426 @@ impl MetaLoad {
...
+
+        Ok(LhsNode { member, mask, cast })
+    }
+
+    fn parse_lhs(pair: pest::iterators::Pair<Rule>) -> Result<Lhs> {
+        let mut lhs = Vec::new();
+        for inner in pair.into_inner() {
+            if inner.as_rule() == Rule::ident {
+                lhs.push(Self::parse_ident(inner.clone())?);
```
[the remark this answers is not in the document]
vlrpl: The truth is that the grammar definition already imposes a given structure, meaning pest will raise a parse error in the error case. Most of the error paths are essentially there to make Rust happy (hence the terse error message, as there should be no way to trigger it).

retis/src/core/filters/meta/filter.rs
```diff
@@ -136,118 +129,426 @@ impl MetaLoad {
...
+
+        for inner in pair.into_inner() {
+            match inner.as_rule() {
+                Rule::mask => {
+                    mask = Self::parse_mask(inner)?;
+                }
+                Rule::uident => {
+                    cast = Some(inner.as_str().to_owned());
+                }
+                _ => bail!("parse_ident"),
```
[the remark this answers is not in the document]
vlrpl: See the comment below, but we can improve it for consistency (as it's been done with the others).

retis/src/core/filters/meta/filter.rs
```diff
@@ -474,216 +1266,156 @@ impl FilterMeta {
...
-        let (mut fields, op, rval) = Self::parse_filter(&fstring)?;
+    fn backpatch(&mut self, list: &[usize], target: usize) -> Result<()> {
+        let mut insn;
 
-        // At least two elements are present
-        let init_sym = fields.remove(0).member;
+        for pos in list.iter() {
+            if target > *pos {
+                insn = self.filter.get_raw_insn_mut(*pos)?;
+                insn.set_off_raw(i16::try_from(target - *pos - 1)?);
```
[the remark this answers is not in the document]
vlrpl: I understand your point and I even thought about converting the whole insn.rs into a builder (or giving it some capabilities), but I don't think it's really useful for our purpose (not a library), as we need to mangle the instructions (hence `_raw`). Offset is an existing field, whatever the instruction is. The point of `_raw` is "you're on your own".

retis/src/core/filters/meta/filter.rs
```diff
@@ -136,118 +129,426 @@ impl MetaLoad {
...
+        let mut rhs = Rhs::default();
+        for inner in inner_pairs {
+            match inner.as_rule() {
+                Rule::op => {
+                    op = Self::parse_operator(inner)?;
+                }
+                Rule::rhs => {
+                    rhs = Self::parse_rhs(inner)?;
+                }
+                _ => bail!("unexpected terminal symbol"),
```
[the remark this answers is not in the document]
vlrpl: Same as the comment above about the error path. There should be a major problem in the grammar (or in pest?) to enter this path.
The user "is not supposed" to see these.

retis/src/core/filters/meta/filter.rs
```diff
@@ -268,160 +569,558 @@ impl MetaOp {
             };
         }
 
-        if mask > 0 {
-            if lop.is_ptr() || (lop.is_num() && !lop.is_signed()) {
-                lop.mask = mask;
+        if field.mask > 0 {
+            if ctx.load.is_ptr() || (ctx.load.is_num() && !ctx.load.is_signed()) {
+                ctx.load.mask = field.mask;
```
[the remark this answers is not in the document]
vlrpl: `char *` to aligned mem embedding info in the lsbits?
Can't provide a real example of the above in the kernel, but in general, we want to introduce 

`sk_buff.foo_pointer:mask != 0`

So, having it half there is not bad, right?

retis/src/core/filters/meta/filter.rs
vlrpl: off the top of my head is the `char *` case. Can't determine statically the lval size, so let's just probe_read (safe but could be incorrect, up to the user) rval_len bytes.

retis/src/core/filters/meta/filter.rs
```diff
@@ -268,160 +569,558 @@ impl MetaOp {
...
+                        // Emit load Ptr
+                        tf_list = Some(self.emit_load_ptr(offset / 8, field.mask)?);
+                    }
+                    std::cmp::Ordering::Greater => {
+                        bail!("pointers of pointers are not supported")
+                    }
+                    _ => {
+                        if field.mask != 0 {
+                            bail!("intermediate members masking is only supported for pointers and unsigned numbers");
+                        }
```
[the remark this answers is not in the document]
vlrpl: Not sure I got your question. We don't support everything, but can you provide a specific example?

retis/src/core/filters/meta/filter.rs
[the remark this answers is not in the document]
atenart: Ah right, makes sense!

retis/src/core/filters/meta/filter.rs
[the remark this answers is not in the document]
atenart: I see, let's not add unused code then.

retis/src/core/filters/meta/filter.rs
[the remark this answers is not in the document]
atenart: I don't fully recall what was my point, it's probably this: in case there's a pointer and we mask it, things should work as expected. The error message mentions unsigned numbers should work too, can't those have an indirection of 0?

retis/src/core/filters/meta/filter.rs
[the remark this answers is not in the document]
atenart: Ah OK, in which case `nmemb` will be 0 I guess (not a string).

retis/src/core/filters/meta/filter.rs
atenart: I think I got confused. If we have a pointer to a struct and we want to use the value, would it match `ctx.load.is_ptr()` ?

retis/src/core/filters/meta/filter.rs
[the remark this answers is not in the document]
vlrpl: The branch could theoretically be reached in case of a bug and not an unexpected input (parsing should handle that). I wrote the printout in terms of "terminal symbol" to make that stand out, but we could think of something better. I'm open to suggestions.

retis/src/core/filters/meta/filter.rs
```diff
@@ -474,216 +1266,156 @@ impl FilterMeta {
...
+
+                match ty {
+                    s @ Type::Struct(_) | s @ Type::Union(_) => {
+                        match Self::walk_btf_node(btf, &s, node_name, offset + member.bit_offset())?
+                        {
+                            Some((offt, bfs, x)) => return Ok(Some((offt, bfs, x))),
+                            _ => continue,
+                        }
+                    }
+                    _ => return Ok(None),
```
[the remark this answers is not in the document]
vlrpl: This should be the case of walking anonymous walkable types (structs/unions). I don't think an unnamed member can be of a different type at all (but regardless, we support only those).

Just to clarify, this is old code (simply moved). 
There's room for improvement, nonetheless, I'd focus on the additions (unless we spot a blocker that can't see ATM) as it may be too much at once.
My suggestion/preference, if you agree, is to discuss potential enhancements regarding older parts offline (or open an issue) and follow up.

retis/src/core/filters/meta/filter.rs
```diff
@@ -474,216 +1266,156 @@ impl FilterMeta {
...
-            )
-        })?;
-
-        mask = if not { !mask } else { mask };
-        if mask == 0 {
-            bail!("mask cannot be zero");
+        for member in r#type.members.iter() {
+            let fname = btf.resolve_name(member)?;
+            if fname.eq(node_name) {
+                match btf.resolve_chained_type(member).ok() {
```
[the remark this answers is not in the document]
vlrpl: Moved code, but I don't see a valid reason to hide it.

retis/src/core/filters/meta/filter.rs
vlrpl: The bound check on rval_len is required indeed to avoid later potential errors.

retis/src/core/filters/meta/filter.rs
vlrpl: I would not name it against the fact that it returns a context, though (looks a little redundant). 
`finalize_expr()` ?
Feel free to suggest a better name

docs/filtering.md
```diff
@@ -74,53 +74,141 @@ These filters can match against any subfield of the `sk_buff` and subsequent
...
+
+rhs = { num | string }
+num = @{ hex | bin | ext_dec }
+ext_dec = @{ (neg)? ~ ASCII_DIGIT+ }
+neg = @{ "-" }
+string = @{ PUSH("\"" | "'") ~ any_string+ ~ POP }
+any_string = { (!PEEK ~ ANY)+ }
+
+WHITESPACE   =   _{ " " | "\t" }
+EOF          =   _{ EOI | ";" }
```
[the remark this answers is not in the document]
vlrpl: IMO it's ok to keep it (people can skip it if they don't feel comfortable with it), but I still tried to find a way to add a link.
IMO, it must be linked to the current branch (e.g., avoid breakages on stable branches if it gets moved/renamed on main), but I didn't find a handy way (also, relative paths didn't work as the root dir should be docs).

The only thing I've been able to do is include the grammar from the current tree (at least it is in sync).
https://retis--496.org.readthedocs.build/en/496/filtering/

IDK, maybe that could be collapsed in some way, but my mkdocs/rendering fu is (and most probably will remain) limited

retis/src/core/filters/meta/filter.rs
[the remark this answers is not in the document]
atenart: Maybe having a generic error with the file name + line number. Something saying this should never be reached and there is a bug? (Doing so helps not having to provide a real error message and still implementing something that makes sense while reading the code). WDYT?

docs/filtering.md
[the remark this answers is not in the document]
atenart: No strong opinion (we can keep the current proposal too).

retis/src/core/filters/meta/meta.pest
```diff
@@ -0,0 +1,30 @@
...
+ident = { uident ~ ident_modifiers? }
+ident_modifiers = { ":" ~ mask ~ (":" ~ uident)? }
+mask =   { not? ~ (hex | bin | dec) }
+hex = @{ "0x" ~ ('0'..'9' | 'a'..'f' | 'A'..'F')+ }
+bin = @{ "0b" ~ ("0" | "1")+ }
+dec = @{ ASCII_DIGIT+ }
+not = @{ "~" }
+uident = @{ so_ident ~ re_ident }
+so_ident = _{ ('a'..'z' | 'A'..'Z' | "_") }
+re_ident = _{ (so_ident | "_")* }
```
atenart, line 20: I'm not sure to understand what are `so_ident` and `re_ident` and how the parser distinguishes the two (as `_` is part of both)?

It seems the code just reads the `uident` value as a string and does not distinguish between the two members.

retis/src/core/filters/meta/bpf/include/meta_filter.h
```diff
@@ -3,322 +3,16 @@
...
+#define meta 0xdeadbeed
+enum meta_filter_type {
+	META = meta,
+} __binding;
 
-	do {
-		ret = *sp1 - *sp2;
-		if (ret)
-			break;
+// BINDING_DEF(META, meta)
```
atenart: nit: probably a leftover

retis/src/core/filters/meta/bpf/include/meta_filter.h
```diff
@@ -3,322 +3,16 @@
...
-
-	if (ctx.type & PTR_BIT || ctx.nmemb > 0)
-		return filter_bytes(&ctx);
-
-	return filter_num(&ctx);
-}
+/* Please keep in sync with its Rust counterpart. */
+#define __META_TARGET_MAX	32
+static __hidden __binding const u32 META_TARGET_MAX = meta;
+//BINDING_DEF(meta_target_max, META_TARGET_MAX)
```
atenart: Is the above used in the C part?

nit: part of it is commented

retis/src/core/filters/meta/filter.rs
```diff
@@ -268,160 +578,562 @@ impl MetaOp {
...
 
-impl Rval {
-    fn from_str(rval: &str) -> Result<Rval> {
-        let detected = if (rval.starts_with('"') && rval.ends_with('"'))
-            || (rval.starts_with('\'') && rval.ends_with('\''))
-        {
-            Rval::Str(rval[1..rval.len() - 1].to_string())
+        if ctx.load.bf_size > 0 {
+            bf_rshf = 64 - ctx.load.bf_size as i32;
+            bf_lshf = bf_rshf - (ctx.load.offt % 8) as i32;
```
atenart, line 885: Shouldn't that be `% 64` as `ctx.load.offt` is in bits?

retis/src/core/filters/meta/filter.rs
```diff
@@ -268,160 +578,562 @@ impl MetaOp {
...
+        self.filter.add(eBpfInsn::ld(
+            LdInfo::Reg {
+                src: BpfReg::FP,
+                dst: BpfReg::R5,
+                off: -8,
+            },
+            (sz as u8).try_into()?,
+        ));
+
+        if (ctx.target.sz != 8 && ctx.load.is_signed()) || ctx.load.bf_size > 0 {
```
atenart, line 962: Why restrict on signed values here? Shouldn't extra bits be discarded from unsigned non 8-bytes values too?

retis/src/core/filters/meta/filter.rs
```diff
@@ -268,160 +578,562 @@ impl MetaOp {
...
+            dst: BpfReg::R0,
+            imm: 0x00,
+        }));
+
+        tf_list.push_false(self.filter.len());
+        self.filter.add(eBpfInsn::jmp_a(0));
+
+        self.filter.add(eBpfInsn::mov(MovInfo::Imm {
+            dst: BpfReg::R0,
+            imm: 0xff00,
```
atenart: Do I understand correctly that on success this will return `0xff00`, the array comp will return `!0` for equality success and `!!0` for neq success? Any comment on why those values?

retis/src/core/filters/meta/filter.rs
atenart: OK, makes sense to probe read as this should be safe. Can you check this is for `char *` and add a comment?

vlrpl: The case is that, no doubt.
The problem is finding an actual `char *` in the `sk_buff` :)

retis/src/core/filters/meta/meta.pest
vlrpl: That's the purpose of silent rules (`_{...}`) 
`Silent rules do not create token pairs during parsing`

retis/src/core/filters/meta/bpf/include/meta_filter.h
vlrpl: It is a leftover, but it's a good one.
Bindgen doesn't always play well with macros (expansion doesn't seem to happen, at least not always), and it was the reason I initially pushed back macros. We're kind of ok, but this should be handled at some point (or wait bindgen to fix it/add full support)

retis/src/core/filters/meta/bpf/include/meta_filter.h
vlrpl: should be used, but there's still a local variable

retis/src/core/filters/meta/filter.rs
vlrpl: yes, it is in bits and we want to know how it is offseted WRT a rounded byte. 
It should be logically equivalent to something like `bit_offset - byte_offset * 8` (e.g. 1003 - 1000).
Consider e.g. the case:
```
offt = 1003
bf_size = 5
```

You should shift 56 left, 59 right

retis/src/core/filters/meta/filter.rs
vlrpl: because this is a sign fixup for non-bitfields, whereas for bitfields the idea is to discard left and right bits not belonging to the bitfield (and there's compiler sign extensions for them to take care of, so both signed and unsigned have to be fixed up)

retis/src/core/filters/meta/filter.rs
vlrpl: any positive value. There's `0x40000` from libpcap that also has very very little meaning to us (but there's also a comment that doesn't add anything useful about it, so we can use that). I genuinely don't recall how it ended up being `0xff00` here (can perfectly be the first number that initially came to mind, even though I'm surprised it was not 0xf00d :)

retis/src/core/filters/meta/bpf/include/meta_filter.h
atenart: I only see `META_TARGET_MAX` being used in `retis/src/core/filters/meta/filter.rs`. Can't it be removed from the C part and the UAPI then?

retis/src/core/filters/meta/filter.rs
atenart: Makes sense. +1 for `0xf00d` if we need an arbitrary value one day :laughing:

retis/src/core/filters/meta/bpf/include/meta_filter.h
vlrpl: you're right, I was assuming it was consumed for some nonexistent reason by the BPF programs. Makes sense.

retis/src/core/filters/meta/filter.rs
atenart: OK, I get it after reading it again. It's just that for now we only support reading strings in this case, not other pointer values.

retis/src/core/filters/meta/filter.rs
atenart:
```
offt = 1003
bf_size = 5

-> bits 47:43
```

Shouldn't the left shift be `63-47=16`?

Or do I need a lot more coffee? :-)

vlrpl: I guess this is ok for a quick offline sync, but the idea is:
You read at 1000 (offt / 8), and so you have:

X{56}...11111XXX

Left shifting you clean the left side `0{56}` (also functional for sign fix up), with the [s]rshift you do the remaining part (clean right bits potentially belonging to the next contiguous memory unit and complete the sign fix up).

Why bits 47:43? bf_size is in bits.

Am I missing something?

atenart: I meant 1043 sorry; but then I got it (also thanks to offline sync) and this makes sense now. Thanks!

PR #497, opened 2025-02-27
[RFC] Bump btf-rs to 1.2

atenart:
RFC as btf-rs 1.2 is not released yet. This make Retis use the btf-rs `BtfCollection`.


vlrpl:
I quickly went through this.
As a general comment, I'm in favor of this as it will be logically useful to decouple things in the outstanding effort in #496.
But other than that, things look more natural this way.
After a quick glance, I didn't spot anything worthy of comments code-wise, but I'll reserve the right to take a second look soon if/when we decide to move this on.

atenart:
I'm still planning on submitting such changes but extra work might be needed on the btf-rs side to release a new version. In the meantime closing this.

PR #502, opened 2025-03-06
Improve startup time

atenart:
As of now Retis can take quite some time to handle various things before starting a collection. Investigating where time is spent boils down to three main points:

- A good amount of time is spent reading `/proc/kallsyms`; not much we can do here.
- It turns out removing noops when loading BPF programs is far from being optimized. This is fixed by #496.
- Loading BPF and attaching programs take a lot of time and are the major contributor to slowness. This is the goal of this PR (mainly reducing the number of programs loaded).

Combining this and #496 and using the `generic` profile, we can go from ~6.5s to ~2.5s starting time. In addition to this using kprobe multi helps when using lots of kprobes, especially at exit time (1 link vs 1 per probe). The raw tracepoints benefit from an optimization too, which helps a lot when using a lot of them — this was a major pain point in the past and one of the important reasons Retis could be slow to start.

Note that the last commit, grouping targeted probes with no hooks, might not show a lot of improvement atm but could be in the future. Nevertheless it is more logical, future proof and the changes are really not invasive.

This is based on multiple `libbpf-rs` changes, which have been merged upstream already but not released yet. This is the only reason this is marked as an RFC.

Some extra details in the commit logs.


atenart:
I just realized support for getting the BPF cookie in raw tracepoints was only added a year ago[1]. For those kernels we could not use the cookie in raw tracepoints and keep the legacy one program per probe logic.

[1] 68ca5d4eebb8 ("bpf: support BPF cookie in raw tracepoint (raw_tp, tp_btf) programs")

atenart:
> I just realized support for getting the BPF cookie in raw tracepoints was only added a year ago[1]. For those kernels we could not use the cookie in raw tracepoints and keep the legacy one program per probe logic.
> 
> [1] 68ca5d4eebb8 ("bpf: support BPF cookie in raw tracepoint (raw_tp, tp_btf) programs")

Setting and getting cookies, as well as kprobe multi, are not supported on all kernel versions we want to support (e.g. rhel8). For this reason the old behavior was kept as a fallback when the wanted solution is not available. I also merged the kprobe and kretprobe support to ease such changes and maintenance.

atenart:
TODO: regenerate both `vmlinux.h` files to latest kernel.

retis/src/core/probe/kernel/raw_tracepoint.rs
```diff
@@ -128,20 +145,61 @@ impl RawTracepointBuilder<'_> {
...
             self.links
-                .push(prog.attach_raw_tracepoint_with_opts(probe.symbol.attach_name(), opts)?);
-        } else {
+                .push(prog.attach_raw_tracepoint_with_opts(symbol.attach_name(), opts)?);
+        }
+
+        Ok(())
+    }
+
+    fn attach_raw_tracepoints_no_cookie(&mut self, probes: &[Probe]) -> Result<()> {
```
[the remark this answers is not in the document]
atenart: > The idea looks nice. It looks like this goes in the direction of limiting (non-coalesceable/common) rodata. I can't think of specific reasons why this could be limiting, but in general it seems the idea here is to.

Yes, that's right. This was already the case in practice for kprobes, as the same loaded program was reused. Kprobe multi does not change this. The new part is about raw tracepoints, where not having per-probe rodata helps with loading time.

I think that's fine, the ideal scenario is using rodata for shared capabilities / configuration and the per-symbol map for per-probe configuration. I actually wanted to go further and provide `nargs` as part of this (allowing to load a single tp program) but that doesn't work with the verifier.

IMO it's actually not a design change, only tp were following this and only because of limitations.

> For this to work, we need to adopt cookies usage for kernels supporting them, but still handling the "legacy" case (here I wonder if/when we may want to start considering removing the support for those kernels rather than special casing legacy scenarios).

Yes, 100% agree there will be a time when we'll drop support for this. For now only c8s kernel do not support multi kprobe and cookies. Raw tp cookies are less well supported as support was added ~a year ago (they are in recent c9s kernels).

> Also, it's fair to say the more we squash, the more the gain, but it's also fair to say that start-up delay is generally more acceptable.

If we have a really good reason and increasing startup time is the only way, that might be ok. Otherwise > 6 secs seems quite long IMO.

> Furthermore, I'm thinking about freplace removal (that probably is not impactful, but potentially consuming `.rodata` gaining dead code removal), but I'm also thinking about another non-rodata scenario that could lead to handling the exception, and it's about per-probe filtering. My point is of course raising potential (at least half envisioned) points that are worth a discussion WRT this change.

I think it comes down to the need to load a given program + its configuration for a set of probes that can share it. If a single program can, then that's fine. Otherwise that's better. Then it's about how we implement our features to maximize reuse of loaded programs, but it's not any different from the current situation with kprobes. That's a good point to keep in mind though.

> Do you have some numbers handy about the specific gain here?

The load of raw tp are currently the most significant reason startup time is long. The change to share the same program for some of them (based on nargs) is thus the most significant gain in this PR. For example adding multi kprobe does not change things much. IIRC sharing raw tp programs alone was a gain of at least 1.5s, which is half the total gain here.

(For completeness, the above statement is when using the `generic` profile. When adding even more tracepoints the gain is even better, but also there is only a limited number of them. For kprobes, multi support really shows when adding 100s of probes, and mostly at exit time — one link vs one per-probe).

atenart:
Libbpf-rs was released which allowed to complete this PR, however the new functionalities in the probe builders require an updated vmlinux.h header. It turned out generating one was again broken and this took some time as we hit multiple issues.

Things looks OKish now, but the updated `gen_vmlinux.sh` script relies on non-upstream pahole fixes. We might want to wait for those to be submitted to see if the approach (and the generated vmlinux.h file) is right.

retis/src/core/probe/kernel/kprobe.rs
```diff
@@ -85,34 +86,14 @@ impl<'a> ProbeBuilder for KprobeBuilder<'a> {
...
-            );
-        }
+    fn add_probe(&mut self, probe: Probe) -> Result<()> {
+        self.probes.push(probe);
+        Ok(())
+    }
 
+    fn attach(&mut self) -> Result<()> {
+        let tmp = std::mem::take(&mut self.probes);
+        self.attach_kprobes(&tmp)?;
```
amorenoz: nit: why not draining `self.probes` inside `attach_kprobes` or just inlining it here?

retis/src/core/probe/kernel/bpf/raw_tracepoint.bpf.c
```diff
@@ -6,9 +6,11 @@
 /* It is safe to have these values per-object as the loaded object won't be
  * shared between attached programs for raw tracepoints.
  */
-const volatile u64 ksym = 0;
 const volatile u32 nargs = 0;
 
+/* Only used in legacy mode where the raw tracepoint program can't be reused */
```
amorenoz: nit: we are kind of messy with this but at least in this file we are adding full stops at the end of sentences...

retis/src/core/inspect/inspect.rs
```diff
@@ -38,6 +38,27 @@ impl Inspector {
     }
 }
 
+/// Parses a struct and returns its variant names, trimed if asked to.
```
amorenoz: s/a struct/an enum

retis/src/core/probe/kernel/kprobe.rs
```diff
@@ -19,17 +19,16 @@ mod kprobe_bpf {
...
 
+#[derive(Default, PartialEq, Eq)]
 enum KprobeVariant {
-    Kprobe,
+    KprobeMulti,
+    // We're not supporting every variant in between (kprobe + cookie, kprobe
+    // multi - cookies) to keep things reasonable. It's also fine to fallback in
+    // case the kernel is not recent enough.
+    #[default]
     KprobeNoCookie,
```
amorenoz: Having a two variant enum that needs such an explanation suggests to me that we could replace this with a simpler `bool legacy`, or `bool compat'.

retis/src/core/probe/kernel/raw_tracepoint.rs
```diff
@@ -133,20 +150,61 @@ impl RawTracepointBuilder<'_> {
...
 
-        if self.cookie_support {
+        Ok(skel)
+    }
+
+    fn attach_raw_tracepoints(&mut self, nargs: u32, probes: &[Probe]) -> Result<()> {
+        #[allow(clippy::map_entry)] // Fixes double mutable refs.
+        if !self.skels.contains_key(&nargs) {
+            let new = self.init_skel(nargs, None)?;
+            self.skels.insert(nargs, new);
```
amorenoz, line 151: Why would `self.skels` contain the key? AFAICS, it is called once per distinct `nargs`.
Isn't it cleaner to always create the skel and insert it into `skels` at end?

retis/src/core/probe/manager.rs
```diff
@@ -492,7 +489,28 @@ impl ProbeRuntimeManager {
             Ok(())
         })?;
 
-        self.generic_builders = builders;
+        Ok(builders)
+    }
+
+    /// Populates generic builders.
+    fn gen_generic_builders(&mut self) -> Result<()> {
+        // Already initialized? Bail out early.
```
amorenoz: nit: the comment is rather redundant.

retis/src/core/probe/manager.rs
```diff
@@ -401,6 +402,7 @@ pub(crate) struct ProbeRuntimeManager {
     #[cfg(not(test))]
     counters_map: libbpf_rs::MapHandle,
     generic_builders: HashMap<usize, Box<dyn ProbeBuilder>>,
+    targeted_nohook_builders: HashMap<usize, Box<dyn ProbeBuilder>>,
```
amorenoz: nit: Unrelated but when trying to read this code after a while its not trivial what the key of this HashMap (and the others) is. Maybe  we could `type ProbeTypeIdx = usize`

retis/src/core/probe/manager.rs
```diff
@@ -492,7 +489,28 @@ impl ProbeRuntimeManager {
...
+        // Already initialized? Bail out early.
+        if !self.generic_builders.is_empty() {
+            return Ok(());
+        }
+
+        self.generic_builders = self.gen_builders(true)?;
+        Ok(())
+    }
+
+    /// Populates targeted nohhok builders.
```
amorenoz: s/nohhok/nohook/

retis/src/core/probe/manager.rs
```diff
@@ -492,7 +489,28 @@ impl ProbeRuntimeManager {
...
+    fn gen_targeted_nohook_builders(&mut self) -> Result<()> {
+        // Already initialized? Bail out early.
```
amorenoz: ditto

retis/src/core/probe/manager.rs
```diff
@@ -460,12 +462,7 @@ impl ProbeRuntimeManager {
     }
 
     /// Populates generic builders.
```
amorenoz: Comment is now  confusing

retis/src/core/probe/kernel/kprobe.rs
atenart: Indeed. I'll move to using `drain(..)` in the helpers directly.

retis/src/core/probe/kernel/kprobe.rs
atenart: Fair point. Initially I supported more than 2 variants in my dev branch but though that this was too much hassle for little benefits and decided to only support the latest one and a compatibility one. The enum stayed. The minor nice thing is it is more descriptive than a bool and can support additional flavors if the need arises one day.

However I really don't have a strong opinion on this, so let me know what you prefer.

retis/src/core/probe/kernel/raw_tracepoint.rs
atenart: We can have the following set of events: `add_probe -> attach -> add_probe -> attach`, in which case it's better if we can reuse an existing `skel` if the `nargs` matches.

retis/src/core/probe/manager.rs
atenart: That's just a c/p but I'll remove it.

retis/src/core/probe/manager.rs
atenart: Makes sense, I'll add a patch.

vlrpl (approved):
lgtm

retis/src/core/probe/kernel/kprobe.rs
amorenoz: I guess if additional flavors arise, we can revisit it but for me, a `bool compat` is clearer.
You will also need to explain the meaning of `compat` in a comment (as the current implementation does).
But when a reader is reading along (not having the exact definition of  "compat" in mind), and sees a KprobeMulti-related decision made based on the value of "compat" their brain will say "that makes sense" without necessarily having to refresh what "compat" means.
However, when a reader reads about KprobeMulti associated with a variant taking "NoCookie" option, their brain will raise a flag that is only calmed by going back and reading the definition of the enum. It happened to me a couple of times while reviewing this PR and I had most of it in my mind already.

retis/src/core/probe/kernel/raw_tracepoint.rs
amorenoz: Oh, stack probing! Right.

PR #506, opened 2025-03-14
Container kconfig fixes

atenart:
The removal of the `--kconf` root level option in #500  lead to some issues, namely:
- The `Containerfile` was not updated and this lead to generating incompatible images.
- The container script is linked from our documentation but does not work anymore for older images, including the current stable one.
- A syntax error could be seen while running the script.

Another potential issue would be running newer images while using the old container script. IMO that's not ideal but probably OK. If we want, we could add a hidden top level `--kconf` option and match it to display an error message, e.g. "option was removed, please update your script". I have not done that here as I'm not 100% convinced.


vlrpl (approved):
lgtm, thanks!

PR #507, opened 2025-03-21
ovs_operation: fix load issue in Rawhide

vlrpl:
Remove the barrier_var() as this seems to alleviate the issue on Rawhide.

There's still the potential to break the verification on the previous versions of clang as this was introduced exactly for that.
Let's check this giving it a spin of functional tests.

### DIFF
--- a/retis/src/collect/collector/ovs/bpf/include/ovs_operation.h
+++ b/retis/src/collect/collector/ovs/bpf/include/ovs_operation.h
@@ -120,10 +120,6 @@ static __always_inline bool batch_is_processing(const struct upcall_batch *batch
 
 /* Retrieve the current upcall being processed. */
 static __always_inline struct user_upcall_info *batch_current(struct upcall_batch *batch) {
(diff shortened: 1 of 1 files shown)


retis/src/collect/collector/ovs/bpf/include/ovs_operation.h
```diff
@@ -123,7 +123,6 @@ static __always_inline struct user_upcall_info *batch_current(struct upcall_batc
 	/* In some cases the branch get optimized. The barrier below
 	 * appear to prevent that making the verifier happy.
 	 */
```
atenart, line 125: Can you remove the comment as well?

vlrpl: yeah, removed it, thanks. I was too focused on "whether it works with the older distros" side  :)

atenart (approved):
LGTM, thanks!

PR #514, opened 2025-04-09
Stack trace filter

vlrpl:
Had this old patch hanging locally.
RFC as I'm mostly sending it as it was (just rebased).

- it's in addition to the stack option but can be as an alternative to those (e.g. --stack-list ... conflicting with the others)
- It's a global option remapped per-probe (we have to take into account automatic probes)

Some remarks:
- it doesn't formally check whether the symbol exists (we can extend it to do it, but I'm not sure what the advantage is), but just the syntax (consistency with the `-p`)
- even with the former bullet, it would be possible to specify an unprobed but existing symbol. 

The last bullet would implicitly cover the probe existence check but would require a bit of special casing.
I don't have a strong preference for the previous remarks (e.g. whether special casing while checking through the set option is worth it), but feel free to propose a preferred approach.

### DIFF
--- a/retis/src/collect/collect.rs
+++ b/retis/src/collect/collect.rs
@@ -221,10 +221,15 @@ impl Collectors {
         self.run.register_term_signals()?;
 
         // Check if we need to report stack traces in the events.
-        if collect.stack || collect.probe_stack {
+        if collect.stack {
             self.probes
                 .builder_mut()?
-                .set_probe_opt(probe::ProbeOption::StackTrace)?;
+                .set_probe_opt(probe::ProbeOption::ReportStack)?;
+        }
+        if collect.probe_stack {
+            self.probes
+                .builder_mut()?
+                .set_probe_opt(probe::ProbeOption::ProbeStack)?;
         }
 
         // Generate an initial event with the startup section.
@@ -542,7 +547,6 @@ impl Collectors {
 
         let (mut iccount, mut eccount) = (0, 0);
         let mut probe_stack = ProbeStack::new(
-            collect.stack,
             self.probes.runtime_mut()?.attached_probes(),
             self.known_kernel_types.clone(),
         );
--- a/retis/src/core/probe/kernel/probe_stack.rs
+++ b/retis/src/core/probe/kernel/probe_stack.rs
@@ -10,7 +10,7 @@ use crate::{
     core::{
         inspect::inspector,
         kernel::Symbol,
-        probe::{Probe, ProbeRuntimeManager},
+        probe::{Probe, ProbeOption, ProbeRuntimeManager},
     },
     events::{Event, KernelEvent, SectionId},
 };
@@ -21,19 +21,13 @@ pub(crate) struct ProbeStack {
     /// Local cache of functions that were already considered, probed or not.
     /// Used to optimize the event processing.
     probes: HashSet<String>,
-    /// Should the stack stay in the event?
-    keep_stack: bool,
     /// Set of kernel types known by collectors, so we only probe functions that
     /// can generate an event.
     known_kernel_types: HashSet<String>,
 }
 
 impl ProbeStack {
-    pub(crate) fn new(
-        keep_stack: bool,
-        attached_probes: Vec<String>,
-        known_kernel_types: HashSet<String>,
-    ) -> Self {
+    pub(crate) fn new(attached_probes: Vec<String>, known_kernel_types: HashSet<String>) -> Self {
         // Stack probe mode works with kprobes only and keeps track of the
         // function name.
         let attached_probes = attached_probes
@@ -44,9 +38,20 @@ impl ProbeStack {
 
         Self {
             probes: HashSet::from_iter(attached_probes),
-            keep_stack,
             known_kernel_types,
         }
+    }
+
+    fn keep_stack(&mut self, mgr: &mut ProbeRuntimeManager, evt: &KernelEvent) -> bool {
+        let r#type = match evt.probe_type.as_str() {
+            "raw_tracepoint" => "tp",
+            s => s,
+        };
+
+        let sym = format!("{}:{}", r#type, evt.symbol);
+
+        mgr.get_probe_opts(&sym)
+            .is_some_and(|opts| opts.contains(&ProbeOption::ReportStack))
     }
 
     /// Process a new event and detect additional functions to add a probe too.
@@ -119,7 +124,7 @@ impl ProbeStack {
             Ok(())
         })?;
 
-        if !self.keep_stack {
+        if !self.keep_stack(mgr, kernel) {
             kernel.stack_trace = None;
         }
 
--- a/retis/src/core/probe/kernel/utils.rs
+++ b/retis/src/core/probe/kernel/utils.rs
@@ -1,8 +1,10 @@
+use std::collections::HashSet;
+
 use anyhow::{bail, Result};
 
 use crate::core::{
     kernel::symbol::{matching_events_to_symbols, matching_functions_to_symbols, Symbol},
-    probe::Probe,
+    probe::{Probe, ProbeOption},
 };
 
 /// Probe type for probes given through cli arguments.
@@ -23,23 +25,47 @@ impl CliProbeType {
     }
 }
 
+/// Parses the probe options given as a cli argument and returns their
+/// ProbeOption representation.
+fn parse_cli_probe_opts(options: &str) -> Result<HashSet<ProbeOption>> {
+    let opts = options.split('/');
+
+    if opts.clone().next_back() == Some("") {
+        bail!("Empty options are not allowed. Check your option list doesn't terminate with '/'.");
+    }
+
+    opts.map(ProbeOption::try_from)
+        .try_fold(HashSet::new(), |mut hset, option| {
+            if !hset.insert(option?.clone()) {
+                bail!("duplicate options detected in {options}.");
+            }
+
+            Ok(hset)
+        })
+}
+
 /// Parses a probe given as a cli argument and returns its type and the probe
 /// with the type identifier (if any).
-pub(crate) fn parse_cli_probe(input: &str) -> Result<(CliProbeType, &str)> {
+pub(crate) fn parse_cli_probe(input: &str) -> Result<(CliProbeType, &str, HashSet<ProbeOption>)> {
     use CliProbeType::*;
+
+    let (input, opts) = match input.split_once('/') {
+        Some((probe, options)) => (probe, parse_cli_probe_opts(options)?),
+        None => (input, HashSet::new()),
+    };
 
     Ok(match input.split_once(':') {
         Some((type_str, target)) => match type_str {
-            "kprobe" | "k" => (Kprobe, target),
-            "kretprobe" | "kr" => (Kretprobe, target),
-            "raw_tracepoint" | "tp" => (RawTracepoint, target),
+            "kprobe" | "k" => (Kprobe, target, opts),
+            "kretprobe" | "kr" => (Kretprobe, target, opts),
+            "raw_tracepoint" | "tp" => (RawTracepoint, target, opts),
             // If a single ':' was found in the probe name but we didn't match
             // any known type, defaults to trying using it as a raw tracepoint.
-            _ if input.chars().filter(|c| *c == ':').count() == 1 => (RawTracepoint, input),
+            _ if input.chars().filter(|c| *c == ':').count() == 1 => (RawTracepoint, input, opts),
             x => bail!("Invalid TYPE {}. See the help.", x),
         },
         // If no ':' was found, defaults to kprobe.
-        None => (Kprobe, input),
+        None => (Kprobe, input, opts),
     })
 }
 
@@ -51,7 +77,7 @@ where
 {
     use CliProbeType::*;
 
-    let (r#type, target) = parse_cli_probe(probe)?;
+    let (r#type, target, options) = parse_cli_probe(probe)?;
 
     // Convert the target to a list of matching ones for probe types
     // supporting it.
@@ -67,11 +93,17 @@ where
             continue;
         }
 
-        probes.push(match r#type {
+        let mut probe = match r#type {
             Kprobe => Probe::kprobe(symbol)?,
             Kretprobe => Probe::kretprobe(symbol)?,
             RawTracepoint => Probe::raw_tracepoint(symbol)?,
-        })
+        };
+
+        options
+            .iter()
+            .try_for_each(|o| probe.set_option(o.clone()))?;
+
+        probes.push(probe)
     }
 
     Ok(probes)
--- a/retis/src/core/probe/manager.rs
+++ b/retis/src/core/probe/manager.rs
@@ -101,23 +101,11 @@ impl ProbeManager {
             _ => bail!("Probe manager is already at runtime state"),
         };
 
-        // Prepare all hooks:
-        // - Reuse global maps.
-        // - Set global options.
+        // Prepare hooks.
         builder
             .generic_hooks
             .iter_mut()
             .for_each(|h| h.maps.extend(builder.maps.clone()));
-        builder.probes.values_mut().try_for_each(|p| {
-            builder
-                .maps
-                .iter()
-                .try_for_each(|(name, fd)| p.reuse_map(name, *fd))?;
-            builder
-                .global_probes_options
-                .iter()
-                .try_for_each(|o| p.set_option(o.clone()))
-        })?;
 
         // Set up filters and their handlers.
         for filter in builder.filters.iter() {
@@ -169,6 +157,7 @@ impl ProbeManager {
             generic_builders: HashMap::new(),
             targeted_builders: Vec::new(),
             probes: HashSet::new(),
+            global_probes_options: builder.global_probes_options.into_iter().collect(),
             filters: builder.filters,
         };
 
@@ -412,6 +401,7 @@ pub(crate) struct ProbeRuntimeManager {
     map_fds: Vec<(String, RawFd)>,
     hooks: Vec<Hook>,
     probes: HashSet<String>,
+    global_probes_options: Vec<ProbeOption>,
     filters: Vec<Filter>,
 }
 
@@ -501,9 +491,20 @@ impl ProbeRuntimeManager {
         Ok(())
     }
 
+    /// Make the probe inherit the global options.
+    fn prepare_probe(&mut self, probe: &mut Probe) -> Result<()> {
+        self.map_fds
+            .iter()
+            .try_for_each(|(name, fd)| probe.reuse_map(name, *fd))?;
+        self.global_probes_options
+            .iter()
+            .try_for_each(|o| probe.set_option(o.clone()))
+    }
+
     /// Attach a new targeted probe.
     #[cfg(not(test))]
     fn attach_targeted_probe(&mut self, probe: &mut Probe) -> Result<()> {
+        self.prepare_probe(probe)?;
         if !self.probes.insert(probe.key()) {
             bail!("A probe on {probe} is already attached");
         }
@@ -535,6 +536,7 @@ impl ProbeRuntimeManager {
     /// Attach a new generic probe.
     #[cfg(not(test))]
     pub(crate) fn attach_generic_probe(&mut self, probe: &mut Probe) -> Result<()> {
+        self.prepare_probe(probe)?;
         if !self.probes.insert(probe.key()) {
             bail!("A probe on {probe} is already attached");
         }
--- a/retis/src/core/probe/manager.rs
+++ b/retis/src/core/probe/manager.rs
@@ -2,7 +2,7 @@
 #![cfg_attr(test, allow(unused_imports))]
 use std::{
     cmp,
-    collections::{HashMap, HashSet},
+    collections::HashMap,
     os::fd::{AsFd, AsRawFd, RawFd},
 };
 
@@ -156,7 +156,7 @@ impl ProbeManager {
             hooks: builder.generic_hooks.into_iter().collect(),
             generic_builders: HashMap::new(),
             targeted_builders: Vec::new(),
-            probes: HashSet::new(),
+            probes: HashMap::new(),
             global_probes_options: builder.global_probes_options.into_iter().collect(),
             filters: builder.filters,
         };
@@ -400,7 +400,7 @@ pub(crate) struct ProbeRuntimeManager {
     targeted_builders: Vec<Box<dyn ProbeBuilder>>,
     map_fds: Vec<(String, RawFd)>,
     hooks: Vec<Hook>,
-    probes: HashSet<String>,
+    probes: HashMap<String, Vec<ProbeOption>>,
     global_probes_options: Vec<ProbeOption>,
     filters: Vec<Filter>,
 }
@@ -505,7 +505,11 @@ impl ProbeRuntimeManager {
     #[cfg(not(test))]
     fn attach_targeted_probe(&mut self, probe: &mut Probe) -> Result<()> {
         self.prepare_probe(probe)?;
-        if !self.probes.insert(probe.key()) {
+        if self
+            .probes
+            .insert(probe.key(), probe.options.clone().into_iter().collect())
+            .is_some()
+        {
             bail!("A probe on {probe} is already attached");
         }
 
@@ -537,7 +541,11 @@ impl ProbeRuntimeManager {
     #[cfg(not(test))]
     pub(crate) fn attach_generic_probe(&mut self, probe: &mut Probe) -> Result<()> {
         self.prepare_probe(probe)?;
-        if !self.probes.insert(probe.key()) {
+        if self
+            .probes
+            .insert(probe.key(), probe.options.clone().into_iter().collect())
+            .is_some()
+        {
             bail!("A probe on {probe} is already attached");
         }
 
@@ -549,7 +557,12 @@ impl ProbeRuntimeManager {
 
     /// Get the list of all currently attached probes.
     pub(crate) fn attached_probes(&self) -> Vec<String> {
-        self.probes.clone().into_iter().collect()
+        self.probes.keys().cloned().collect()
+    }
+
+    /// Get the list of all currently attached probes with a specific option.
+    pub(crate) fn get_probe_opts(&self, probe: &str) -> Option<&Vec<ProbeOption>> {
+        self.probes.get(probe)
     }
 
     /// Detach all probes.
--- a/retis/src/core/probe/probe.rs
+++ b/retis/src/core/probe/probe.rs
@@ -26,8 +26,20 @@ pub(crate) enum ProbeType {
 /// Probe options, to toggle opt-in/out features.
 #[derive(Clone, Eq, Hash, PartialEq)]
 pub(crate) enum ProbeOption {
-    StackTrace,
+    ProbeStack,
+    ReportStack,
     NoGenericHook,
+}
+
+impl TryFrom<&str> for ProbeOption {
+    type Error = anyhow::Error;
+
+    fn try_from(option: &str) -> Result<Self> {
+        Ok(match option {
+            "stack" => Self::ReportStack,
+            _ => bail!("'{option}' is an invalid probe option."),
+        })
+    }
 }
 
 /// Represents a probe we can install in a target (kernel, user space program,
@@ -182,11 +194,14 @@ impl Probe {
         }
 
         // Merge options.
-        // - ProbeOption::StackTrace: if any of the probes has it, it should be
+        // - ProbeOption::{ProbeStack,ReportStack}: if any of the probes has it, it should be
         //   set in the resulting probe.
         // - ProbeOption::NoGenericHook: has to be set in both probes to be set in the
         //   resulting probe.
-        if let Some(opt) = other.options.take(&ProbeOption::StackTrace) {
+        if let Some(opt) = other.options.take(&ProbeOption::ProbeStack) {
+            self.options.insert(opt);
+        }
+        if let Some(opt) = other.options.take(&ProbeOption::ReportStack) {
             self.options.insert(opt);
         }
         if !other.options.contains(&ProbeOption::NoGenericHook) {
--- a/retis/profiles/drop.yaml
+++ b/retis/profiles/drop.yaml
@@ -4,7 +4,7 @@ about: Drop monitor-like
 collect:
   - args:
       collectors: skb-drop,skb
-      stack: ~
+      probe: tp:skb:kfree_skb/stack
 pcap:
   - args:
       probe: tp:skb:kfree_skb
@@ -16,7 +16,7 @@ collect:
   - args:
       collectors: nft,skb
       nft_verdicts: drop
-      stack: ~
+      probe: kprobe:__nft_trace_packet/stack
 pcap:
   - args:
       probe: kprobe:__nft_trace_packet
--- a/retis/src/core/probe/kernel/kernel.rs
+++ b/retis/src/core/probe/kernel/kernel.rs
@@ -45,7 +45,7 @@ impl KernelProbe {
 
         #[allow(clippy::single_match)]
         options.iter().for_each(|o| match o {
-            ProbeOption::StackTrace => {
+            ProbeOption::ProbeStack | ProbeOption::ReportStack => {
                 config.stack_trace = 1;
             }
             _ => (),
--- a/retis/src/process/cli/pcap.rs
+++ b/retis/src/process/cli/pcap.rs
@@ -242,7 +242,7 @@ impl SubCommandParserRunner for Pcap {
         // The following unwrap() will never fail as Clap makes sure that either
         // list_probes is true, or probe is Some().
         let probe = self.cmd.probe.as_ref().unwrap();
-        let (probe_type, target) = parse_cli_probe(probe)?;
+        let (probe_type, target, _) = parse_cli_probe(probe)?;
         let symbol = Symbol::from_name_no_inspect(target);
 
         // Filtering logic.


retis/src/collect/cli.rs
```diff
@@ -108,6 +108,12 @@ defaults to \"retis.data\"."
 not released. If exhausted, no stack trace will be included."
     )]
     pub(super) stack: bool,
+    #[arg(
+        long,
+        value_delimiter=',',
+        help = "Requires to be used at with at least --stack or --probe-stack. The option will limit the stack traces generated to the specified symbols."
+    )]
+    pub(crate) stack_trace_limit: Vec<String>,
```
atenart, line 116: This is one way of doing it; for each additional per-probe option we would end up with a new option. The other way would be to allow specifying per-probe options in the cli, e.g. `--probe ip_rcv+stack`.

vlrpl: > This is one way of doing it; for each additional per-probe option we would end up with a new collect option. The other way would be to allow specifying per-probe options in the cli, e.g. `--probe ip_rcv+stack`.

yes. That is a good idea. 
In particular, I thought about this as well (picking a preferred approach is one of the reasons for the RFC tag) but it looked weird for automatic probes (e.g. skb-drop and `skb:kfree_skb`), but maybe it's not that weird. 
I guess this would have the advantage of setting the per-probe options without requiring 1:N remap and implicit checks (very nice pros). 
It would require some sort of special casing e.g. for --stack and `+stack` local options as we should make sure `--stack` is not applied to all the probes but just to the `+stack` ones.
OTOH, we can convert `--stack` into `--print-stack` and introduce `+stack` to control where the stack traces should be generated. Alternatively, something like, keep `--stack` for global activation and introduce  `--print-stack` + `+stack` for finer-grained control?

vlrpl: > > This is one way of doing it; for each additional per-probe option we would end up with a new collect option. The other way would be to allow specifying per-probe options in the cli, e.g. `--probe ip_rcv+stack`.
> 
> yes. That is a good idea. In particular, I thought about this as well (picking a preferred approach is one of the reasons for the RFC tag) but it looked weird for automatic probes (e.g. skb-drop and `skb:kfree_skb`), but maybe it's not weird at all. I guess this would have the advantage of setting the per-probe options without requiring 1:N remap and implicit checks (very nice pros). It would require some sort of special casing e.g. for --stack and `+stack` local options as we should make sure `--stack` is not applied to all the probes but just to the `+stack` ones. OTOH, we can convert `--stack` into `--print-stack` and introduce `+stack` to control where the stack traces should be generated. Alternatively, something like, keep `--stack` for global activation and introduce `--print-stack` (mutually exclusive with `--print-stack`) + `+stack` for finer-grained control?

Not sure about the above, there will be the same problem with --probe-stack.
We should probably think of something like: `if no +stack but just --stack or --probe-stack, set the global option, local opts, otherwise`.

atenart: > > This is one way of doing it; for each additional per-probe option we would end up with a new collect option. The other way would be to allow specifying per-probe options in the cli, e.g. `--probe ip_rcv+stack`.
> 
> yes. That is a good idea. In particular, I thought about this as well (picking a preferred approach is one of the reasons for the RFC tag) but it looked weird for automatic probes (e.g. skb-drop and `skb:kfree_skb`),

What's the issue with automatic probes? The user is free to explicitly add them which in turns allows setting per-probe config options. There are actually two cases:

- Probes reporting events (added by collectors): that's fine, the behavior would be the same. E.g. you can already do `retis collect -c skb-drop -p skb:kfree_skb` which does exactly the same thing as `retis collect -c skb-drop`.
- Probes not reporting events (added by the core and/or collectors): there the user already have to explicitly define such probes to get events. Adding conf options would change anything.

In both solution above the logic is the same: to better control a probe behavior, it must be explicitly set. Otherwise you get the default behavior.

> but maybe it's not weird at all. I guess this would have the advantage of setting the per-probe options without requiring 1:N remap and implicit checks (very nice pros). It would require some sort of special casing e.g. for --stack and `+stack` local options as we should make sure `--stack` is not applied to all the probes but just to the `+stack` ones. OTOH, we can convert `--stack` into `--print-stack` and introduce `+stack` to control where the stack traces should be generated. Alternatively, something like, keep `--stack` for global activation and introduce `--print-stack` (mutually exclusive with `--print-stack`) + `+stack` for finer-grained control?

I'm not sure we need such complexity. If a global *and* a per-probe option exist in parallel (which won't always be the case), I would expect the global option to take precedence and not be a toggle for per-probe flags. So:

- `retis c -p ip_rcv -p ip6_rcv+stack` would print stacks for the second probe only.
- `retis c -p ip_rcv -p ip6_rcv+stack --stack` would print stacks for all probes.

vlrpl: > > > This is one way of doing it; for each additional per-probe option we would end up with a new collect option. The other way would be to allow specifying per-probe options in the cli, e.g. `--probe ip_rcv+stack`.
> > 
> > 
> > yes. That is a good idea. In particular, I thought about this as well (picking a preferred approach is one of the reasons for the RFC tag) but it looked weird for automatic probes (e.g. skb-drop and `skb:kfree_skb`),
> 
> What's the issue with automatic probes? The user is free to explicitly add them which in turns allows setting per-probe config options. There are actually two cases:
> 
>     * Probes reporting events (added by collectors): that's fine, the behavior would be the same. E.g. you can already do `retis collect -c skb-drop -p skb:kfree_skb` which does exactly the same thing as `retis collect -c skb-drop`.
> 
>     * Probes not reporting events (added by the core and/or collectors): there the user already have to explicitly define such probes to get events. Adding conf options would not change anything.
> 
> 
> In both solution above the logic is the same: to better control a probe behavior, it must be explicitly set. Otherwise you get the default behavior. Actually the "report event" behavior is just a per-probe config option, like "stack" would be, there's no difference really (except for being exposed in the cli).
> 

They are fine,
`-p` implies "probe it" as well (that's all the weirdness). But being automatic/hidden there's not much alternative other than being explicit (that's the reason for "but maybe it's not weird at all" below).

> > but maybe it's not weird at all. I guess this would have the advantage of setting the per-probe options without requiring 1:N remap and implicit checks (very nice pros). It would require some sort of special casing e.g. for --stack and `+stack` local options as we should make sure `--stack` is not applied to all the probes but just to the `+stack` ones. OTOH, we can convert `--stack` into `--print-stack` and introduce `+stack` to control where the stack traces should be generated. Alternatively, something like, keep `--stack` for global activation and introduce `--print-stack` (mutually exclusive with `--print-stack`) + `+stack` for finer-grained control?
> 
> I'm not sure we need such complexity. If a global _and_ a per-probe option exist in parallel (which won't always be the case), I would expect the global option to take precedence and not be a toggle for per-probe flags. So:
> 
>     * `retis c -p ip_rcv -p ip6_rcv+stack` would print stacks for the second probe only.
> 
>     * `retis c -p ip_rcv -p ip6_rcv+stack --stack` would print stacks for all probes.

Global over per-probe is fine, it's just a matter of choice (and doc) meaning `--probe-stack`.
Is `+stack` 

> > > This is one way of doing it; for each additional per-probe option we would end up with a new collect option. The other way would be to allow specifying per-probe options in the cli, e.g. `--probe ip_rcv+stack`.
> > 
> > 
> > yes. That is a good idea. In particular, I thought about this as well (picking a preferred approach is one of the reasons for the RFC tag) but it looked weird for automatic probes (e.g. skb-drop and `skb:kfree_skb`),
> 
> What's the issue with automatic probes? The user is free to explicitly add them which in turns allows setting per-probe config options. There are actually two cases:
> 
>     * Probes reporting events (added by collectors): that's fine, the behavior would be the same. E.g. you can already do `retis collect -c skb-drop -p skb:kfree_skb` which does exactly the same thing as `retis collect -c skb-drop`.
> 
>     * Probes not reporting events (added by the core and/or collectors): there the user already have to explicitly define such probes to get events. Adding conf options would not change anything.
> 
> 
> In both solution above the logic is the same: to better control a probe behavior, it must be explicitly set. Otherwise you get the default behavior. Actually the "report event" behavior is just a per-probe config option, like "stack" would be, there's no difference really (except for being exposed in the cli).
> 

They are fine,
`-p` implies "probe it" as well (that's all the weirdness). But being automatic/hidden there's not much alternative other than being explicit (that's the reason for "but maybe it's not weird at all" below).

> > but maybe it's not weird at all. I guess this would have the advantage of setting the per-probe options without requiring 1:N remap and implicit checks (very nice pros). It would require some sort of special casing e.g. for --stack and `+stack` local options as we should make sure `--stack` is not applied to all the probes but just to the `+stack` ones. OTOH, we can convert `--stack` into `--print-stack` and introduce `+stack` to control where the stack traces should be generated. Alternatively, something like, keep `--stack` for global activation and introduce `--print-stack` (mutually exclusive with `--print-stack`) + `+stack` for finer-grained control?
> 
> I'm not sure we need such complexity. If a global _and_ a per-probe option exist in parallel (which won't always be the case), I would expect the global option to take precedence and not be a toggle for per-probe flags. So:
> 
>     * `retis c -p ip_rcv -p ip6_rcv+stack` would print stacks for the second probe only.
> 
>     * `retis c -p ip_rcv -p ip6_rcv+stack --stack` would print stacks for all probes.

The RFC is about filtering out the disjunction between the attached probes and the probes marked as "report the stack trace". The "report the stack trace" flag affects both `--stack` and `--probe-stack options`.

Global over per-probe is fine, it's just a matter of choice (and doc) meaning `--probe-stack`.
But I guess there's still room for misleading behaviors (if not clarified).

What `--probe-stack ... -p <probe1>%stack -p <probe2>%stack` is going to affect? just the display side? overall stack trace report and so probing? (in this second case, again, we're going to handle the case of no `%stack` vs `%stack`).
Depending on the case the runtime may lack (meaning as it is now, but they can be carried in some way or extracted from the cli option) some information required to e.g. display only some stack traces.

atenart: > > > > This is one way of doing it; for each additional per-probe option we would end up with a new collect option. The other way would be to allow specifying per-probe options in the cli, e.g. `--probe ip_rcv+stack`.
> > > 
> > > 
> > > yes. That is a good idea. In particular, I thought about this as well (picking a preferred approach is one of the reasons for the RFC tag) but it looked weird for automatic probes (e.g. skb-drop and `skb:kfree_skb`),
> > 
> > 
> > What's the issue with automatic probes? The user is free to explicitly add them which in turns allows setting per-probe config options. There are actually two cases:
> > ```
> > * Probes reporting events (added by collectors): that's fine, the behavior would be the same. E.g. you can already do `retis collect -c skb-drop -p skb:kfree_skb` which does exactly the same thing as `retis collect -c skb-drop`.
> > 
> > * Probes not reporting events (added by the core and/or collectors): there the user already have to explicitly define such probes to get events. Adding conf options would not change anything.
> > ```
> > 
> > In both solution above the logic is the same: to better control a probe behavior, it must be explicitly set. Otherwise you get the default behavior. Actually the "report event" behavior is just a per-probe config option, like "stack" would be, there's no difference really (except for being exposed in the cli).
> 
> They are fine, `-p` implies "probe it" as well (that's all the weirdness). But being automatic/hidden there's not much alternative other than being explicit (that's the reason for "but maybe it's not weird at all" below).

I would say it means "as a user, I want this to be probed and configured the way I want", knowing that Retis also can have its own probes. Sometimes both the user and Retis want a probe to be there and we merge the options.

> > > but maybe it's not weird at all. I guess this would have the advantage of setting the per-probe options without requiring 1:N remap and implicit checks (very nice pros). It would require some sort of special casing e.g. for --stack and `+stack` local options as we should make sure `--stack` is not applied to all the probes but just to the `+stack` ones. OTOH, we can convert `--stack` into `--print-stack` and introduce `+stack` to control where the stack traces should be generated. Alternatively, something like, keep `--stack` for global activation and introduce `--print-stack` (mutually exclusive with `--print-stack`) + `+stack` for finer-grained control?
> > 
> > 
> > I'm not sure we need such complexity. If a global _and_ a per-probe option exist in parallel (which won't always be the case), I would expect the global option to take precedence and not be a toggle for per-probe flags. So:
> > ```
> > * `retis c -p ip_rcv -p ip6_rcv+stack` would print stacks for the second probe only.
> > 
> > * `retis c -p ip_rcv -p ip6_rcv+stack --stack` would print stacks for all probes.
> > ```
> 
> The RFC is about filtering out the disjunction between the attached probes and the probes marked as "report the stack trace". The "report the stack trace" flag affects both `--stack` and `--probe-stack options`.
> 
> Global over per-probe is fine, it's just a matter of choice (and doc) meaning `--probe-stack`. But I guess there's still room for misleading behaviors (if not clarified).
> 
> What `--probe-stack ... -p <probe1>%stack -p <probe2>%stack` is going to affect? just the display side? overall stack trace report and so probing? (in this second case, again, we're going to handle the case of no `%stack` vs `%stack`). Depending on the case the runtime may lack (meaning as it is now, but they can be carried in some way or extracted from the cli option) some information required to e.g. display only some stack traces.

I think we can either make `--probe-stack` incompatible with individual "stack" options, or convert the global `keep_stack` flag to a per-probe one in `ProbeStack`. The latter should be fine to implement as we already have a list of current probes, and only the initial ones would have a per-probe option set.

For completness the current meaning of `--stack` is to control the display of stack traces. `--probe-stack` controls if we should probe what's reported in stack traces. Then it's the responsibility of Retis to enable or not retrieving stack traces; it could very well always report all stacks and only decide what to do with them (of course enabling that on-demand depending on if one of those stack options is set is way better).

vlrpl: [...]
> > > In both solution above the logic is the same: to better control a probe behavior, it must be explicitly set. Otherwise you get the default behavior. Actually the "report event" behavior is just a per-probe config option, like "stack" would be, there's no difference really (except for being exposed in the cli).
> > 
> > 
> > They are fine, `-p` implies "probe it" as well (that's all the weirdness). But being automatic/hidden there's not much alternative other than being explicit (that's the reason for "but maybe it's not weird at all" below).
> 
> I would say it means "as a user, I want this to be probed and configured the way I want", knowing that Retis also can have its own probes. Sometimes both the user and Retis want a probe to be there and we merge the options.
> 

that's fine and that's obvious to us, but stressing it in the doc (or --help) is probably helpful for the casual user.

> > > > but maybe it's not weird at all. I guess this would have the advantage of setting the per-probe options without requiring 1:N remap and implicit checks (very nice pros). It would require some sort of special casing e.g. for --stack and `+stack` local options as we should make sure `--stack` is not applied to all the probes but just to the `+stack` ones. OTOH, we can convert `--stack` into `--print-stack` and introduce `+stack` to control where the stack traces should be generated. Alternatively, something like, keep `--stack` for global activation and introduce `--print-stack` (mutually exclusive with `--print-stack`) + `+stack` for finer-grained control?
> > > 
> > > 
> > > I'm not sure we need such complexity. If a global _and_ a per-probe option exist in parallel (which won't always be the case), I would expect the global option to take precedence and not be a toggle for per-probe flags. So:
> > > ```
> > > * `retis c -p ip_rcv -p ip6_rcv+stack` would print stacks for the second probe only.
> > > 
> > > * `retis c -p ip_rcv -p ip6_rcv+stack --stack` would print stacks for all probes.
> > > ```
> > 
> > 
> > The RFC is about filtering out the disjunction between the attached probes and the probes marked as "report the stack trace". The "report the stack trace" flag affects both `--stack` and `--probe-stack options`.
> > Global over per-probe is fine, it's just a matter of choice (and doc) meaning `--probe-stack`. But I guess there's still room for misleading behaviors (if not clarified).
> > What `--probe-stack ... -p <probe1>%stack -p <probe2>%stack` is going to affect? just the display side? overall stack trace report and so probing? (in this second case, again, we're going to handle the case of no `%stack` vs `%stack`). Depending on the case the runtime may lack (meaning as it is now, but they can be carried in some way or extracted from the cli option) some information required to e.g. display only some stack traces.
> 
> I think we can either make `--probe-stack` incompatible with individual "stack" options, or convert the global `keep_stack` flag to a per-probe one in `ProbeStack`. The latter should be fine to implement as we already have a list of current probes, and only the initial ones would have a per-probe option set.
> 

Either way `--probe-stack` will keep probing everything with `--stack` and `%stack` affecting only the display side.
I guess it's fine (just different than the RFC behavior), we could also introduce (maybe in the future) a `%probe-stack` option (or whatever fancy name we like :)

> For completness the current meaning of `--stack` is to control the display of stack traces. `--probe-stack` controls if we should probe what's reported in stack traces. Then it's the responsibility of Retis to enable or not retrieving stack traces; 

Certainly, and that was the reason for `--stack-trace-limit` logic (probe and/or display only what you need) with `--probe-stack` and `--stack` considered only as mere on/off.

> it could very well always report all stacks and only decide what to do with them (of course enabling that on-demand depending on if one of those stack options is set is way better).

yeah, limited stack traces :)

vlrpl:
updated to per-probe. Still RFC as, given the `--probe-stack` automatic behaviors, multiple approaches are possible, including cli reparse and/or options incompatibilities (e.g. `--probe-stack` and `/stack`)

retis/src/core/probe/kernel/utils.rs
```diff
@@ -67,11 +89,18 @@ where
...
             Kprobe => Probe::kprobe(symbol)?,
             Kretprobe => Probe::kretprobe(symbol)?,
             RawTracepoint => Probe::raw_tracepoint(symbol)?,
-        })
+        };
+
+        options.iter().try_for_each(|o| -> Result<()> {
+            probe.set_option(o.clone())?;
+            Ok(())
+        })?;
```
atenart, line 101: nit: `options.iter().try_for_each(|o| probe.set_option(o.clone()))?;`

retis/src/core/probe/manager.rs
```diff
@@ -546,10 +566,25 @@ impl ProbeRuntimeManager {
     }
 
     /// Get the list of all currently attached probes.
-    pub(crate) fn attached_probes(&self) -> Vec<String> {
+    pub(crate) fn attached_probes(&self) -> Vec<(String, Vec<ProbeOption>)> {
```
atenart, line 569: nit: not used anymore.

retis/src/core/probe/manager.rs
```diff
@@ -546,10 +566,25 @@ impl ProbeRuntimeManager {
...
         self.probes.clone().into_iter().collect()
     }
 
+    /// Get the list of all currently attached probes with a specific option.
+    pub(crate) fn attached_probes_opt(&self, option: ProbeOption) -> Vec<String> {
```
atenart, line 574: nit: attached_probes_with_opt

retis/src/core/probe/manager.rs
```diff
@@ -112,11 +115,7 @@ impl ProbeManager {
             builder
                 .maps
                 .iter()
-                .try_for_each(|(name, fd)| p.reuse_map(name, *fd))?;
-            builder
-                .global_probes_options
-                .iter()
-                .try_for_each(|o| p.set_option(o.clone()))
+                .try_for_each(|(name, fd)| p.reuse_map(name, *fd))
         })?;
```
atenart, line 119: We should keep this, even if no option is used here. This is key to how global options work.

atenart (changes requested):
I agree with the general goal of the series, and also find parts of it to be nice. However I don't think we should be moving `--stack` to a global option (see the dedicated comment) and IMO this is the key point that makes other things to require quirks and/or to not work as expected.

retis/src/core/probe/manager.rs
vlrpl: yeah, but the suggestion here is (the RFC aims at discussing exactly these aspects), given the per-probe syntax, to use the global map for the global switches and the per-probe for the per-probe CLI options.

With the per-probe opts we're going to have:
- global map (for actual global shared data)
- per-probe map (with per-probe opts)
- global to per-probe (to accommodate the probe stack scenario as it is)

Is there any use case the third bullet can handle that the former two can't?

retis/src/core/probe/kernel/probe_stack.rs
```diff
@@ -107,6 +112,7 @@ impl ProbeStack {
                     _ => return Ok(()),
                 };
 
+                probe.set_option(crate::core::probe::ProbeOption::NoStackTrace)?;
```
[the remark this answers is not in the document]
vlrpl: This was quickly drafted, but the idea is to mimic the current behavior.
I might have overlooked/missed the place where the `ProbeStack` processor remaps global to per-probe options before attaching, and I have no time now to further check it RN, but if memory serves, let's assume:

`retis c --stack --probe-stack`

We're not going to see stack traces for symbols retrieved from stack traces (meaning we have no stacks at all).

Removing:
`probe.set_option(NoStackTrace)?;`

is the equivalent of adding:

`probe.set_option(StackTrace)?;`

To the current code, isn't it?

retis/src/core/probe/bpf/include/common_defs.h
```diff
@@ -24,19 +24,27 @@
...
+	/* Indicates everything is set so the collection can start. */
 	u8 enabled;
-};
+	/* Indicates stack traces must be generated for every probe.
+	 * Potentially valid for USDT probes as well, but
+	 * intentionally ignored.
+	 * It doesn't replace the per-probe configuration, but
+	 * predates it in precedence.
+	 */
+	u8 stack_trace;
```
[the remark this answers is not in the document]
vlrpl: see https://github.com/retis-org/retis/pull/514/commits/2cb73ca3a710117e37bfbeeec09cac8e191e45f7#r2069106020

retis/src/core/probe/probe.rs
```diff
@@ -27,9 +27,21 @@ pub(crate) enum ProbeType {
...
     NoGenericHook,
 }
 
+impl TryFrom<&str> for ProbeOption {
+    type Error = anyhow::Error;
+
+    fn try_from(option: &str) -> Result<Self> {
+        Ok(match option {
+            "stack" => Self::StackTrace,
+            _ => bail!("{option} is an invalid probe option."),
```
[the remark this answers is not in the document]
vlrpl: makes sense

retis/src/core/probe/kernel/probe_stack.rs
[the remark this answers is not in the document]
atenart: Sorry, I overlooked the current state (and thought I tested this, but it seems I didn't). You're right, we're not reporting stack traces from probes installed by `--probe-stack`. That's because the global probe options are only set in probes when transitioning from the builder to the runtime manager. We should definitively fix that.

retis/src/core/probe/kernel/probe_stack.rs
```diff
@@ -119,7 +125,13 @@ impl ProbeStack {
...
         })?;
 
-        if !self.keep_stack {
+        let r#type = match kernel.probe_type.as_str() {
+            "raw_tracepoint" => "tp",
+            s => s,
+        };
+
+        let sym = format!("{}:{}", r#type, kernel.symbol);
+        if !self.keep_stack && !self.stack_trace_probes.contains(&sym) {
```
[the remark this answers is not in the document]
atenart: Same here, I overlooked the current state of the code. But it would prevent probes in `stack_trace_probes` to report stack trace which IMO should be fixed.

retis/src/core/probe/bpf/include/common_defs.h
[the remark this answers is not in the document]
atenart: The above link only opens the full commit log, not sure if it was intended to point to a particular part of it?

retis/src/core/probe/manager.rs
atenart: That's right, the question is whether or not using a global config map for probe options or not. If so, this can be removed.

retis/src/core/probe/bpf/include/common_defs.h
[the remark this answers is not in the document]
vlrpl: I don't know how I ended up linking that :)
The link was supposed to be this:
https://github.com/retis-org/retis/pull/514#discussion_r2069106020

retis/src/core/probe/kernel/probe_stack.rs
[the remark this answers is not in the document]
vlrpl: I guess that the `--stack` case here is covered as long as `probe.set_option(NoStackTrace)?` gets removed.

retis/src/core/probe/manager.rs
vlrpl: The problem is that every time we have to add a probe at runtime, we have to fix up the global switch remapping.
This can be overcome by simply setting the probe opt with `probe.set_option(StackTrace)?` or maybe globally, but in general, this is inherited by the remapping logic, whereas splitting global and per probe doesn't lack this.
One more thing to notice with the remapping is that we cannot enforce any precedence on the eBPF side on specific per-probe conf, as we're funneling everything in a single representation (the cfg probe map).

retis/src/core/probe/kernel/probe_stack.rs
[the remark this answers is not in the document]
atenart: Yes, mapping probes that report stacks to ones that should display it is not covered by our probe options so far. Having a list we maintain here is an option, another one could be having a more generic cache (accessible from the manager) of `Probe` objects which could allow checking their options. There might be other solutions too, I haven't thought too much about that.

retis/src/core/probe/manager.rs
vlrpl: > > The problem is that every time we have to add a probe at runtime, we have to fix up the global switch remapping.
> 
> I'm not sure what do you refer to with "remapping"?
> 

probably "mapping" is more appropriate, but I meant from `--stack` to `ebpf_cfg[symbol].stack` (for each symbol)

> > This can be overcome by simply setting the probe opt with `probe.set_option(StackTrace)?` or maybe globally, but in general, this is inherited by the remapping logic, whereas splitting global and per probe doesn't lack this. One more thing to notice with the remapping is that we cannot enforce any precedence on the eBPF side on specific per-probe conf, as we're funneling everything in a single representation (the cfg probe map).
> 
> We do set probe option globally already, to report stack traces for example.
> 
> As for setting generic flags in probes, we already do this when the manager transition to runtime; it's just missing that probe option fixup when adding probes at runtime. Or am I missing something here?

AFAICT, I didn't see the nitty gritty details, but yes, that's the "generic" way I was hinting at.

atenart: Thanks, I get it now :)

Yeah, that' right, each time we insert a probe we have to set an entry in the per-probe configuration map. But regardless of global settings, we have to do it right? Also it's a one time thing, vs accessing the shared global configuration each time a probe runs.

vlrpl: With the global switch splitting, we avoid the fixup (ebpf handles the two configs), but yes, we have another lookup (as you said, we can consider caching).
Anyways, I don't want to be pushy on the splitting, just wanted to evaluate all the aspects (there is at least another theoretical pro that comes to mind, but that'll probably never happen).
I have to take another look at the "unified conf" and this feature, as I remember there were some issues here and there with that (but maybe I was simply biased with the global config).

vlrpl: > happen). I have to take another look at the "unified conf" and this feature, as I remember there were some issues here and there with that (but maybe I gave in too early to the idea of reusing the global config).

... and the issue is that currently `ProbeOption::StackTrace` is set for each probe (whether --stack and/or `--probe-stack` or `/stack` is set). This is not expressive enough in terms of global/local settings.

In practical terms here:
```rust
        let mut probe_stack = ProbeStack::new(
            collect.stack,
            if !collect.stack {
                self.probes
                    .runtime_mut()?
                    .attached_probes_with_opt(ProbeOption::StackTrace)
            } else {
                Vec::new()
            },
            self.known_kernel_types.clone(),
        );
```

When invoking with `-p ifdump collect --probe-stack -p kp:foo/stack`, `.attached_probes_with_opt(ProbeOption::StackTrace)` will result in all the attached probes (as `--probe-stack` sets the option for every probe).
With the global to local option, we require something like `ProbeOption::StackTrace(bool)` with the bool expressing whether this is global or local (or something different if we want to have a different approach, preferably without carrying around a lot of data aimed at handling just this case).

TL;DR: we map 3:1 (`--probe-stack`,`--stack`,`/stack` to `ProbeOption::StackTrace`) and this is not always enough.

vlrpl: I guess decoupling ReportStack from StackTrace (or ProbeStack) while keeping exactly the same existing approach followed in this series works:

```rust
let mut probe_stack = ProbeStack::new(
            collect.stack,
            if !collect.stack {
                self.probes
                    .runtime_mut()?
                    .attached_probes_with_opt(ProbeOption::ReportStack)
            } else {
                Vec::new()
            },
            self.known_kernel_types.clone(),
        );
```

This is more or less the same as doing `StackTrace(bool|more_structured_date)`, but without the real need to handle inner types/enums.
If there are "local" `/stack` options in some probes with no `--stack`, we'll make `ProbeStack` aware of the probes it should keep the stack trace for reporting (locally set). If `--stack` is set instead, it'll keep all the stack traces for reporting.
Was it something like this you were considering?

atenart: You're right, it's really similar as we'll need to know what to do when inserting new probes. I'm fine with both approaches (embedding the `report` flag in the option or having different options).

vlrpl: tbf, I like the idea of splitting with the perspective you proposed. Looks more natural and less cli-bound.

atenart (changes requested):
In addition to the inline comments there's something that bugs me a bit, which was not introduced by your PR but this is making it more obvious. (Won't be a blocker if we want this in now; as this is working, looking good and the "issue" purely internal).

`ProbeStack` caches information about if the probe stack should be reported and about which probes should report it if the switch was not global. A combination of probe options and cli parameters are used for this, and it makes the `ProbeStack` option to _require_ `ProbeStack` (the struct) to be used. This makes 1) decision on what to do to be scattered 2) not coming from a single source and 3) makes setting some options to only have an effect if used exactly the right way. For example if an other part of Retis would be inserting probes with `ReportStack` at runtime we could have incorrect behavior.

A few pointers to improve the situation:

- Remove the stack from the event only based on options set in a probe, at runtime. Instead of providing a global `keep_stack` and a fixed list of probes; requests to the manager could be done at runtime (and cached).
- Make the "display stack" decision outside `ProbeStack`; could be a function fixing events before printing, or could be a hidden flag in the stack event section to prevent from displaying them (in turn would also prevent them from being written to an event file).
- I don't have a particular preference but maybe using probe options tied to the collect and post-process behavior instead of mapped to cli options (`StackTrace` to enable retrieving stacks and `PrintStack` to display them; instead of `ProbeStack` and `ReportStack`) would make the above easier too. It could also allow more future use cases, e.g. reporting stacks w/o printing them for processing other than probing at runtime (e.g. differentiating from paths leading to a probe).

WDYT?

(Note that I can do the above in a follow-up if we agree on the direction).

retis/src/collect/collect.rs
```diff
@@ -544,7 +549,13 @@ impl Collectors {
...
         let mut probe_stack = ProbeStack::new(
             collect.stack,
-            self.probes.runtime_mut()?.attached_probes(),
+            if !collect.stack {
+                self.probes
+                    .runtime_mut()?
+                    .attached_probes_with_opt(ProbeOption::ReportStack)
+            } else {
+                Vec::new()
+            },
```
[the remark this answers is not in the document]
vlrpl: No, it's not needed; it's just a way to avoid a useless iteration

retis/src/core/probe/kernel/probe_stack.rs
```diff
@@ -107,6 +129,10 @@ impl ProbeStack {
                     _ => return Ok(()),
                 };
 
+                mgr.global_options()
+                    .iter()
+                    .try_for_each(|o| probe.set_option(o.clone()))?;
```
atenart, line 134: I understand what you meant in the discussion now. I would prefer this to be handled in the runtime manager before attaching probes as this is already the behavior when using the `ProbeBuilderManager` and this would make the behavior different based on its state. I actually see the lack of this as a bug in my original manager implementation. Also we might even be able to consolidate things and use the same logic (once moved to the runtime manager, and not while converting the manager type).

If for some reason we really want a probe not to follow a global setting, we should add an API for that instead (when the need arises).

retis/src/collect/collect.rs
```diff
@@ -543,8 +548,9 @@ impl Collectors {
 
         let (mut iccount, mut eccount) = (0, 0);
         let mut probe_stack = ProbeStack::new(
-            collect.stack,
-            self.probes.runtime_mut()?.attached_probes(),
+            self.probes
+                .runtime_mut()?
+                .attached_probes_with_opt(ProbeOption::ProbeStack),
```
atenart, line 553: Sorry, missed that. This is not right as what we're looking for is the full list of probes being attached, otherwise if a probe w/o `ProbeStack` is reported in a stack trace, attaching a probe at runtime to it will fail. This chunk can be dropped.

retis/src/core/probe/kernel/probe_stack.rs
```diff
@@ -44,11 +41,36 @@ impl ProbeStack {
...
+
+        let keep_stack = if let Some(opts) = mgr.get_probe_opts(&sym) {
+            opts.contains(&crate::core::probe::ProbeOption::ReportStack)
+        } else {
+            false
+        };
+
+        if keep_stack {
+            self.display_stack_cache.insert(sym);
+        }
```
atenart, line 69: This means each time a probe is not reporting a stack trace, we'll do the lookup for each event? We should cache both true and false values to have a 100% hit.

atenart (changes requested):

retis/src/core/probe/kernel/probe_stack.rs
vlrpl: I honestly see no value in caching here. I added it right after pushing. (How many options does a probe have to make the caching worth it?)

vlrpl: we rather have a double lookup in the worst case

retis/src/collect/collect.rs
vlrpl: This had to be dropped. It is already in my local version, not sure why was not part of this version.

retis/src/core/probe/kernel/probe_stack.rs
vlrpl: also caching makes things less dynamic. What if the runtime starts removing probes? What if something cached in the false becomes true because it gets added with that option?

atenart: Right, if the probe options are changed at runtime or removed, we would have incorrect information. This is not supported ATM by the manager and would require heavy changes though.

I don't have strong opinion wrt to caching vs not caching. I guess we're looking at a single local lookup vs a lookup + iter + check. Idk if the difference in performances is noticeable.

vlrpl: > Right, if the probe options are changed at runtime or removed, we would have incorrect information. This is not supported ATM by the manager and would require heavy changes though.
>
> I don't have strong opinion wrt to caching vs not caching. I guess we're looking at a single local lookup vs a lookup + iter + check. Idk if the difference in performances is noticeable.

it's 1 lookup + iter (constant though) vs [3 lookup (worst case); 1 lookup best case (let's say for the true case); 2 lookups (for the false case)]

The point is that the constant iteration is probably negligible (if we have tons of stacktraces, we're going to have a performance drop for other reasons)

atenart: Makes sense, let's remove the cache.

retis/src/core/probe/manager.rs
```diff
@@ -547,7 +561,17 @@ impl ProbeRuntimeManager {
...
     /// Get the list of all currently attached probes.
     pub(crate) fn attached_probes(&self) -> Vec<String> {
-        self.probes.clone().into_iter().collect()
+        self.probes.keys().cloned().collect()
+    }
+
+    /// Get the list of all current global options.
+    pub(crate) fn global_options(&self) -> Vec<ProbeOption> {
+        self.global_probes_options.clone().into_iter().collect()
+    }
```
atenart, line 570: This is added in patch 1 and then the only user is removed.

retis/src/core/probe/kernel/probe_stack.rs
```diff
@@ -44,11 +38,25 @@ impl ProbeStack {
...
+            s => s,
+        };
+
+        let sym = format!("{}:{}", r#type, evt.symbol);
+
+        if let Some(opts) = mgr.get_probe_opts(&sym) {
+            opts.contains(&crate::core::probe::ProbeOption::ReportStack)
+        } else {
+            false
+        }
```
atenart, line 57: nit (since there were other comments): I think this can be simplified to:

```rust
mgr.get_probe_opts(&sym).is_some_and(|opts| opts.contains(&crate::core::probe::ProbeOption::ReportStack))
```

(Super nit: you might want to import `ProbeOption` instead of using the fully qualified name).

retis/src/core/probe/manager.rs
```diff
@@ -112,11 +111,7 @@ impl ProbeManager {
             builder
                 .maps
                 .iter()
-                .try_for_each(|(name, fd)| p.reuse_map(name, *fd))?;
-            builder
-                .global_probes_options
-                .iter()
-                .try_for_each(|o| p.set_option(o.clone()))
+                .try_for_each(|(name, fd)| p.reuse_map(name, *fd))
```
atenart, line 114: Commenting here as it's related: we're also missing a `reuse_map` call for probes inserted at runtime, otherwise if they contain specific hooks (which is not the case ATM) those won't work as expected. The fix is the same (calling `reuse_map` before loading the probe) and then I guess the above can be removed too.

retis/src/core/probe/manager.rs
vlrpl: argh, I squashed the wrong patches

atenart (approved):
LGTM, thanks!

PR #518, opened 2025-04-16
Profiles and Python script improvements

atenart:
A few improvements related to profiles and python scripts; mainly:

- Stop using /etc to store Retis files[1].
- Allow to specify profiles using a full path.
- Support using a custom profiles directory.
- Support storing Python script in known directories.

[1] I used `/usr/share/retis`, but feel free to share better suggestions.


retis/src/cli/cli.rs
```diff
@@ -165,6 +167,12 @@ pub(crate) struct MainConfig {
 - Path to a profile, in which case the file should contain a single profile."
     )]
     pub(crate) profile: Vec<String>,
+    #[arg(
+        long,
+        help = "Path to a directory with custom profiles. When used this overrides the
```
amorenoz, line 172: nit: /When used/When used,/?

docs/index.md
```diff
@@ -153,6 +153,14 @@ like any other profile if stored in `$HOME/.config/retis/profiles`. Here is an
 profile](https://github.com/retis-org/retis/blob/main/retis/test_data/profiles/example.yaml)
 with inlined comments. If a profile is generic enough, consider contributing it!
 
+A custom directory containing profiles can be used. This is handy for sharing
+out-of-tree profiles.
```
amorenoz, line 157: If the goal is to share out-of-tree profiles, why not adding the directory to the profile list instead of replacing it? something like "--extra-profile-dir" so "generic" and all other in-tree ones can also be accessed.

retis/src/process/cli/python.rs
```diff
@@ -16,12 +16,53 @@ pub(crate) struct PythonCli {
         help = "File from which to read events"
     )]
     pub(super) input: PathBuf,
-    #[arg(help = "Python script to execute. Omit to drop into an interactive shell.")]
+    #[arg(
+        help = "Python script to execute. Omit to drop into an interactive shell.
+Alternatively scripts can be stored in $HOME/.config/retis/python and
+/usr/share/retis/python, in which case the file name only can be
+provided."
```
amorenoz, line 23: Maybe mention the fact we'll strip the ".py"?

PR #522, opened 2025-05-09
Pcap fix timestamp

vlrpl:
Timestamps in pcapng generated with the pcap command are essentially broken.
The series fixes both a resolution problem and applies the offset from the startup event.
Now the timestamps look reliable compared to captures taken in parallel.


atenart (approved):
LGTM, thanks!

PR #524, opened 2025-05-20
Update pnet_packet to 0.35

musicinmybrain:
Upstream release notes / changelogs: https://github.com/libpnet/libpnet/releases/tag/v0.35.0

While this is a SemVer-breaking update for `pnet_packet`, it doesn’t appear that any code changes are required, and `make test` and `make pytest` still pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -963,18 +963,18 @@ checksum = "b4596b6d070b27117e987119b4dac604f3c58cfb0b191112e24771b2faeac1a6"
 
 [[package]]
 name = "pnet_base"
-version = "0.34.0"
+version = "0.35.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


atenart:
The c8s CI error is expected and should be fixed by #521

Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
+version = 4
```
atenart, line 3: Can you remove this chunk as this will break compilation on c8s? We have a fix for this in #521 but it's not merged yet.

musicinmybrain: Sure. I wouldn’t normally hand-edit a `Cargo.lock` file, but it ought to be safe enough to do so in this case.

atenart: Thanks! Agreed, that is not ideal and only a temporary solution. Hopefully the real fix will land soon.

atenart (approved):
LGTM, thanks!

PR #525, opened 2025-05-21
Retrieve the netns cookie in addition to the inum

atenart:
It provides a more stable identifier.

In addition the last commit is making the retrieval of the dev & ns sections always-on (dev was the default value of `--skb-sections` already). Rationale is in the commit log. This is entirely up for discussion. Here, note that with #485 the packet display should be on its own line for easier readability.

One thing I'm wondering is if we should invert the dev and ns order in the output for easier readability, from

`ns 0x1 (4026531840) if 4 (wlp82s0)`

to

`if 4 (wlp82s0) ns 0x1 (4026531840)`


atenart:
`net_cookie` isn't there in older kernels (c8s), we might make it optional after all...

atenart:
Fixed a conflict with recently merged commits.

retis/src/collect/collector/skb/bpf.rs
```diff
@@ -303,9 +312,32 @@ fn unmarshal_l4(
 pub(crate) struct SkbEventFactory {
     // Should we report the Ethernet header.
     pub(super) report_eth: bool,
+    // Does the kernel support net cookies?
+    net_cookie: bool,
 }
 
 impl SkbEventFactory {
+    pub(crate) fn new() -> Result<Self> {
```
vlrpl, line 129: I understand `u64` is not exactly the ideal type for optional fields, but this feels like a duplicate logic, and it is, as a matter of fact.
Can come up with a quick and effective alternative either.
Yeah, we can have a presence bitfield or an entire byte or whatever (there should be room at the end of `struct skb_netns_event` anyway), but it's not perfect (especially in the C bitfield flavor as they are not ideal in Rust), or we can simply always show cookie 0 (not perfect as well).
The point is that it's ok for one (as we have today), but maybe we may think to revisit if this need (optional u64) will show up again.

atenart: I'm not sure to understand the suggestion. Would you prefer if the information was coming from the BPF part?

docs/collectors/skb.md
```diff
@@ -30,9 +30,12 @@ or invalid fields are automatically hidden.
 ### Ns event section
 
 ```none
-ns {namespace id}
+ns [{unique id}/]{inum}
 ```
 
+The unique id is generated by the kernel and guaranteed not to be reused. It
+might not be available on older kernels.
```
amorenoz: I think we should explain what is `inum` then. Before it was kind of clear "namespace id", but now it's not clear (for someone that does know kernel internals).

retis/src/collect/collector/skb/bpf/skb_hook.bpf.c
```diff
@@ -245,28 +247,38 @@ static __always_inline int process_skb(struct retis_raw_event *event,
 
 	if (cfg->sections & BIT(SECTION_NS)) {
 		struct skb_netns_event *e;
-		u32 netns;
+		bool get_cookie;
+		u64 cookie = 0;
+		u32 inum;
+
+		/* Netns cookies are not available on older kernels */
```
amorenoz: nit: full stop?

amorenoz (commented):
Small comments, otherwise LGTM

retis/src/collect/collector/skb/bpf.rs
vlrpl: yes, avoid duplication, but it's just one case, so it's fine

vlrpl (approved):
lgtm, I leave it up to @amorenoz for a last check

PR #526, opened 2025-05-21
tree: use tracefs instead of debugfs when possible

atenart:
Initially tracefs was part of debugfs and only accessible when debugfs was mounted. However this changed and tracefs can now be mounted even if debugfs is not. This is now a common behavior by default in many distributions.

This is indeed better; in Retis try using the standalone tracefs mount instead of the nested one.

### DIFF
--- a/retis/src/collect/collect.rs
+++ b/retis/src/collect/collect.rs
@@ -4,6 +4,7 @@ use std::{
     collections::{HashMap, HashSet},
     fs::OpenOptions,
     io::{self, BufWriter},
+    path::{Path, PathBuf},
     process::{Command, Stdio},
     sync::Arc,
     time::Duration,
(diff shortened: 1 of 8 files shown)


retis/src/collect/collect.rs
```diff
@@ -108,8 +109,8 @@ pub(crate) struct Collectors {
     tracking_config_map: Option<libbpf_rs::MapHandle>,
     // Retis events factory.
     events_factory: Arc<RetisEventsFactory>,
-    // Did we mount debugfs ourselves?
-    mounted_debugfs: bool,
+    // Did we mount a fs ourselves? If so contains the target dir.
```
vlrpl, line 112: small nit: this can be tracefs or debugfs only. Isn't `fs` too generic?

vlrpl (commented):
Thanks for taking care of this.
I have just a [very] small nit. Other than that the patch lgtm.

retis/src/collect/collect.rs
atenart: Fixed!

vlrpl (approved):
LGTM, thanks!

PR #529, opened 2025-06-11
[RFC] Introduce stack tracking

vlrpl:
The series is a DRAFT/RFC aiming at discussing the way this was approached and how this should be shaped.
So far, the stack address has been added, extending the kernel section (also present in skb tracking).
The initial approach is conservative (we'll expand in follow-ups, if needed) as kretprobes and ovs hooks can be optimized (carrying their context inside the stack tracking map).
This should fix a bunch of issues we had. Aarch64 remains completely untested, but the init_stack trick should work there as well.

Discussions about details or style are not required for this iteration, as this will probably change anyway.
The series is intentionally incomplete (some FIXME are also present) as the agreement around the events may change the additions required (series, post processing, ...).


vlrpl:
> I had an initial look at this and this is very promising!
> 
> The stack tracking logic is heavily integrated into the skb, for good reasons. In general in Retis, additional tracking mechanisms would probably be as well as our goal is to link events to packets. (That doesn't prevent the stack tracking logic from helping in internal things too).
> 
> Because of this IMO we could integrate it a bit more and use the stack tracking id as an indirection to the skb one (basically storing the skb tracking key in the stack tracking map as the value, and maintaining it). We could then issue skb-tracking information in all events (using the `skb-tracking` collector) and stop reporting the stack id as part of the kernel event section. This would make it clear that an event is linked to an skb.
> 
> It would also make the following cases to work:
> 
>     * Automatically enable the `sort` command to group things.
> 
>     * In case the stack id changes over the life of an skb (remote free, skb put on a list and processed later, etc) the same tracking information would still be reported, allowing to identify the event in the whole flow.
> 

That's one of the options (together with a separate collector that I don't really like, TBF), and probably a good one.
There were some things to consider about it:
- whether we want to consider stack tracking a superset of skb tracking or not
- known types filter (we should just lift the "let's not probe functions w/ unknown parameters"; Not sure about the `--probe-stack` feature, but it's safer to avoid probing non-skb functions there)
- The reported skb address in the event will always be NULL (not a real issue, but we might want to avoid printing out the skb at all if 0)

> 
> A few other high level comments:
> 
>     * While I understand why allowing probes w/o hooks I don't think this should be part of this series. It is a different feature that might need some thinking (per-probe option, global option, some magic, ?). The stack tracking feature is useful regardless as some collectors allow retrieving information from non-skb enabled probes.
> 
>     * For v1 the `sort` command and the `SkbTrackingEvent` Python bindings should support using and/or comparing stack ids. IMO this is important from the start as it will help testing the feature and might as well unveil some limitations or small issues in the design (if any).

Of course, the lack of them is called out in the PR description, and it's intentional for the RFC. For v1, it's a no-brainer; they have to be part of the series, but if we go with an skb-tracking event, things significantly simplify there

atenart:
> That's one of the options (together with a separate collector that I don't really like, TBF), and probably a good one. There were some things to consider about it:
> 
> * whether we want to consider stack tracking a superset of skb tracking or not

But can this work without relying on another tracking mechanism to help distinguish between flows? ATM w/o the skb tracking information we would not detect the stack id being reused.

Another thing to consider is the use case. The main goal is really to get a packet at some point and not to issue events that can't be linked to skbs. If that holds true there's no real need to go in splitting it from skb tracking.

> * The reported skb address in the event will always be NULL (not a real issue, but we might want to avoid printing out the skb at all if 0)

Not if the tracking information reported comes from the tracking information stored in the tracking map.

retis/src/collect/collect.rs
```diff
@@ -399,23 +404,10 @@ impl Collectors {
...
-                    .parameter_offset(t)
-                    .is_ok_and(|offset| offset.is_some())
-            });
-            if !ok {
-                info!(
-                    "No probe was attached to {symbol} as no collector could retrieve data from it"
-                );
-            }
-            ok
-        };
```
atenart, line 415: I would not remove this as part of this PR, this is not related to stack tracking and needs to be discussed.

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -53,6 +56,11 @@ struct {
 	__type(value, struct tracking_info);
 } tracking_map SEC(".maps");
 
+static __always_inline struct tracking_info *__skb_tracking_info(u64 *key)
```
atenart: nit: I would instead rename the existing one with an `_skb` suffix or something like that.

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -137,17 +147,39 @@ static __always_inline int track_skb_start(struct retis_context *ctx)
...
+	 * deallocation).
+	 */
+	if (!free) {
+		if (ti->stack_ref != ctx->stack_base)
+			ti->stack_ref = ctx->stack_base;
+
+		if (!track_stack_start(ctx->stack_base,
+				       inv_head
+				       ? (u64)skb
+				       : (u64)head))
```
atenart: We also need to update the stack tracking map with the new `head` when seen. This is done for the skb tracking part below the "If found, index it by its data address from now on..." comment.

retis/src/core/tracking/bpf/include/stack_tracking.h
```diff
@@ -0,0 +1,104 @@
...
+	val = bpf_map_lookup_elem(&stack_tracking_map, &stack_base);
+	if (!val)
+		return 0;
+
+	return *val;
+}
+
+static __always_inline bool stack_is_tracked(u64 stack_base)
+{
+	return get_stack_id(stack_base) > 0;
```
atenart: nit: since the function returns an `u64` we can just check it's not 0; that might be better for the compiler?

retis/src/core/tracking/bpf/include/stack_tracking.h
```diff
@@ -0,0 +1,104 @@
...
+{
+	u64 addr;
+
+	switch (type) {
+	case KERNEL_PROBE_KRETPROBE:
+		fallthrough;
+	case KERNEL_PROBE_KPROBE:
+		if (!bpf_core_enum_value_exists(enum bpf_attach_type___compat,
+					       BPF_TRACE_KPROBE_MULTI) ||
+		    !bpf_core_field_exists(((struct bpf_kprobe_multi_link *)0)->cookies)) {
```
atenart: We have a similar check in `int probe_kprobe` (the one above is even better), let's factorize both into an helper defined in `retis/src/core/probe/kernel/bpf/include/helpers.h`.

retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -346,8 +349,17 @@ static __always_inline int chain(struct retis_context *ctx)
 	 * logic runs even if later ops fail: we don't want to miss information
 	 * because of non-fatal errors!
 	 */
-	if (RETIS_TRACKABLE(ctx->filters_ret))
+	if (RETIS_TRACKABLE(ctx->flags))
 		track_skb_start(ctx);
+	else if (skb)
```
atenart, line 351: Maybe we can also check on `flags & RETIS_F_STACK_PASS` to save a map lookup when the stack is not tracked?

retis/src/core/tracking/bpf/include/stack_tracking.h
```diff
@@ -0,0 +1,104 @@
...
+
+enum bpf_attach_type___compat {
+	BPF_TRACE_KPROBE_MULTI,
+};
+
+const volatile unsigned int THREAD_SIZE;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_LRU_HASH);
+	__uint(max_entries, 8192);
```
atenart, line 17: Should we define this in a shared macro to keep the skb tracking and stack tracking map in sync wrt the number of entries?

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -137,17 +147,39 @@ static __always_inline int track_skb_start(struct retis_context *ctx)
...
 
+	/* If the skb gets tracked but the stack_base doesn't match it
+	 * may mean that a packet got queued and handled in a
+	 * different context in terms of stack.  cfg->free is an
+	 * exception as we want to keep the old reference and consume
+	 * it to delete the original stack_id entry in the case of
+	 * deallocation happening in different contexts (e.g. deferred
+	 * deallocation).
+	 */
+	if (!free) {
```
atenart: With the tight integration between the skb and stack tracking, we can probably remove this check and let the hooks access stack tracking information if they want, even in a free function. This would align with the skb tracking behavior.

retis/src/collect/collect.rs
vlrpl: I agree, no real need, this probably slipped when converting from kernel section evt to skb tracking evt

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: I may miss something but:
`RETIS_F_STACK_PASS` is set in the non-skb case, but here we're in the existing but not trackable skb case.

retis/src/core/tracking/bpf/include/stack_tracking.h
vlrpl: we might, the concern I have is the garbage collection strategy (LRU vs userspace GC).
I guess it's safer to do so if we decide to move skb tracking to LRU as well.
WDYT?

retis/src/core/tracking/bpf/include/skb_tracking.h
vlrpl: `s/skb_tracking_info/skb_tracking_info_skb()` ?
not sure about the double skb.

What about `s/__skb_tracking_info/skb_tracking_info_lookup()/` ?

retis/src/core/tracking/bpf/include/stack_tracking.h
vlrpl: I guess `return get_stack_id(stack_base)` implicitly does something equivalent to `return !!get_stack_id(stack_base)` because of the `bool`. 
Not sure what's better, but makes sense anyways.

retis/src/collect/collect.rs
vlrpl: well, looking at it twice, without it we have:

```
No probe was attached to ip_send_check as no collector could retrieve data from it
```

that is, this is still relevant.
Maybe we should start that discussion or find a lightweight alternative :)

retis/src/core/tracking/bpf/include/skb_tracking.h
vlrpl: This should be covered (we're still referencing that with `head`). What I think is not covered is:

```diff
diff --git a/retis/src/core/tracking/bpf/include/skb_tracking.h b/retis/src/core/tracking/bpf/include/skb_tracking.h
index c22f4c907..409bdc651 100644
--- a/retis/src/core/tracking/bpf/include/skb_tracking.h
+++ b/retis/src/core/tracking/bpf/include/skb_tracking.h
@@ -170,10 +170,7 @@ static __always_inline int track_skb_start(struct retis_context *ctx)
                if (ti->stack_ref != ctx->stack_base)
                        ti->stack_ref = ctx->stack_base;
 
-               if (!track_stack_start(ctx->stack_base,
-                                      inv_head
-                                      ? (u64)skb
-                                      : (u64)head))
+               if (!track_stack_start(ctx->stack_base, (u64)head))
                        log_error("While tracking stack. Unable to update the entry");
        }
 
@@ -183,8 +180,10 @@ static __always_inline int track_skb_start(struct retis_context *ctx)
        /* If the function invalidates the skb head, we can't know what will be
         * the new head value. Temporarily track the skb using its skb address.
         */
-       if (inv_head)
+       if (inv_head) {
                bpf_map_update_elem(&tracking_map, (u64 *)&skb, ti, BPF_NOEXIST);
+               track_stack_start(ctx->stack_base, (u64)skb);
+        }
 
        return 0;
 }
```

Are you referring to this?
Of course, the diff above cannot work as we cannot track both the new head and the skb address (WRT the same stack) unless we introduce some kind of list for that.

retis/src/core/tracking/bpf/include/skb_tracking.h
vlrpl: looks sane to me, and also the comment requires an update (this part no longer applies):
```
	 * cfg->free is an
	 * exception as we want to keep the old reference and consume
	 * it to delete the original stack_id entry in the case of
	 * deallocation happening in different contexts (e.g. deferred
	 * deallocation).
```

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -73,10 +76,11 @@ static __always_inline struct tracking_info *skb_tracking_info(struct sk_buff *s
 
 static __always_inline int track_skb_start(struct retis_context *ctx)
 {
-	bool inv_head = false, no_tracking = false;
+	bool inv_head = false, no_tracking = false, free = false;
 	struct tracking_info *ti = NULL, new;
 	struct tracking_config *cfg;
 	u64 head, ksym = ctx->ksym;
+	bool deferred_update = false;
```
amorenoz: nit: xmas tree

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -137,17 +142,31 @@ static __always_inline int track_skb_start(struct retis_context *ctx)
...
-		bpf_map_update_elem(&tracking_map, &head, &new, BPF_NOEXIST);
+		deferred_update = true;
 	}
 
 	/* Track when we last saw this skb, as it'll be useful to garbage
 	 * collect tracking map entries if we miss some events.
 	 */
 	ti->last_seen = ctx->timestamp;
 
+	/* If the skb gets tracked but the stack_base doesn't match it
```
amorenoz: nit: match, it

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -190,6 +209,30 @@ static __always_inline int track_skb_end(struct retis_context *ctx)
...
+	 * - kp:non_skb_sym1 -> kp:skb_sym
+	 *
+	 * Potentially, even the case of delete stack entries through
+	 * deferred deallocations:
+	 *
+	 * - kp:non_skb_sym1 -> kp:skb_sym -- deferred --> kfree_skb
+	 *
+	 * is affected to a lesser extent by false positives (with the
+	 * overhead of synchronization).
+	 *
```
amorenoz: nit: extra newline

retis/src/core/tracking/bpf/include/stack_tracking.h
```diff
@@ -0,0 +1,74 @@
+#ifndef __CORE_FILTERS_STACK_TRACKING__
+#define __CORE_FILTERS_STACK_TRACKING__
```
amorenoz: nit: why FILTERS?

retis/src/core/tracking/bpf/include/stack_tracking.h
```diff
@@ -0,0 +1,74 @@
...
+struct {
+	__uint(type, BPF_MAP_TYPE_LRU_HASH);
+	__uint(max_entries, 8192);
+	__type(key, u64);
+	__type(value, u64);
+} stack_tracking_map SEC(".maps");
+
+struct stack_addr {
+	u64 addr;
+};
```
amorenoz: A wrapper struct seems unnecessary

retis/src/core/probe/kernel/bpf/kprobe.bpf.c
```diff
@@ -46,11 +46,9 @@ int probe_kprobe(struct pt_regs *ctx)
 		context.ksym = kprobe_get_func_ip(ctx);
 
 	if (kretprobe) {
-		u64 tgid = bpf_get_current_pid_tgid();
-
 		/* Store the current context and let the kretprobe run the hooks. */
-		if (bpf_map_update_elem(&kretprobe_context, &tgid, &context,
-					BPF_NOEXIST))
+		if (!bpf_map_update_elem(&kretprobe_context, &context.stack_base, &context,
+					 BPF_NOEXIST))
```
amorenoz: why fail on success (return  0)?

vlrpl: no reason, probably a typo after conflict resolution.

retis/src/core/tracking/bpf/include/stack_tracking.h
vlrpl: this is something I wanted to revisit (and double check), but this was initially introduced because of a verification error.

retis/src/core/tracking/bpf/include/stack_tracking.h
vlrpl: my guess is completion&bad deletion :D

retis/src/core/tracking/bpf/include/skb_tracking.h
vlrpl: ahh, our new target `make check-ebpf` failed with this :D

retis/src/collect/collect.rs
atenart: What Retis command did you use to get the above?

retis/src/core/tracking/bpf/include/skb_tracking.h
atenart: Or `by_skb`. IMO the idea would be to distinguish between the two variants using a different parameter as the input. The one using an `u64 key` is now the normal API, the other one, skb based, is an helper provided for ease of use. The name should reflect it in some way.

retis/src/core/tracking/bpf/include/skb_tracking.h
atenart: Ah right, it already works because the entry is updated every time `track_skb_start`, not just when the entry doesn't exist yet.

Yes the above is something that doesn't work as we can't track both entries at once. I think that's fine (a comment could be added if you want) and choosing the `skb` over the `head` when invalidating the head makes sense. We would just have false negatives of the invalidation fails but that shouldn't be too bad. A real fix would be to handle the invalidation in a kretprobe / fexit but that's out of scope of this PR.

retis/src/core/probe/kernel/bpf/include/common.h
atenart: That's right.

amorenoz:
To  try to summarize what I expressed in the chat:

I have a concern with how stack tracking allows us to generate events for non-skb functions. AFAICS, the current implementation does the following:
- start stack tracking when an skb function satisfies the filter
- stop stack tracking when an skb function does not satisfy the filter or when the tracked skb is freed
- generate events for non-skb functions if stack is being tracked

I think this makes tracking of non-skb functions quite confusing IMHO. Stack tracking would stop in an undetermined point in the future (skbs can live long and the thread might not process any other skb in a long time) generating a potentially huge amount of noise.

Let's say we have:
```
some_skb_func(skb) {
   -> foo()
       -> bar()
}
foo()
  -> bar() 
foo()
   -> bar()  
```

I see an extremely powerful use-case in being able to generate events from `foo` and `bar` **when called inside**  `some_skb_func(skb)` with an skb that satisfies my filter **but not outside**. E.g:
- determine what condition caused an error that leads to my packet being dropped or sent somewhere unexpected
- measuring what causes an increase latency inside my function, e.g: mutex_lock?

However, having the stack tracking "spill" outside of `some_skb_func` is confusing to me because it's not clear when we'll stop tracking nor when was `foo` or `bar` the function called.

So, my suggestion would be to add an option to a probe to enable stack tracking, let's call it **ftrace** because of its similarity to one of the coolest kernel debugging features. With this, running:
```
retis collect -p kprobe:some_skb_func/ftrace -p foo -p bar
```
will generate exactly what I wanted assuming:
- stack tracking is started when `some_skb_func` is called with with an skb that satisfies the filter or is already skb-tracked
- stack tracking is stopped when `some_skb_func` returns

Based on this, the implementation would have to use kprobe/kretprobe mechanism which in turns imposes the limitation of not being able to start/stop tracking on tracepoints, which I think it's OK (see note 2 below).

This would give clarity on what events we'll actually get reported and significantly reduce the noise. Would it be possible to also generate events which are outside of `some_skb_func`? Yes, we could easily also allow non-skb functions to start/stop the stack tracking so it'd be as easy as using `some_skb_func`'s parent as trigger.

Note 1: None of the above applies to using stack pointer instead of pid to track a thread in both kprobe/retprobe and OVS collector. This is is very much needed and an indisputable improvement.

Note 2: A more advanced (confusing and error prone) option would be to also stop tracking explicitly, e.g:
```
retis collect -p some_probe/ftrace_start -p some_other_probe/ftrace_stop
```
This would bring non-determinism back into the equation without a huge benefit other than what seems to me as extreme cases so I'd only consider it as a follow up improvement if we really see it necessary.

retis/src/collect/collect.rs
vlrpl: Not exactly on top of this series, but a reasonably safe update.

`sudo ./target/debug/retis collect -c skb,skb-tracking -p "k:__ip_local_out" -p "kr:__ip_local_out" -p "k:ip_send_check" -f "icmp"`

atenart: That is expected as we do not support probing functions we can't extract data from. This was my initial point, while it could be useful, that is an extra feature on top of the stack tracking one. Or am I missing something?

atenart:
If I sum up the above comment we would (did I got this right?):

1. Keep the existing stack tracking dance to allow internal use of it in specific known places (kretprobe, ovs tracking).
2. Do not use and report stack tracking for events unless explicitly enabled.

A few comments:

1/

> ```
> retis collect -p some_probe/ftrace_start -p some_other_probe/ftrace_stop
> ```
This would be quite unreliable as in many cases it's not guaranteed to reach the stop probe (e.g. skb is dropped in between).

2/

While I agree with the concerns and the solution for tracking functions called from within a selected probe, this won't work for tracking things like `netdev_core_stats_inc` in a generic way. This can be called from various paths and it's not possible to come up with a list of functions to enable stack tracking.

I haven't thought about this too much and do not have solution for this yet. Or maybe since the stack tracking logic would always be on (for internal use) we can have two modes: the one proposed above to trace calls from a given function and another one where stack tracking could be forced on for a specific probe (opt-in)?

atenart:
> 1. Keep the existing stack tracking dance to allow internal use of it in specific known places (kretprobe, ovs tracking).

If we agree on that we can make this PR about that and merge it, to avoid keeping those patches oot for too long. Then we could focus on what / how to report events. Up to you of course.

amorenoz:
@atenart 
your example evades me: how would you like to track  `netdev_core_stats_inc` exactly?

If you want an event every time a particular packet causes a netdev stat increase you have the problem you've mentioned: it's called from many places, but with the current implementation you have an additional problem.
Right now, you need a probe with an skb to start stack-tracking, so you still need to guess where you're packet will first appear (that's essentially the problem you've mentioned: not sure where it'll be called).
But you have another problem, false positives: you might generate an event way after your probe has returned and you're running some code that has nothing to do with your packet.

The second problem does not exist in my proposal because there is clarity on where the tracking is disabled.

The first problem still persists, we only have events if we try to guess where `netdev_core_stats_inc` was called.
However, for this problem we could combine the global stack-tracking with `--probe-stack` option (which, IMHO, should also be on a per-probe basis), imagine:
```
retis collect --probe-stack --ftrace -p netdev_core_stats_inc -f "port 9999"
```
Every time `netdev_core_stats_inc` is hit, the stack will be traversed and probes with `ftrace` added to all skb-related funcs.

This feels like black magic now

vlrpl:
> > 1. Keep the existing stack tracking dance to allow internal use of it in specific known places (kretprobe, ovs tracking).
> 
> If we agree on that we can make this PR about that and merge it, to avoid keeping those patches oot for too long. Then we could focus on what / how to report events. Up to you of course.

Yes, that's of course preferable.
If we decide not to allow probing of non-SKB functions (as appears to be so far), then it seems reasonable to have all discussions about how to present the new feature user-side afterward.

retis/src/collect/collect.rs
vlrpl: > That is expected as we do not support probing functions we can't extract data from. This was my initial point, while it could be useful, that is an extra feature on top of the stack tracking one. Or am I missing something?

yeah, I guess it was very late for me after a long day :)

retis/src/core/tracking/bpf/include/stack_tracking.h
vlrpl: I confirm there's a verification problem using u64 (addr vs pointer).
`u64 *` seems to be enough here hopefully it is enough for older kernels as well (CI will tell :)

vlrpl:
rebased on top of main

retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -337,7 +334,13 @@ static __always_inline int chain(struct retis_context *ctx)
 	if (ret)
 		log_warning("ctx extension failed: %d", ret);
 
-	filter(ctx);
+	skb = retis_get_sk_buff(ctx);
+	if (skb)
+		ctx->flags = filter(skb);
+	else if (stack_is_tracked(ctx->stack_base))
+		ctx->flags = RETIS_F_STACK_PASS;
```
atenart, line 340: Wouldn't this only be used when stack tracking is used for reporting events? We're not 100% sure how filtering because of stack tracking will work, maybe this should come when used?

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -63,18 +72,19 @@ static __always_inline struct tracking_info *skb_tracking_info(struct sk_buff *s
...
 
 	return ti;
 }
 
 static __always_inline int track_skb_start(struct retis_context *ctx)
 {
-	bool inv_head = false, no_tracking = false;
+	bool inv_head = false, no_tracking = false, free = false;
 	struct tracking_info *ti = NULL, new;
+	bool deferred_update = false;
```
atenart: nit: Should be in the above `bool` variable definitions.

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -190,6 +223,29 @@ static __always_inline int track_skb_end(struct retis_context *ctx)
...
+	 *
+	 * Potentially, even the case of delete stack entries through
+	 * deferred deallocations:
+	 *
+	 * - kp:non_skb_sym1 -> kp:skb_sym -- deferred --> kfree_skb
+	 *
+	 * is affected to a lesser extent by false positives (with the
+	 * overhead of synchronization).
+	 */
+	struct tracking_info *ti = bpf_map_lookup_elem(&tracking_map, &head);
```
atenart: nit: defines at start of function

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -190,6 +223,29 @@ static __always_inline int track_skb_end(struct retis_context *ctx)
...
 
+	/* Remove the stack tracking entry only if the free is not
+	 * deferred, otherwise this would be racy requiring some way
+	 * to synchronize the access. In general the following
+	 * assumption and restriction is applied: every probe chain
+	 * MUST start with an skb-aware function.
+	 * This approach may probably produce more false positive,
+	 * especially in the case:
+	 *
+	 * - kp:non_skb_sym1 -> kp:skb_sym
```
atenart: Isn't that forbidden now and skb availability enforced to generate a stack tracking id?

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -190,6 +223,29 @@ static __always_inline int track_skb_end(struct retis_context *ctx)
...
+	 * deferred deallocations:
+	 *
+	 * - kp:non_skb_sym1 -> kp:skb_sym -- deferred --> kfree_skb
+	 *
+	 * is affected to a lesser extent by false positives (with the
+	 * overhead of synchronization).
+	 */
+	struct tracking_info *ti = bpf_map_lookup_elem(&tracking_map, &head);
+
+	if (ti && ctx->stack_base == ti->stack_ref)
```
atenart, line 225: I might not remember correctly; is the intent of checking the strack ref matches to avoid removing a stack ref used somewhere else (and valid)?

retis/src/core/tracking/stack_tracking.rs
```diff
@@ -0,0 +1,36 @@
+//! # Stack tracking
+//!
+//! FIXME: add doc
```
atenart: :)

retis/src/core/probe/kernel/bpf/kprobe.bpf.c
```diff
@@ -34,21 +34,20 @@ int probe_kprobe(struct pt_regs *ctx)
 	context.timestamp = bpf_ktime_get_ns();
 	context.probe_type = KERNEL_PROBE_KPROBE;
 	context.orig_ctx = ctx;
+	context.stack_base = get_stack_base(ctx, context.probe_type);
```
atenart, line 37: This can in theory return 0, in which case we could have collisions and wrong initial contexts. Is that possible? If yes, should we fall back to the tgid?

retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -346,8 +349,17 @@ static __always_inline int chain(struct retis_context *ctx)
 	 * logic runs even if later ops fail: we don't want to miss information
 	 * because of non-fatal errors!
 	 */
-	if (RETIS_TRACKABLE(ctx->filters_ret))
+	if (RETIS_TRACKABLE(ctx->flags))
```
atenart: Not really related but we could just pass ctx here and hide the internals?

retis/src/core/probe/kernel/bpf/include/retis_context.h
```diff
@@ -77,11 +77,14 @@ struct retis_context {
...
 	void *orig_ctx;
-	/* Contains the bits identifying what filters yield a hit outcome.
-	 * A bit is set means that the filter matched the data based on its
-	 * criteria.
+	u64 stack_base;
+	/* Contains flags associated with the packet or the execution
+	 * in its context.
+	 * The first two bits are used for filtering, namely,
+	 * if a bit is set means that the filter matched the
+	 * data based on its criteria.
```
atenart: I would not document this, flags are flags and can be used for various things; including filtering results.

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -1,3 +1,4 @@
+#include "stack_tracking.h"
```
atenart: nit: shouldn't that be included after the common ones? Also s/"/</?

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -14,12 +15,17 @@ DEFINE_HOOK(F_AND, RETIS_ALL_FILTERS,
...
-	if (!skb)
-		return 0;
 
-	ti = skb_tracking_info(skb);
+	if (!skb) {
+		skb_ref = stack_get_skb_ref(ctx->stack_base);
+		ti = skb_tracking_info(&skb_ref);
+	} else {
+		ti = skb_tracking_info_by_skb(skb);
+	}
```
atenart: This would be useful when we'll support generating events based on stack tracking only. As such it should come along the feature.

retis/src/core/tracking/bpf/include/stack_tracking.h
```diff
@@ -0,0 +1,98 @@
...
+
+static __always_inline u64 stack_get_skb_ref(u64 stack_base)
+{
+	u64 *val;
+
+	val = bpf_map_lookup_elem(&stack_tracking_map, &stack_base);
+	if (!val)
+		return 0;
+
+	return *val;
```
atenart, line 90: Shouldn't `val` be an `u64` directly? The returned value should be the address of the skb/head already.

retis/src/core/probe/kernel/bpf/include/helpers.h
```diff
@@ -109,4 +113,11 @@ static __always_inline bool is_transport_data_valid(struct sk_buff *skb)
 	       !(is_network_valid(network) && network == transport);
 }
 
+static __always_inline bool kprobes_multi_has_cookies(void)
+{
+	return bpf_core_enum_value_exists(enum bpf_attach_type___x,
+					  BPF_TRACE_KPROBE_MULTI) &&
+	       bpf_core_field_exists(((struct bpf_kprobe_multi_link *)0)->cookies);
+}
+
```
atenart: really a nit but can you put this under `kprobe_get_func_ip` and leave all the skb head helpers grouped at the end?

retis/src/core/tracking/bpf/include/stack_tracking.h
```diff
@@ -0,0 +1,98 @@
...
+}
+
+static __always_inline u64 get_stack_base(void *ctx, enum kernel_probe_type type)
+{
+	u64 stack_addr = { get_stack_addr(ctx, type) };
+
+	return get_base_addr(&stack_addr);
+}
+
+static __always_inline u64 track_stack_start(u64 key, u64 value)
```
atenart: track_stack_update? I find "start" confusing, as the entry is always updated.

retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -88,9 +88,10 @@ enum {
 		/* Let the verifier be happy */					\
 		if (!ctx || !event)						\
 			return 0;						\
-		if (!((fmode == F_OR) ?						\
-		      (ctx->filters_ret & (fflags)) :				\
-		      ((ctx->filters_ret & (fflags)) == (fflags))))		\
+		if (!(ctx->flags & RETIS_F_STACK_PASS) &&			\
```
atenart: Shouldn't we drop this for now as we don't support non-skb events?

retis/src/core/tracking/bpf/include/skb_tracking.h
atenart: Does this still stands?

retis/src/core/probe/kernel/bpf/kprobe.bpf.c
vlrpl: I don't think it should be possible, unless the first page can be mapped in the kernel (not sure about NULL dereference, then).
Not sure about tgid, I guess they should not be considered reliable.

retis/src/core/tracking/bpf/include/skb_tracking.h
vlrpl: it is now and theoretically (excluding weird cases) even in the future. It has to be removed

retis/src/core/tracking/bpf/include/skb_tracking.h
vlrpl: garbage collection attempt when stack tracking mismatches with skb tracking. This assumes that stack entry is "hanging". The point is, how can I be in context B if I'm in context A?

retis/src/core/tracking/stack_tracking.rs
vlrpl: I trusted clippy here :)

retis/src/core/probe/kernel/bpf/include/retis_context.h
vlrpl: sounds good

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
vlrpl: yes, not sure why it's like this, especially the double quote part. I suspect lsp-mode inserted it automatically, but it's easy to blame tooling :)

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
vlrpl: This would also require the removal of the helpers, wouldn't it?

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: I guess this could potentially be helpful for the ovs case, but I'll check this

retis/src/core/tracking/bpf/include/skb_tracking.h
vlrpl: it slipped

retis/src/core/tracking/bpf/include/stack_tracking.h
vlrpl: the idea was to keep naming consistent with `track_skb_start()`. I guess that also updates.
If update makes things clearer, I'm ok to rename it, though.

retis/src/core/probe/kernel/bpf/include/helpers.h
vlrpl: sure

retis/src/core/tracking/bpf/include/skb_tracking.h
atenart: Since if `free` is true we're not updating the context, this can happen for various reasons. Are we not freeing the entry in case it's used by another `tracking_info`? Otherwise why keeping the entry?

A similar question arises in `track_skb_start`:

```
if (ti->stack_ref != ctx->stack_base)                           
       ti->stack_ref = ctx->stack_base;
```

Why not removing the stack tracking entry too?

retis/src/core/probe/kernel/bpf/kprobe.bpf.c
atenart: If 0 is considered being a bug, then there's no real need to fallback on another way of tracking things. We're also already reporting an error. In such case maybe we should just bail early from the users as it's a bug?

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
atenart: As you prefer, IMO the helper can stay as it's part for the stack tracking API. Or can be added later.

retis/src/core/tracking/bpf/include/stack_tracking.h
atenart: I would personally prefer as I'm always wondering how the temp `skb` tracking works when reading the code... until I remember this is always updating the entry :)

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: the PR remained paused for a while and I may misrecall, but this could be useful to simplify the ctx logic in ovs

retis/src/core/probe/kernel/bpf/kprobe.bpf.c
vlrpl: The log is already a bit defensive as 0 should really be an internal problem. I guess bailing is not a big deal, although not a great difference (we're in a theoretical corrupted state anyways).

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
vlrpl: my preference is to keep all the parts that are obviously non-invasive

retis/src/core/tracking/bpf/include/skb_tracking.h
vlrpl: Got split as that exceeds 80 chars and looked weird with a newline. I'm fine with it, though.

retis/src/core/tracking/bpf/include/skb_tracking.h
vlrpl: just to clarify, I was speaking about start, not end. End just terminates the current stack/context.
"end" can safely delete the context as it is the owner (no multiple context are supposed to be possible per skb tracked).
We don't remove it from start as being different (and not being the owner), we can delete someone else's. Also, this is an LRU map, so we rely on self clean-up capabilities, if we need entries.
The idea is, "control what you own".

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
vlrpl: the question is, why would we fix this (and not remove it) if we have to remove the other hunk?

atenart: Of course, if it's not needed anymore it can be removed.

retis/src/core/tracking/bpf/include/skb_tracking.h
atenart: I see. As you prefer then (if not splitting you can keep it on a single line). Or maybe checkpatch is too strict :)

retis/src/core/tracking/bpf/include/skb_tracking.h
atenart: All right, makes sense. Maybe a small comment like "We're not removing entries that might be used elsewhere" could be added. As you want.

retis/src/core/probe/kernel/bpf/kprobe.bpf.c
atenart: If this happens the error will be seen during collection, but the potentially wrong event will stay at post-processing time. So maybe bailing out would be better.

Although if it's really unexpected and a major issue it might be pointless to handle it.

vlrpl: still unsure about bailing as it also will prevent to go through chain() in all the (and there are cases in which things would work anyways regardless the stack base correctness). Still bailing doesn't make much difference.

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: it should not hurt in general, it all depends on how much we want to remove and readd afterwards.

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: same as above. We cannot actually probe non skb functions dynamically. Ovs probes some non skb functions (see e.g. `ovs_flow_tbl_lookup_stats`). Again, it all boils down on how much real interest there is in following this up soon (as it should be).

retis/src/core/tracking/bpf/include/stack_tracking.h
vlrpl: I don't follow, `bpf_map_lookup_elem()` returns the pointer of the value or NULL.

atenart: Right, I got confused, sorry.

retis/src/core/probe/kernel/bpf/kprobe.bpf.c
atenart: Let's keep it that way then.

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -16,10 +16,8 @@ DEFINE_HOOK(F_AND, RETIS_ALL_FILTERS,
 	struct sk_buff *skb;
 
 	skb = retis_get_sk_buff(ctx);
-	if (!skb)
-		return 0;
```
atenart, line 20: Shouldn't this be left?

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -161,13 +186,14 @@ static __always_inline int track_skb_end(struct retis_context *ctx)
...
 	u64 head, ksym = ctx->ksym;
+	struct tracking_info *ti;
 	struct sk_buff *skb;
 
 	cfg = bpf_map_lookup_elem(&tracking_config_map, &ksym);
 	if (!cfg)
 		return 0;
 
-	/* We only supports free functions below */
+	/* We only supports   functions below */
```
atenart: Looks like the "free" removal was a bit too broad :)

(There's at least another one)

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
vlrpl: yes, I noticed both :(

retis/src/core/probe/kernel/bpf/include/common.h
atenart: I agree setting the flag is nice as it convey information about stack tracking. However letting non-raw hooks be called if not being skb tracked seems out of scope here. The above chunk is nice but I would remove the `DEFINE_NAMED_HOOK` chunk. E.g. allowing hooks to be called based on stack tracking alone will probably require another flag to be on, e.g. `ctx->flags & RETIS_F_STACK_PASS && stack_track_on` (to allow only reporting events between known points).

While looking at the diff I'm also wondering if returning early in (the chunk is later reworked but the `return` stays):

```
+       else if (!stack_is_tracked(ctx->stack_base))
+               return 0;
```

won't prevent raw hooks from being called if they're hit before an skb-aware path? Which would break the raw hook expectations.

WDYT?

retis/src/core/probe/kernel/bpf/include/common.h
atenart: [being discussed in the other thread]

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: > I agree setting the flag is nice as it convey information about stack tracking. However letting non-raw hooks be called if not being skb tracked seems out of scope here. The above chunk is nice but I would remove the `DEFINE_NAMED_HOOK` chunk. E.g. allowing hooks to be called based on stack tracking alone will probably require another flag to be on, e.g. `ctx->flags & RETIS_F_STACK_PASS && stack_track_on` (to allow only reporting events between known points).
> 
> While looking at the diff I'm also wondering if returning early in (the chunk is later reworked but the `return` stays):
> 
> ```
> +       else if (!stack_is_tracked(ctx->stack_base))
> +               return 0;
> ```
> 
> won't prevent raw hooks from being called if they're hit before an skb-aware path? Which would break the raw hook expectations.
> 
> WDYT?

This should be true both the conditions are true, but for raw hooks the filtering should always be pass, shouldn't it?

retis/src/core/tracking/bpf/include/skb_tracking.h
vlrpl: I've been testing evil-mode lately (well, more an hybrid version of it).
I was not expecting evil changes (like this) but just vim shortcuts :)

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: We are talking about the ovs reinjection path.
As long as there's an skb (any), the raw hook is a pass (whether there is a flag or not) and the hook processes the data.
We have skbs in those functions, so this should be no problem (and should fail badly in the CI).
In the path without skb (e.g. table lookup) we should always be stack tracked (as everything stays in the kernel.
That being said, in terms of pure assumptions, this, without a queue_id resolution (and map to stack tracking) might be a bit brittle.
Mapping stack tracking on queue_id resolution and trying to simplify the ovs logic (where possible) was one of the next steps.
It's probably safer to remove this and the check in the hook.
It's a pity, though, how things got a bit dismantled, but that's ok, follow ups will come.

vlrpl: just to clarify, it's a pity because of the many non-trivial assumptions part of the design (e.g. this was a self clean up part of it). Non-trivial because they potentially impacted more components :)

atenart: Also raw hooks were supposed to be called regardless of filtering and tracking, this is more in line. Thanks for the modifications!

amorenoz (approved):
LGTM, thanks

PR #535, opened 2025-06-24
Vagrantfile: resize c8s disk and filesystem

vlrpl:
Only c8s is affected by the issue, so the changes have only been applied locally.


retis/src/core/filters/meta/filter.rs
```diff
@@ -1462,6 +1462,11 @@ impl FilterMeta {
 
         Ok(mf)
     }
+
+    #[cfg(feature = "debug")]
```
atenart, line 1466: Shouldn't we use `debug_assertions` here instead?

retis/src/core/filters/meta/filter.rs
```diff
@@ -750,7 +750,9 @@ impl<'a> MetaExpr<'a> {
             ));
 
             idx_sz += 1;
-            if idx_sz >= ctx.target.sz {
+            // idx_sz >= ctx.target.sz is just enough, but this is
+            // needed to make old verifier happy.
+            if idx_sz >= std::mem::size_of_val(&ctx.target.md) {
                 break;
```
atenart, line 756: Shouldn't `sz` also checked as my understanding is `sz` can be < `size_of_val(md)`?

retis/src/core/filters/meta/filter.rs
vlrpl: Why is that?
This follows what we have already for the rest of disasm() usage.
If the suggestion is that we can switch to that, I suppose that should be a different change, isn't it?

retis/src/core/filters/meta/filter.rs
vlrpl: Here we store zeros (already in md) up to 32 bytes, regardless of sz.
Do you have a problematic filter example because of this?

atenart: No, I was just reading the snippet w/o having the full context.

retis/src/core/filters/meta/filter.rs
atenart: Forgot about that, makes sense then.

atenart (approved):
LGTM, thanks!

PR #541, opened 2025-07-11
Filtering add tests

### LINKED ISSUE #58: Implement tests for filtering

vlrpl:
Fixes #58, although it would be nice to expand them.

This does not replace the need for functional tests, as actual kernel verification is required, but makes it extremely easy to introduce even complex tests (particularly convenient for regressions) without creating cumbersome setups nor requiring caps


vlrpl:
Given we don't run the test target while creating packages, this can be set back to the v1.6 milestone. OTOH, being a test, there's no actual rush (but it would just be nice to have) to include it before v1.6

retis/src/core/filters/meta/filter.rs
```diff
@@ -1619,4 +1631,159 @@ mod tests {
         let _ = FilterMeta::from_string(bool_expr.to_string())?;
         Ok(())
     }
+
+    // Dummy function with no real probe capabilities.
+    // Use it providing dst and src accordingly.
+    fn bpf_probe_read_kernel_helper(dst: u64, size: u64, src: u64, _arg4: u64, _arg5: u64) -> u64 {
```
atenart: Why returning `u64`; those return negative values on error and are `long` upstream.

retis/src/core/filters/meta/filter.rs
```diff
@@ -1619,4 +1631,159 @@ mod tests {
...
+                *mdst.add(i) = byte;
+                if byte == 0 {
+                    return (i + 1) as u64;
+                }
+            }
+        }
+        size
+    }
+
+    use skb_gen::{net_device, nf_conn, sk_buff};
```
atenart: nit: please keep includes grouped at the top.

retis/src/core/filters/meta/filter.rs
```diff
@@ -1619,4 +1631,159 @@ mod tests {
...
+        // +----------------+-------------+------------+-----------+
+        // | len            | 112         | 0          | 4 bytes   |
+        // | queue_mapping  | 124         | 0          | 2 bytes   |
+        // | cloned         | 126         | 0          | 1 bit     |
+        // | pkt_type       | 128         | 0          | 3 bits    |
+        // | nf_trace       | 128         | 4          | 1 bit     |
+        // | ip_summed      | 128         | 5          | 2 bits    |
+        // | vlan_tci       | 154         | 0          | 2 bytes   |
+        // +----------------+-------------+------------+-----------+
+        // net_device.name is a fixed-size (16) array within struct net_device.
```
atenart: Let's put this as the function comment.

retis/src/core/filters/meta/filter.rs
```diff
@@ -1619,4 +1631,159 @@ mod tests {
...
+    #[test_case("sk_buff.pkt_type == 0b110 and sk_buff.nf_trace == 0b01" => true; "two bitfields with true and true on the same unit")]
+    #[test_case("sk_buff.vlan_tci == 1 or sk_buff.dev.name == 'verylongtruncatedname'" => false; "negative two fields false or false")]
+    #[test_case("sk_buff._nfct:0x7 == 0x2 and sk_buff._nfct:~0x7:nf_conn.mark > 2" => true; "two fields with cast and mask+cast (true and true)")]
+    #[test_case("sk_buff._nfct:0x7 == 0x2 and sk_buff._nfct:~0x7:nf_conn.mark != 3" => false; "negative two fields with cast and mask+cast (true and true)")]
+    #[test_case("sk_buff.vlan_tci == 1 and sk_buff.dev.name == 'foo' or sk_buff.dev.name == 'verylongtruncat'" => true; "three field default precedence (false and false) or true")]
+    #[test_case("sk_buff.vlan_tci == 1 and (sk_buff.dev.name == 'foo' or sk_buff.dev.name == 'verylongtruncat')" => false; "negative three field false and (false or true)")]
+    fn meta_filter_runtime(expr: &'static str) -> bool {
+        use std::{mem, slice};
+
+        use rbpf;
```
atenart: nit: please put this at the top of the mod.

retis/src/core/filters/meta/filter.rs
```diff
@@ -1619,4 +1631,159 @@ mod tests {
...
+
+        let (skb, _net_dev, _nfct) = init_sk_buff();
+
+        let mf = FilterMeta::from_string(format!("{expr}").to_string());
+        assert!(mf.is_ok());
+        let mf = mf.unwrap();
```
atenart: No need to `assert` then `unwrap`, unwrapping alone will make the test to fail.

retis/src/core/filters/meta/filter.rs
```diff
@@ -1619,4 +1631,159 @@ mod tests {
...
+
+        let mem = vec![];
```
atenart: nit: `Vec::new()` ?

retis/src/core/filters/meta/filter.rs
```diff
@@ -1619,4 +1631,159 @@ mod tests {
...
+
+        let mbuff = unsafe {
+            slice::from_raw_parts((&skb as *const _) as *const u8, mem::size_of::<sk_buff>())
+        };
```
atenart, line 1733: We can reuse (and export) `fn as_u8_vec` here ?

retis/src/core/filters/meta/filter.rs
```diff
@@ -1619,4 +1631,159 @@ mod tests {
...
+        let prog = &mf.to_bytes();
+
+        let mut vm = rbpf::EbpfVmMbuff::new(Some(prog)).unwrap();
+        vm.register_helper(113, bpf_probe_read_kernel_helper)
+            .unwrap();
+        vm.register_helper(115, bpf_probe_read_kernel_str_helper)
+            .unwrap();
+        vm.execute_program(&mem, &mbuff).unwrap() != 0
```
atenart, line 1741: Maybe we can `assert` instead of returning a `bool` here?

retis/test_data/skb_gen.rs
```diff
@@ -0,0 +1,7412 @@
+/* Generated out of:
+ * pahole --compile -C sk_buff,sock,net_device,nf_conn test_data/vmlinux > skb_gen.h && \
+ * bindgen --with-derive-default skb_gen.h > skb_gen.rs
+ * automatically generated by rust-bindgen 0.70.1
```
atenart: nit: empty space at EOL.

retis/src/core/filters/packets/filter.rs
```diff
@@ -69,3 +69,102 @@ impl FilterPacket {
         Ok(self.0.to_bytes())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use crate::bindings::packet_filter_uapi::*;
+    use test_case::test_matrix;
```
atenart: nit: we usually do std then external crates then crate.

retis/src/core/filters/packets/filter.rs
```diff
@@ -69,3 +69,102 @@ impl FilterPacket {
...
+
+    // Dummy function with no real probe capabilities.
+    // Use it providing dst and src accordingly.
+    fn bpf_probe_read_kernel_helper(dst: u64, size: u64, src: u64, _arg4: u64, _arg5: u64) -> u64 {
```
atenart: Can we reuse the one already defined? (Or add a mod grouping those?)

retis/src/core/filters/packets/filter.rs
```diff
@@ -69,3 +69,102 @@ impl FilterPacket {
...
+
+        use rbpf;
+
+        let packet_bytes: [u8; 54] = [
+            0x00, 0x22, 0x22, 0x22, 0x22, 0x22, 0x00, 0x11, 0x11, 0x11, 0x11, 0x11, 0x08, 0x00,
+            0x45, 0x00, 0x00, 0x28, 0x00, 0x01, 0x00, 0x00, 0x40, 0x06, 0x65, 0xd1, 0x0a, 0x00,
+            0x00, 0x01, 0x0a, 0x00, 0x00, 0xfe, 0xe9, 0x85, 0x00, 0x50, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x50, 0x12, 0x20, 0x00, 0x90, 0xfe, 0x00, 0x00,
+        ];
+        // Same packet. The L3 flavor does not include the MAC header.
```
atenart: Same packet? Also shouldn't the comment be before the packet definition?

vlrpl: things changed, but the comment wasn't really updated

retis/src/core/filters/meta/filter.rs
vlrpl: This would lead to add more boilerplate.
We perform negative and positive tests using `=> ...` [output operator](https://github.com/frondeus/test-case/wiki/Syntax#output-matcher) that eventually performs the assertion.
Do you have any particular preference over this approach?

retis/src/core/filters/packets/filter.rs
vlrpl: yep, I know  :)

retis/src/core/filters/meta/filter.rs
vlrpl: It's because the vm expects that signature (see e.g. the built-in [bpf_trace_printk](https://docs.rs/rbpf/latest/rbpf/helpers/fn.bpf_trace_printf.html)).

retis/src/core/filters/meta/filter.rs
vlrpl: meaning moving it under Retis?
I considered something like this, but this resulted in creating a new helper with just that function to replace a one liner with a function call (of course, this hides some syntax sugar repetition), and so I left it up to when at least a couple of consumers requires this in Retis.
If you prefer to do it now rather than once we have more than one user, we can of course proceed.
WDYT?

retis/src/core/filters/meta/filter.rs
atenart: Ahh I missed that part. Thanks, that makes sense!

retis/src/core/filters/meta/filter.rs
atenart: As you prefer. This helper was introduced for very similar use cases and would make sense here, although it's true this is only a one liner.

retis/src/core/filters/meta/filter.rs
atenart: There is the same construction in the other filtering flavor.

retis/src/core/filters/meta/filter.rs
atenart: There is the same construction in the other filtering flavor.

retis/test_data/skb_gen.rs
atenart: This still stands in the latest version pushed (just in case you missed it).

retis/src/core/filters/meta/filter.rs
vlrpl: right, missed this

atenart (approved):
LGTM, thanks!

PR #543, opened 2025-07-15
Events: refactor the packet, dev and ns collector/sections

atenart:
The `packet` `dev` and `ns` event sections are not tied to `struct sk_buff` nor to the `skb` collector. Move them to be top level sections, and move the `dev` and `ns` data retrieval to their own collectors. This helps retrieving such information if part of a probe argument and will help future additions (e.g. core stat retrieval with the `dev` collector).

This will break the API as the following data is moving from one place to another in the event:
- `skb/packet` to `packet`
-  `skb/dev` to `dev`
- `skb/ns` to `netns`

In addition split collector and event documentation pages, to make things more clear. This is not 100% linked to this PR (but is required for some of the changes) and could have been done as part of unlinking the collectors and event sections.

The documentation generated using this PR can be see at https://retis.readthedocs.io/en/at-event-sections-move/.

Also adding a small commit fixing the VLAN accel data retrieval, as this also would break compatibility if we delay it to 1.7. It's currently not as the feature was not released yet. And a few Python changes as discussed offline.

### DIFF
--- a/retis/src/core/tracking/bpf/include/skb_tracking.h
+++ b/retis/src/core/tracking/bpf/include/skb_tracking.h
@@ -140,9 +140,6 @@ static __always_inline int track_skb_start(struct retis_context *ctx)
 		ti->last_seen = ctx->timestamp;
 		ti->orig_head = head;
 
-		/* No need to globally track it if the first time we see this
-		 * skb is when it is freed.
-		 */
 		bpf_map_update_elem(&tracking_map, &head, &new, BPF_NOEXIST);
 	}
 


docs/events/intro.md
```diff
@@ -0,0 +1,18 @@
...
+An event is always composed of a common section (containing a timestamp) and a
+set of optional other sections. Those sections data can come from the eBPF
+probes or from Retis directly. Sections are grouped in an event if they share a
+common property (eg. they are all linked to a given packet + probe).
+
+```none
+<common> <kernel or userspace> <tracking> <drop> <stack trace> <...>
+```
+
+Each event section has its own dedicated documentation page.
```
amorenoz, line 18: Nit, a list of links to those sections would make navigation a bit easier.

docs/collectors/overview.md
```diff
@@ -33,8 +33,8 @@ The `skb` collector has a single specific argument, `--skb-sections`. This is
 used to choose which parts of the `skb` metadata to retrieve and
 export in the events. See the `retis collect --help` for a detailed description.
 
-The `skb` collector produces the [skb](../events/skb.md) and
-[packet](../events/packet.md) event sections.
+The `skb` collector produces the [skb](../events/skb.md),
+[packet](../events/packet.md) and [dev](../events/dev.md) event sections.
```
amorenoz: This is confusing. `skb` collector does not longer produce the `dev` section.

retis/src/collect/collector/skb/skb.rs
```diff
@@ -83,12 +82,18 @@ impl Collector for SkbCollector {
...
-                "packet" | "arp" | "ip" | "tcp" | "udp" | "icmp" | "dev" | "ns" => {
+                "packet" | "arp" | "ip" | "tcp" | "udp" | "icmp" | "ns" => {
                     warn!(
                         "Use of '{}' in --skb-sections is depreacted",
                         category.as_str(),
                     );
                 }
+                "dev" => {
+                    warn!(
+                        "Use of '{}' in --skb-sections is depreacted: is now a collector",
```
amorenoz: s:/is now a collector/it is now a collector/

retis/src/collect/cli.rs
```diff
@@ -24,7 +24,7 @@ pub(crate) struct Collect {
         short,
         long,
         value_parser=PossibleValuesParser::new([
-            "skb-tracking", "skb", "skb-drop", "ovs", "nft", "ct",
+            "skb-tracking", "skb", "skb-drop", "ovs", "nft", "ct", "dev",
```
amorenoz: Maybe enable it in `dropmon` et.al profiles?

amorenoz: Also maybe in `pcap` one?

docs/collectors/overview.md
```diff
@@ -34,7 +34,8 @@ used to choose which parts of the `skb` metadata to retrieve and
 export in the events. See the `retis collect --help` for a detailed description.
 
 The `skb` collector produces the [skb](../events/skb.md),
-[packet](../events/packet.md) and [dev](../events/dev.md) event sections.
+[packet](../events/packet.md), [dev](../events/dev.md) and
+[netns](../events/netns.md) event sections.
```
amorenoz: Confusing: netns has it's own collector now.

docs/python.md
```diff
@@ -216,6 +216,8 @@ With the top-level event and all sections:
 8974965787422 (5) [ping] 100854
 >>> print(event.ct)
 ct_state NEW status 0x188 icmp orig [10.0.42.5 > 1.1.1.1 type 8 code 0 id 1] reply [1.1.1.1 > 10.0.42.5 type 0 code 0 id 1] zone 0 mark 0
+>>> print(repr(e.common))
```
amorenoz: Just typing `e.common` in any python REPL will call `__repr__`

retis-events/src/python.rs
```diff
@@ -104,8 +87,8 @@ impl PyEvent {
     /// Returns internal data as a dictionary
     ///
     /// Returns a dictionary with all key<>data stored (recursively) in the
-    /// event, eg. `e.raw()['skb']['dev']`.
-    fn raw(&self, py: Python<'_>) -> PyResult<PyObject> {
+    /// event, eg. `e.to_raw()['skb']['dev']`.
+    fn to_raw(&self, py: Python<'_>) -> PyResult<PyObject> {
```
amorenoz: About the rationale for this change:
We could change the field in the event, right? Maybe `packet.data` makes more sense?
That way we could keep the old implementation and print a warning.

OTOH, I'm not 100% convinced of "to_raw()" but I don't have an obviously better alternative.
"raw" is not very expressive (from a python users pov, what is the opposite? what is "cooked"?). maybe "to_dict" is a bit clearer, or "to_json" (although it's not json but it's how the json event looks like). IDK, naming is hard.

docs/events/intro.md
atenart: I did not add an explicit list in there as it would be yet another place to maintain whenever we add a new section. Also the list is on the navigation panel, quite visible. Not strong opinion though, I can add it if you prefer.

docs/collectors/overview.md
atenart: Oops.

docs/collectors/overview.md
atenart: Oops

retis/src/collect/cli.rs
atenart: Makes sense. I'll add the `ns` collector too.

docs/python.md
atenart: Right but it's less obvious than explicitly calling `__repr__`. On one hand this makes it obvious for library users, on the other hand it hides the easier use in a REPL. Thoughts?

retis-events/src/python.rs
atenart: > About the rationale for this change: We could change the field in the event, right? Maybe `packet.data` makes more sense? That way we could keep the old implementation and print a warning.

Yes, that is possible too. We have the opportunity to fix things now though and avoid carrying a not well named method. If `packet.data` is better than `packet.raw` though that can be done regardless of keeping `raw()`.

> OTOH, I'm not 100% convinced of "to_raw()" but I don't have an obviously better alternative. "raw" is not very expressive (from a python users pov, what is the opposite? what is "cooked"?). maybe "to_dict" is a bit clearer, or "to_json" (although it's not json but it's how the json event looks like). IDK, naming is hard.

I'm not really experienced in Python, `to_dict()` sounds good though as it's really what's happening and what the result is.

docs/events/intro.md
amorenoz: That's true, no strong opinion either.

retis-events/src/python.rs
amorenoz: > > About the rationale for this change: We could change the field in the event, right? Maybe `packet.data` makes more sense? That way we could keep the old implementation and print a warning.
> 
> Yes, that is possible too. We have the opportunity to fix things now though and avoid carrying a not well named method. If `packet.data` is better than `packet.raw` though that can be done regardless of keeping `raw()`.
> 

Yes, it's completely unrelated. The `raw` word is a bit of an "internal leak": it represents how the thing is seen from the inside of retis. We, retis, did not modify it, therefore, it's raw. `data` or `bytes` actually represents what it is from the user's PoV.

Besides, from an event perspective, we used to have the packet header information `processed` (as opposed to `raw`). We don't have any longer, so `raw` feels a bit un-contextualized. 

> > OTOH, I'm not 100% convinced of "to_raw()" but I don't have an obviously better alternative. "raw" is not very expressive (from a python users pov, what is the opposite? what is "cooked"?). maybe "to_dict" is a bit clearer, or "to_json" (although it's not json but it's how the json event looks like). IDK, naming is hard.
> 
> I'm not really experienced in Python, `to_dict()` sounds good though as it's really what's happening and what the result is.

Maybe @vlrpl can weight in as this is quite subjective. Again, from the user perspective I think it's clearer to describe "what happens" (i.e: a verb) or "what the result is". In this case a `raw` event is the opposite of a `processed` one, but the content is quite similar.

OTOH, `to_XXX` suggests some kind of processing, I read it as `transform_it_to_XXX`. If it's "transformed" it's not "raw". From that point of view, `get_raw()` would be slightly clearer than `to_raw()`. However, it would suggest we keep an internal "unprocessed" representation of the event that we're just retrieving, so it's a cheap thing. But it's actually a relatively expensive operation as we're marshaling the event and transforming it into a dictionary.

In conclusion, for me the best name would be `to_dict()` but I'm open to discussion.

atenart: Makes sense. I'm fine with renaming `packet.raw` to `packet.data` ("data" is used in the skb to point to the same content, PCAP use that keywork too, etc) and to rename `raw()` to `to_dict()`. @vlrpl does that sound good to you as well?

docs/events/intro.md
atenart: Let's keep it that way then.

docs/python.md
amorenoz: Fair enough, if it's just for example purposes ... I just never seen it used explicitly :-)

retis-events/src/python.rs
vlrpl: I mean, the following argument:
> "raw" is not very expressive (from a python users pov, what is the opposite? what is "cooked"?)

almost made me lean toward `to_raw()` but this time I have to just appreciate the food reference (well done :) and express my preference toward `to_dict()` as I find it a better fit for this case.

docs/python.md
atenart: I added a comment to make things more explicit.

retis-events/src/events.rs
```diff
@@ -125,10 +125,22 @@ impl EventFmt for Event {
...
+        // packed. Also the two are quite related and it makes sense to display
+        // them on a single line.
+        let mut is_netns = false;
+        if let Some(netns) = &self.netns {
+            is_netns = true;
+            write!(f, "{sep}")?;
+            netns.event_fmt(f, format)?;
+        }
+        if let Some(dev) = &self.dev {
+            write!(f, "{}", if is_netns { ' ' } else { sep })?;
```
vlrpl: optional nit: `is_netns` can be avoided  `self.netns.is_some()`

vlrpl (commented):
The change LGTM.
I have just a minor optional comment, but AFAIC the change is ready to be merged.

retis-events/src/events.rs
atenart: Indeed, that is better!

PR #545, opened 2025-07-23
Add auto-stale checks

vlrpl:
The idea is to mark a PR or an issue as stale and close it after a given number of days.

- PRs/issues in a milestone do not go stale
- PRs/issues with 'NoStale'
- Adding a comment to a stale PR/issue refreshes the state and removes the label

The current values:
AUTO_STALE_THRESHOLD: 200
AUTO_CLOSE_THRESHOLD: 60

can be reduced as they are admittedly a bit loose.

### DIFF
--- a/.github/workflows/daily_change_check.yaml
+++ b/.github/workflows/daily_change_check.yaml
@@ -21,3 +21,44 @@ jobs:
           gh workflow run build_push_image.yaml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  stale-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          # Do not consider issues/PRs with a milestone set.
+          exempt-all-milestones: true
+          # When 'NoStale' is set, PRs and issues are exempted from going stale.
+          # 'Stale' is the default label of PRs reaching the stale state.
+          exempt-pr-labels: 'NoStale'
+          exempt-issue-labels: 'NoStale'
+          # Override the default stale/close (60d/7d) inactivity timeouts
+          days-before-stale: ${{ env.AUTO_STALE_THRESHOLD }}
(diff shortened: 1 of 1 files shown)


atenart:
This already looks good, thanks! A few small comments below.

- I would not keep issues from being marked as being stalled with a 'NoStall' flag. This will help having to review them once a year (we'll get a notification) and make sure they're still aligned. Manually removing the 'stall' flag once a year for issues is probably OK.
- However we could have a 'WIP' flag, preventing an issue from being auto-closed. Some issues require months to get done and this way we would ensure it's known to be worked on. An even better way would be to not close issues having a assignee but from a quick look at the action that doesn't seem possible.
- I would increase the stall delay to be a year exactly. The closing delays of 2 months looks fine.

vlrpl:
> This already looks good, thanks! A few small comments below.
> 
>     * I would not keep issues from being marked as being stalled with a 'NoStall' flag. This will help having to review them once a year (we'll get a notification) and make sure they're still aligned. Manually removing the 'stall' flag once a year for issues is probably OK.
> 
>     * However we could have a 'WIP' flag, preventing an issue from being auto-closed. Some issues require months to get done and this way we would ensure it's known to be worked on. An even better way would be to not close issues having a assignee but from a quick look at the action that doesn't seem possible.
> 

I guess we can turn 'NoStale' label to 'WIP'. I picked NoStale just because it is generic (unlike WIP, which is fairly specific in its name).

>     * I would increase the stall delay to be a year exactly. The closing delays of 2 months looks fine.

A whole year feels ok for issues, but sounds a little too much for PRs (but this may just be my impression).

A PR with no activity (not even a review request from the author) for e.g., 4 months sounds already abandoned.
I wonder if, although it might have happened in the past (not 100% sure here, as most of the "slow ones" got updated/rebased), this is common enough to justify a whole year of inactivity timeout.
Also, if marked as Stale (after the hypothetical 4 months), there's always the option to remove and reset the counter with a comment (or a proper review, of course) or set it as WIP.
WDYT?

vlrpl:
Here's a link to a test (days set to 0 for both days before stale and before close)
https://github.com/vlrpl/retis/pull/7

PR #550, opened 2025-07-29
Backward compatibility of JSON events

atenart:
This introduces a layer for handling and expressing backward compatibility handling as well as an implementation for JSON events (which are read from the events file). This provides backward compatibility when reading an older file with a newer version of Retis (`sort`, `print`, `python`, `pcap` sub-commands), starting with 1.5.0 (we were not embedding the version before).

Efforts were done to come up with an implementation for Python events, but it happened to be more complex than initially thought. It is still a goal mid-term, although the way forward is not clear.

If we want JSON compatibility regardless, having a common layer or not, or having some details to fixup should be fine as the overall logic (fixing up an intermediate `serde_json` representation) proved to be working and is quite flexible.


retis-events/src/compat/compat.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+impl CompatVersion {
+    pub const UNKNOWN: Self = Self::V0;
+    pub const LATEST: Self = Self::V1;
+
+    /// Create a compatibility version representation given a Retis version.
+    pub(crate) fn from_retis_version(version: &str) -> Self {
+        let compat_version = match version {
+            x if x.starts_with("v1.6.") || (x.starts_with("v1.5.") && x.contains("-")) => {
```
amorenoz: Is the goal to match on unreleased versions between 1.5 and 1.6?

amorenoz: If so, how do we really know what `CompatVersion` they represent?

retis-events/src/compat/compat.rs
```diff
@@ -0,0 +1,138 @@
...
+                CompatVersion::V1
+            }
+            x if x.starts_with("v1.5.") => CompatVersion::V0,
+            // The retis version was not provided prior to v1.5.0. Catch all non
+            // explicit versions as newer ones.
+            _ => CompatVersion::LATEST,
```
amorenoz: Don't we have higher chances if we choose the oldest?

retis-events/src/compat/compat.rs
atenart: > Is the goal to match on unreleased versions between 1.5 and 1.6?

Yes.

> If so, how do we really know what `CompatVersion` they represent?

There will always be a gap (between the latest release and the commit introducing changes) but there's no easy way to test for such thing. As the development version in unreleased this should be fine. It's needed otherwise using the development branch won't work.

retis-events/src/compat/compat.rs
atenart: The goal is to match any newer version if not explicitly stated[1], without having to implement a version matcher (to keep things simple). After v1.5.0 all versions should be matched the correct way; before 1.5.0 we don't support backward compat' anyway.

To be clear; we can support 1.0..1.4 but at the cost of having a more complex logic.

[1] This makes it easier to maintain and be sure the latest version always takes the shortcut.

amorenoz: Right, currently we do not support <1.5.
But we should not keep growing this `match` endlessly. At some point we want to cut it and say, we support the last N releases, at least from a maintanance PoV I think it's sane.

So, ideally we should have something like:
```
< FIRST_SUPPORTED => V0 (best effort) or even just  error
= FIRST_SUPPORTED_Y_STREAM = V0
= SECOND_SUPPORTED_Y_STREAM = V1
...
>= CURRENT_RELEASE = VN
```

retis-events/src/compat/compat.rs
amorenoz: This code is assuming 1.6 is released, which makes sense as we're preparing the release.

The development will always be ahead of the latest release, not behind. Meaning, once we release 1.6, development branches will be "1.6.x-" which will match the "catchall" case below. 

If at any point we are working on "1.5.x-1234" it is more likely we are working on a z-stream release of 1.5, e.g: 1.5.2, maybe to fix some bug or something, rather than working on something close to 1.6 given 1.6 has been released. Therefore, I think it'll make more sense to keep it compatible with 1.5 rather than 1.6.

So the only gap here is the development that happens between this patch and the actual 1.6 release which should be small. For that, what  we can do is "always bump the retis version on the same PR that adds the compat version". This PR is special as it's introducing the framework but moving forward this would make sense.

So the idea would be to have
V0:  1.5.*
V1:  1.6.0
V2: > 1.6.0

V2 contains zero fixups.

During the development of 1.7 we add add fixups to V2 as we break stuff
When we release 1.7 we change the above to be:
V0: 1.5.*
V1: 1.6.*
V2: 1.7.0
V3: >1.7.0

If we release 1.6.1 (zstream) before releasing 1.7, instead we:
V0:  1.5.*
V1:  1.6.1
V2: > 1.6.1

WDYT?

> It's needed otherwise using the development branch won't work.
Not sure I  understand this.

retis-events/src/compat/compat.rs
```diff
@@ -0,0 +1,138 @@
+#![allow(dead_code)] // Not everything is used in all versions.
```
amorenoz, line 1: On commit message: s/fundations/foundations/?

retis-events/src/compat/json.rs
```diff
@@ -0,0 +1,123 @@
...
+        match val {
+            serde_json::Value::Object(map) => {
+                val = match map.get_mut(field) {
+                    Some(val) => val,
+                    None => return Ok(None),
+                }
+            }
+            _ => {
+                if i != leaf {
+                    bail!("Field '{field}' is not a leaf");
```
amorenoz: nit: The term `leaf` might be a bit misleading here. IIUC, it's not mandatory that `target` points to a leaf of the JSON "tree", i.e: a non-map element.

retis-events/src/compat/compat.rs
atenart: > The development will always be ahead of the latest release, not behind. Meaning, once we release 1.6, development branches will be "1.6.x-" which will match the "catchall" case below.
> 
> If at any point we are working on "1.5.x-1234" it is more likely we are working on a z-stream release of 1.5, e.g: 1.5.2, maybe to fix some bug or something, rather than working on something close to 1.6 given 1.6 has been released. Therefore, I think it'll make more sense to keep it compatible with 1.5 rather than 1.6.

If so, either the compat' doesn't matter for the fix and we can work on `main`, or we would work on a zstream branch starting at the corresponding ystream release which won't have the new compat version and `-` match.

On a side node I would not expect to break compat' in a zstream release, however this should not be an issue as we can bump there and insert a compat' version in the other impacted branches (e.g. ystream).

> So the only gap here is the development that happens between this patch and the actual 1.6 release which should be small. For that, what we can do is "always bump the retis version on the same PR that adds the compat version". This PR is special as it's introducing the framework but moving forward this would make sense.
> 
> So the idea would be to have V0: 1.5.* V1: 1.6.0 V2: > 1.6.0
> 
> V2 contains zero fixups.
> 
> During the development of 1.7 we add add fixups to V2 as we break stuff When we release 1.7 we change the above to be: V0: 1.5.* V1: 1.6.* V2: 1.7.0 V3: >1.7.0
> 
> If we release 1.6.1 (zstream) before releasing 1.7, instead we: V0: 1.5.* V1: 1.6.1 V2: > 1.6.1
> 
> WDYT?

I'm not a fan of the above due to the increase in maintenance in the compat' code. We should only have to modify it when a breakage happens (or an old version is not supported anymore). This doesn't work with my proposal too though :) But see below.

I'm also going to answer the other comment as the two are related for a potential way forward,

> Right, currently we do not support <1.5. But we should not keep growing this `match` endlessly. At some point we want to cut it and say, we support the last N releases, at least from a maintanance PoV I think it's sane.
> 
> So, ideally we should have something like:
> 
> ```
> < FIRST_SUPPORTED => V0 (best effort) or even just  error
> = FIRST_SUPPORTED_Y_STREAM = V0
> = SECOND_SUPPORTED_Y_STREAM = V1
> ...
> >= CURRENT_RELEASE = VN
> ```

Yes, that makes sense. Although given the current implementation keeping lots of older versions has no cost on maintenance or perfs.

If I sump up the requirements:

- We need to match the `main` version as well as the next release in a single statement.
- We need to match multiple versions at once to ease maintenance (e.g. all versions prior to X.Y or all versions newer than X.Y). But we only have a single catchall in a match statement.

What about something similar to

> always bump the retis version on the same PR that adds the compat version

by switching from a bump the version on release to bump the version after the release. E.g. first commit after v1.6.0 bumps the version to v1.7.0. This would solve the first point above. We can do or not do the same for zstream versions.

For the second point I'm afraid we need to implement a representation of the version that supports comparisons. It's not that bad.

amorenoz: >  by switching from a bump the version on release to bump the version after the release. E.g. first commit after v1.6.0 bumps the version to v1.7.0. This would solve the first point above. We can do or not do the same for zstream versions.

+1
I think that works fine as well. If we have a z-stream branch we branch it before that commit. 

> For the second point I'm afraid we need to implement a representation of the version that supports comparisons. It's not that bad.

How about https://docs.rs/semver/latest/semver/?

retis-events/src/compat/compat.rs
```diff
@@ -0,0 +1,162 @@
...
+    /* v1.5.x; we start the backward compatibility effort with v1.5.0 */
+    V0 = 0,
+    /* v1.6.x.. */
+    V1,
+}
+
+// Retis versions to internal compat version table.
+const VERSION_MATCHES: &[(&str, CompatVersion)] = &[
+    (">= 1.5.0, < 1.6.0", CompatVersion::V0),
+    (">= 1.6.0", CompatVersion::V1),
```
amorenoz, line 52: So, is your plan to add 1.7.0 to VERSION_MATCHES after we release 1.6?

retis-events/src/compat/json.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    // Otherwise we do a two-step deserialization to allow fixing up fields in
+    // between. The first round deserializes the event into a generic
+    // serde_json::Value (which has the advantage of not enforcing any field
+    // presence or type) and the second one generates our Event from the fixed
+    // up serde_json::Value.
+
+    let mut event: serde_json::Value = serde_json::from_str(input)?;
+    super::compatibility_fixup(&mut event, CompatStrategy::Backward(version))?;
```
amorenoz: I think we should print some `info` log here

retis-events/src/compat/json.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    // TODO: map_err to for better UX (mentioning we tried to fixup the event).
```
amorenoz: Is this a plan for this PR?

retis-events/src/compat/json.rs
amorenoz: And maybe a debug level log displaying the json after the fixups?

retis-events/src/compat/compat.rs
atenart: I don't. 1.7 would match `CompatVersion::V1`. We should only add a line in there if we break compatibility. If that happens during the 1.7 development, we would add a 1.7 line but if that only happens during the 1.8 development we'll only do it then.

retis-events/src/compat/json.rs
atenart: > And maybe a debug level log displaying the json after the fixups?

That would be *a lot* of output. One log line per event.

retis-events/src/compat/json.rs
atenart: I can do this.

docs/compatibility.md
```diff
@@ -0,0 +1,25 @@
+# Compatibility
```
vlrpl, line 1: I think that a couple of hints in `CONTRIBUTING.md` mentioning `VERSION_MATCHES`, `FIXUPS`, and so forth might be beneficial for external developers.
Doesn't need to be a fully fledged guide, but a just enough list of hints.

vlrpl (commented):
Thanks for taking care of this, it is very useful!
I have just a small nit about the doc, but other than that, the series, LGTM.

vlrpl (approved):
the series lgtm, thanks!

PR #552, opened 2025-07-29
[WIP] release 1.6

atenart:
Prepare the release of Retis 1.6.0. This contains a few improvements, but mostly multiple fixes of issues encountered while preparing the release.

Note that #550 might go in first, depending on what we decide.


atenart:
Since the version bump is now part of #550, mark this as ready. It does not have to be the latest PR before releasing. Also there's no particular order of merge between this and #550.

retis/src/cli/cli.rs
```diff
@@ -145,10 +145,10 @@ where
     }
 }
 
-/// Trace packets on the Linux kernel
-///
-/// retis is a tool for capturing networking-related events from the system using ebpf and analyzing them.
 #[derive(Args, Debug, Default)]
+#[command(about = "Trace packets in the Linux networking stack & friends.
```
vlrpl, line 149: Normally I find more common the following template:

`.help_template("{usage-heading} {usage}\n{about-section}\nCommands:\n{subcommands}\n\nOptions:\n{options}{after-help}");`

This has the `about-section` before the usage which makes it a little weird.
Not sure about `.help_template()` and formatting (e.g. underlined+bold heading), which also might not be super convenient.

I'll leave whether to change it or not up to you as this is not a big deal.

atenart: I'm fine with putting the usage first. This needs some investigation though (command and sub-commands need a different template, it's not clear to me what is the current default template and we should stay close to it IMO, etc) and testing. I don't think I'll have time to tackle this right now, although having it for 1.6 would be nice.

vlrpl: I do agree. When I was referring to how convenient this can be, I was also referring to different templates.
AFAIC, I'll approve this, if we have time to fix it, fine, otherwise we can definitely skip it and add it later.
I don't see any reason to rush it.

vlrpl: > I do agree. When I was referring to how convenient this can be, I was also referring to different templates. AFAIC, I'll approve this, if we have time to fix it, fine, otherwise we can definitely skip it and add it later. I don't see any reason to rush it.

Please, resolve the conversation if you end up realizing it cannot be included (for whatever reason) and I'll proceed merging it.

vlrpl (approved):

retis/src/cli/cli.rs
atenart: Let's merge those fixes now to ease testing and maintenance. If one of us have a chance to implement the above before the release that'd be nice, otherwise we can do this after.

vlrpl: sgtm

vlrpl:
merged, thanks!

Issue #534, opened 2025-06-20
Handle types more than once


atenart:
Should we handle types more than once? This can happen in some cases:
- Two arguments of the same function share the same type (e.g. two `struct sk_buff *`).
- The returned type might be the same as one of the arguments.
- Etc.

In such case only the first argument currently ends up in an event. We should investigate and decide what to do here (no support, 1 event per occurrence, duplicate event sections, ...).

PR #554, opened 2025-08-08
More pre-release fixes

vlrpl:
- Fixes rawhide root partition too small
- avoid emitting unneeded relocation + fix type attribute


retis/src/collect/collector/ns/bpf/netns_hook.bpf.c
```diff
@@ -12,38 +12,34 @@ struct netns_event {
...
 	/* Netns cookies are not available on older kernels. */
 	bool get_cookie = bpf_core_field_exists(struct net, net_cookie);
+	struct net_device *dev;
 	struct netns_event *e;
+	struct sk_buff *skb;
 	struct net *net;
 	u64 cookie;
 	u32 inum;
 
-	net = retis_get_net(ctx);
```
atenart, line 20: This was removed but it looks like was not added back. Even if we now make sure tracking let us generate an event, we should still support functions having a netns in their arguments (for now in addition to the skb, which later can be extended to if the stack is being tracked).

vlrpl: ok, that was the idea, meaning including the parameter later.
Of course, if we want to collect the data from the struct net parameter (filtering the skb this time, of course), it should take precedence over the sk_buff (turning this into the diff below).
The fact that we use the parameter instead of the skb is not completely straightforward for the user, and we should probably call this out in the collector doc (unless it is and I'm missing it).

```diff
diff --git a/retis/src/collect/collector/ns/bpf/netns_hook.bpf.c b/retis/src/collect/collector/ns/bpf/netns_hook.bpf.c
index d42e7189f..bdde9a81d 100644
--- a/retis/src/collect/collector/ns/bpf/netns_hook.bpf.c
+++ b/retis/src/collect/collector/ns/bpf/netns_hook.bpf.c
@@ -13,20 +13,19 @@ DEFINE_HOOK_RAW(
        /* Netns cookies are not available on older kernels. */
        bool get_cookie = bpf_core_field_exists(struct net, net_cookie);
        struct netns_event *e;
+       struct sk_buff *skb;
        struct net *net;
        u64 cookie;
        u32 inum;
 
+       skb = retis_get_sk_buff(ctx);
+       if (!skb || !skb_is_tracked(skb))
+               return 0;
+
+       /* struct net parameter takes precedence over the skb. */
        net = retis_get_net(ctx);
        if (!net) {
-               struct net_device *dev;
-               struct sk_buff *skb;
-
-               skb = retis_get_sk_buff(ctx);
-               if (!skb || !skb_is_tracked(skb))
-                       return 0;
-
-               dev = BPF_CORE_READ(skb, dev);
+               struct net_device *dev = BPF_CORE_READ(skb, dev);
                if (!dev)
                        dev = retis_get_net_device(ctx);
```

atenart: That's right and it goes down to #534. You're right that this could be confusing and looking at history taking a parameter over what we find in the skb is a change in behavior. Maybe the function parameter one should be last resort (which will be useful for stack tracking).

atenart: > That's right and it goes down to #534. You're right that this could be confusing and looking at history taking a parameter over what we find in the skb is a change in behavior. Maybe the function parameter one should be last resort (which will be useful for stack tracking).

To me more clear: we can either keep the current form of this PR; or add back the direct netns retrieval but as the last option so that it's already there for later. As you prefer.

vlrpl: IMO, it makes sense to do something like the diff below (fallback option).
I admit that the other way around also makes sense, but given #534, it's better to stay consistent for now and extend the logic as part of #534, in order to include all the combinations (maybe per-probe options can help a little here).

If we agree on the diff below, I'd go ahead and resubmit.

```diff
diff --git a/retis/src/collect/collector/ns/bpf/netns_hook.bpf.c b/retis/src/collect/collector/ns/bpf/netns_hook.bpf.c
index bdde9a81d..5720f24b7 100644
--- a/retis/src/collect/collector/ns/bpf/netns_hook.bpf.c
+++ b/retis/src/collect/collector/ns/bpf/netns_hook.bpf.c
@@ -22,25 +22,27 @@ DEFINE_HOOK_RAW(
        if (!skb || !skb_is_tracked(skb))
                return 0;
 
-       /* struct net parameter takes precedence over the skb. */
-       net = retis_get_net(ctx);
-       if (!net) {
-               struct net_device *dev = BPF_CORE_READ(skb, dev);
-               if (!dev)
-                       dev = retis_get_net_device(ctx);
-
-               /* If the network device is initialized in the skb, use it to
-                * get the network namespace; otherwise try getting the network
-                * namespace from the skb associated socket.
-                */
-               if (dev) {
-                       net = BPF_CORE_READ(dev, nd_net.net);
-               } else {
-                       struct sock *sk = BPF_CORE_READ(skb, sk);
+       struct net_device *dev = BPF_CORE_READ(skb, dev);
+       if (!dev)
+               dev = retis_get_net_device(ctx);
 
-                       if (!sk)
-                               return 0;
+       /* If the network device is initialized in the skb, use it to
+        * get the network namespace; otherwise try getting the network
+        * namespace from the skb associated socket.
+        */
+       if (dev) {
+               net = BPF_CORE_READ(dev, nd_net.net);
+       } else {
+               struct sock *sk = BPF_CORE_READ(skb, sk);
 
+               if (!sk) {
+                       /* we cannot get struct net from the sk_buff,
+                        * try to get it from the parameters.
+                        */
+                       net = retis_get_net(ctx);
+                       if (!net)
+                               return 0;
+               } else {
                        net = BPF_CORE_READ(sk, __sk_common.skc_net.net);
                }
        }
```

atenart: I think `retis_get_net_device` should come third to try all the skb-linked ways first.

vlrpl: makes sense going full skb first

atenart: Thanks!

PR #555, opened 2025-08-12
CI: ensure testing against up-to-date kernels

vlrpl:
The change forces a reboot after system updates to ensure tests run against the latest kernel version.


atenart (approved):
Nice!

PR #556, opened 2025-08-19
A few updates before releasing

atenart:


vlrpl (approved):
the change looks good, thanks!

PR #564, opened 2025-08-29
Allow rotating the output file

atenart:
A new `--out-rotate` option is added in the `collect` sub-command to allow rotating the output file. For now the rotation is based on a size limit but other could be implemented later (e.g. time bound).

### DIFF
--- a/tests/next/include/helpers.py
+++ b/tests/next/include/helpers.py
@@ -40,7 +40,7 @@ def assert_events_present(file, expected):
         found = False
         # Find an event that matches
         for i in range(idx, len(events)):
-            (is_sub, reason) = __is_subset(events[i], ex, aliases)
+            is_sub, reason = __is_subset(events[i], ex, aliases)
             if is_sub:
                 print(
                     f"HIT: Event at index {i}\n"
@@ -123,7 +123,7 @@ def __is_subset(superset, subset, aliases):
 
         # Recursively assert nested dictionaries
         if isinstance(value, dict):
-            (is_sub, reason) = __is_subset(superset[key], value, aliases)
+            is_sub, reason = __is_subset(superset[key], value, aliases)
             if not is_sub:
                 return (
                     False,


retis/src/helpers/file_rotate.rs
```diff
@@ -0,0 +1,153 @@
...
+    path::{Path, PathBuf},
+    str::FromStr,
+};
+
+use anyhow::{anyhow, Result};
+use log::error;
+use regex::Regex;
+
+#[derive(Eq, PartialEq)]
+pub(crate) enum Rotation {
```
amorenoz: nit: `Rotation` seems a bit generic, how about `RotationPolicy`?

retis/src/collect/cli.rs
```diff
@@ -93,6 +93,15 @@ Examples of meta filters:
...
     )]
     pub(super) out: Option<PathBuf>,
+    #[arg(
+        long,
+        value_name = "LIMIT",
+        help = "Rotate the output file (see `--out`) once a given limit is reached. The rotation limit can be:
+- A size: <LIMIT> is in bytes and should be suffixed with a size unit (MB, GB). The file names will be <OUT>.X, with X being a number starting at 0 and increasing over time. e.g. '--out-rotate 64MB'.
+
+Events from the same series might end up on different files. If a previous collection with rotation enabled was not removed only the files required to store the new collection will be overridden."
+    )]
```
amorenoz, line 105: I think we need to either require `--out` or implicitly enable it. Else `retis collect --out-rotate 10` will be confusing.

amorenoz: Given `--out` accepts a Path, I'd say adding `requires = "out"` to the derive macro (and document the dependency) would be enough.

retis/src/collect/cli.rs
```diff
@@ -93,6 +93,15 @@ Examples of meta filters:
         help = "Write the events to a file rather than to sdout. If the flag is used without a file name, defaults to \"retis.data\"."
     )]
     pub(super) out: Option<PathBuf>,
+    #[arg(
+        long,
+        value_name = "LIMIT",
+        help = "Rotate the output file (see `--out`) once a given limit is reached. The rotation limit can be:
+- A size: <LIMIT> is in bytes and should be suffixed with a size unit (MB, GB). The file names will be <OUT>.X, with X being a number starting at 0 and increasing over time. e.g. '--out-rotate 64MB'.
```
amorenoz: This comment needs to be updated according to whatever the regexp ends up being (even if unchanged). See comment on the regexp.

retis/src/helpers/file_rotate.rs
```diff
@@ -0,0 +1,153 @@
...
+    None,
+    Size(usize),
+}
+
+impl Rotation {
+    /// Convert an str representation of a limit to a `Rotation`.
+    ///
+    /// Accepted values are numbers suffixed with a unit size (MB or GB).
+    pub(crate) fn from_str(limit: &str) -> Result<Self> {
+        let re = Regex::new(r"(\d+)(M|G)B")?;
```
amorenoz: Help says:

> A size: <LIMIT> is in bytes and should be suffixed with a size unit (MB, GB).

This is not 100% accurate or at least a bit unclear. It suggests `--out-rotate 10` "should" work and it doesn't.
So how about `(\d+)(K|M|G)?B` (allowing `B` and `MB` and not `KB` seems unfair)?

retis/src/helpers/file_rotate.rs
```diff
@@ -0,0 +1,153 @@
...
+                io::ErrorKind::Unsupported,
+                "Cannot rotate output without a rotation policy",
+            ));
+        }
+
+        // Close and rename the target file.
+        self.close()?;
+
+        // Create the new file.
+        (self.inner, self.written) = Self::new_file(&self.target, self.head.as_deref())?;
```
amorenoz: As stated in the help, events from the same series could end up in different files, making `sort` fail. So I'm trying to think of ways to improve the user experience in that case.

Specifically my mind goes to: Do we need to add the file index in the startup event?

I can see two different use-cases
A ) Rotate collection files but analyze as one big event file
B ) Rotate collection files and analyze each file individually

For A) I would consider also changing `print`, `sort` and `python` to try {FILE} and, if it doesn't exist, try {FILE}.0 and its header says it's a partial event file, start printing/sorting events and when EOF, take {FILE}.1, {FILE}.2, etc. That way you have a cohesive experience in post-processing, regardless of the way you collected. Besides, it keeps the "no need to specify filename" usability treat.

For B) we could raise a warning (or info log) if we `print`, `sort` or `python` a non-zero partial file directly, advising the user to take the previous chunk if events at the beginning of the file look odd.

WDYT?

retis/src/helpers/file_rotate.rs
```diff
@@ -0,0 +1,153 @@
...
+        };
+
+        // Unwrap as the regex already checked the first group was mandatory.
+        let limit = usize::from_str(matches.get(1).unwrap().as_str())? * factor;
+
+        Ok(Rotation::Size(limit))
+    }
+}
+
+pub(crate) struct RotateWriter {
```
amorenoz: Document pub(crate) struct

retis/src/helpers/file_rotate.rs
```diff
@@ -0,0 +1,153 @@
...
+
+#[derive(Eq, PartialEq)]
+pub(crate) enum Rotation {
+    None,
+    Size(usize),
+}
+
+impl Rotation {
+    /// Convert an str representation of a limit to a `Rotation`.
+    ///
```
amorenoz: Extra newline

retis/src/collect/collect.rs
```diff
@@ -539,22 +527,20 @@ impl Collectors {
...
-                )),
+                Box::new(
+                    RotateWriter::new(
+                        out,
+                        match &collect.out_rotate {
+                            Some(s) => Rotation::from_str(s)?,
+                            None => Rotation::None,
+                        },
+                        // When events are stored for later post-processing,
+                        // include a startup event.
```
amorenoz: nit: I don't think this comment adds a lot of information now and having it in the middle of a list of function arguments is kind of distracting for the eye

retis/src/collect/cli.rs
atenart: Makes sense

retis/src/helpers/file_rotate.rs
atenart: > Help says:
> 
> > A size:  is in bytes and should be suffixed with a size unit (MB, GB).
> 
> This is not 100% accurate or at least a bit unclear. It suggests `--out-rotate 10` "should" work and it doesn't.

I meant this as "one of those two unit qualifiers MUST be used", the wording might not be clear though? s/should/must/?

> So how about `(\d+)(K|M|G)?B` (allowing `B` and `MB` and not `KB` seems unfair)?

This was made on purpose, the common event alone is ~160B and adding one or two events can reach the 1KB bar. Not allowing low values helps not being a file bomb. 1MB seemed a good minimal value.

While looking at this I realized we should also forbid 0...

retis/src/helpers/file_rotate.rs
atenart: > I can see two different use-cases A ) Rotate collection files but analyze as one big event file B ) Rotate collection files and analyze each file individually
> 
> For A) I would consider also changing `print`, `sort` and `python` to try {FILE} and, if it doesn't exist, try {FILE}.0 and its header says it's a partial event file, start printing/sorting events and when EOF, take {FILE}.1, {FILE}.2, etc. That way you have a cohesive experience in post-processing, regardless of the way you collected. Besides, it keeps the "no need to specify filename" usability treat.

`print` and `sort` do not read and parse events on demand so that would mean reading everything when we know the collection was too big. On the other hand that would make it more aligned with the non-split case and users could still (as I would expect) use those commands and specifying the input file. I'm a bit worried simply forgetting to specify the events file would end up reading and parsing 100GBs of data.

Also what to do if both `events.data` and `events.dataX` are present? The current logic does not handle this but the expectation is simple. I guess checking for `retis.data` and `retis.data0` would do, if both are there throw an error.

> For B) we could raise a warning (or info log) if we `print`, `sort` or `python` a non-zero partial file directly, advising the user to take the previous chunk if events at the beginning of the file look odd.

That is the main use case here IMO. We can add an info message so expectations are right.

amorenoz: > > I can see two different use-cases A ) Rotate collection files but analyze as one big event file B ) Rotate collection files and analyze each file individually
> > For A) I would consider also changing `print`, `sort` and `python` to try {FILE} and, if it doesn't exist, try {FILE}.0 and its header says it's a partial event file, start printing/sorting events and when EOF, take {FILE}.1, {FILE}.2, etc. That way you have a cohesive experience in post-processing, regardless of the way you collected. Besides, it keeps the "no need to specify filename" usability treat.
> 
> `print` and `sort` do not read and parse events on demand so that would mean reading everything when we know the collection was too big. On the other hand that would make it more aligned with the non-split case and users could still (as I would expect) use those commands and specifying the input file. I'm a bit worried simply forgetting to specify the events file would end up reading and parsing 100GBs of data.
> 

How about asking the user?
```
retis collect -o --out-rotate 10M
retis print retis.data.0
[first chunk appears in pager]
retis.data.1 is available, do you want to print it? ([y], n): 
[second chunk appears in pager]
```

> Also what to do if both `events.data` and `events.dataX` are present? The current logic does not handle this but the expectation is simple. I guess checking for `retis.data` and `retis.data0` would do, if both are there throw an error.
> 

Assuming post-processing is done in the same place where collection is, `retis.data` is overwritten.

Actually, this makes me thing the output file is not very clear now. Even is the non-default value is used, we are not making it clear what's the name of the output file (which we are changing silently changing underneath). Maybe we should inform, e.g:

`INFO retis.data0...retis.data.N files written`

> > For B) we could raise a warning (or info log) if we `print`, `sort` or `python` a non-zero partial file directly, advising the user to take the previous chunk if events at the beginning of the file look odd.
> 
> That is the main use case here IMO. We can add an info message so expectations are right.

retis/src/helpers/file_rotate.rs
amorenoz: > > Help says:
> > > A size:  is in bytes and should be suffixed with a size unit (MB, GB).
> > 
> > 
> > This is not 100% accurate or at least a bit unclear. It suggests `--out-rotate 10` "should" work and it doesn't.
> 
> I meant this as "one of those two unit qualifiers MUST be used", the wording might not be clear though? s/should/must/?
> 

"must" makes it clearer, yes.

> > So how about `(\d+)(K|M|G)?B` (allowing `B` and `MB` and not `KB` seems unfair)?
> 
> This was made on purpose, the common event alone is ~160B and adding one or two events can reach the 1KB bar. Not allowing low values helps not being a file bomb. 1MB seemed a good minimal value.
> 
> While looking at this I realized we should also forbid 0...

OK. That makes sense.

retis/src/helpers/file_rotate.rs
amorenoz: > I'm a bit worried simply forgetting to specify the events file would end up reading and parsing 100GBs of data.

Thinking more about this. This should not be a problem, that's why we have a PAGER, right?
AFAIK, retis would start generating events and `less` would start dequeuing them from the pipe. It will dequeue a few and stop, so retis would block on a write until the users scrolls at which point less will dequeue more, retis will generate more and so on. So, the user controls what events are generated. If the user created a lot of them, and wants to scroll over 100GBs of data, who are we to stop them?

The only caveat to this is the way we currently control EPIPE. This line is problematic:
https://github.com/retis-org/retis/blob/60f63f9cfa623c6676977eb622e19ef90c1c3434/retis/src/process/display.rs#L42
because, when the pager is closed retis still reads, parses and writes (to a broken pipe) the rest of the events. We should not do that and instead stop gracefully knowing that the user is no longer interested in more events. i.e: be more like "git log".

atenart: (I've added a patch to better handle `BrokenPipe`).

You're right about the pager, that is a nice side effect!

The updated version now reads `retis.dataX` files and continue when one reaches EOF. I haven't updated the documentation in `print` or `sort`. The behavior is now:
- Regardless of the input file name (default: "retis.data"), that file is tried and if not present the rotation output name is tried.
- When a rotated input file is detected, upon EOF the next one is opened.

A few comments:
- If a rotation file is explicitly given, e.g. "retis.data0" or "retis.data1", the rotation policy is not detected and that file only will be read.
- This is never true for "retis.data".
- We can't start at e.g. "retis.data3" and continue over the next ones.

I'm not sure how far we should go in terms of reading the next files. Maybe we should rework things to only auto-detect if the input file value was not specified by the user and otherwise handle rotation on demand (new cli option needed). Or restrict the magic to specific cases. Thoughts?

Extra comments:
- Rename "retis.dataX" to "retis.data.X"?
- We might support split output based on time at some point. Those might be named "retis.data.TIMESTAMP". The logic would be slightly more complex to read the next file. (But doable, I guess).

atenart: > I'm not sure how far we should go in terms of reading the next files. Maybe we should rework things to only auto-detect if the input file value was not specified by the user and otherwise handle rotation on demand (new cli option needed). Or restrict the magic to specific cases. Thoughts?

For more clarity, I'm thinking about:

- `retis print`: tries `retis.data` and then `retis.data0` and continues over `retis.data1` etc if any.
- `retis print retis.data` just reads `retis.data`.
- `retis print retis.data0` just reads `retis.data0`.
- `retis print retis.data1 --split-follow` reads `retis.data1` and continues over `retis.data2` etc. (Would `--out-split` better than `--out-rotate`?). This one might be the tricky one to implement, especially if we support more rotation logic (e.g. time based). I'm also thinking we could omit `--split-follow` using a special format, e.g. `retis print retis.data1..`.

amorenoz: I think that approach is  fair. It keeps the expected behavior for `retis print` allowing certain degree of flexibility if we specify the filename. Some comments:
- `retis sort retis.data.N` (not sure about `retis print`) should warn that the output could be incomplete due to splitting.
- `--out-split` looks good to me, although not very superior to `--out-rotate` tbf.
- Do we need to differentiate between `retis.data.0` and `retis.data.[1...]`? Or, more technically: do we need to put add a flag to the startup event to indicate that it's a split file and the split index? I think we don't need it but just double-checking.

atenart: > * `retis sort retis.data.N` (not sure about `retis print`) should warn that the output could be incomplete due to splitting.
> * Do we need to differentiate between `retis.data.0` and `retis.data.[1...]`? Or, more technically: do we need to put add a flag to the startup event to indicate that it's a split file and the split index? I think we don't need it but just double-checking.

If we want to properly implement the first item we should add a split index in the startup event, otherwise we'll end up guessing things based on the file name which could end with a number for other reasons. 

It might actually be a good idea to convey information (index, rotation type) about the file rotation in the events, e.g. in the startup event. This would improve rotation detection at processing time and make it robust, while also be more future proof.

Maybe  we could also set a `last: "true"` when writing the last one (or `next: "..."` in all but the last event), but not sure we can do this in a efficient way. Or we need a dedicated event section we can output at the end.

> * `--out-split` looks good to me, although not very superior to `--out-rotate` tbf.

Ack.

atenart:
Implemented what was discussed lately (mainly improving the reading part, handling split files when asked to) and improved existing things. Functionally speaking things seem to work.

Made the file factory to not depend the specific rotate reader implementation but on a more broad `Read + Seek` one, allowing to decouple the factory from the underlying reader. This helps compartmentalizing things and not use the rotation reader when not needed.

I'm not 100% happy with this though, mainly because the rotation reader is a bit complex and more boiler plate code is required in some places, e.g. for handling the file argument when using the rotation reader. Not sure this can be improved, but happy to get ideas.

So far only the `print` command uses the rotation reader. This can be extended to a few other parts if the implementation is approved.

retis-events/src/helpers/mod.rs
```diff
@@ -0,0 +1,10 @@
+//! # Helpers
+//!
+//! Internal and public helper functions and objects to ease working with
+//! events.
+
+pub mod file;
```
amorenoz: `file` contains logic that is pretty core to what the library does (reading/writing events to files). It is exposed both to retis core create and indirectly to the python library.
OTOH, `helpers` suggest auxiliary stuff, only needed internally.

I don't think both fit well together.

retis-events/src/common.rs
```diff
@@ -18,6 +23,16 @@ impl EventFmt for StartupEvent {
     }
 }
 
+/// Information about a partial event file generated while splitting the full
+/// collection into multiple files.
+#[event_type]
```
amorenoz, line 55: Commented here: https://github.com/retis-org/retis/pull/533#issuecomment-3401591748

retis-events/src/file/rotate.rs
```diff
@@ -0,0 +1,181 @@
...
+    path::{Path, PathBuf},
+};
+
+use anyhow::Result;
+use log::{error, info};
+
+use crate::{helpers::time::*, *};
+
+#[event_type]
+#[derive(Copy, Eq, PartialEq)]
```
amorenoz, line 21: We need a comment on a public event type.

retis-events/src/file/rotate.rs
```diff
@@ -0,0 +1,181 @@
...
+    // Controls how rotation is done.
+    policy: Option<RotationPolicy>,
+    // Target file name for the output. Will be suffixed following the rotation
+    // policy rules.
+    target: PathBuf,
+    // Current index of the file being written, starting with 0.
+    index: u32,
+    // Total bytes written across the set of split files.
+    written: usize,
+    monotonic_offset: TimeSpec,
```
amorenoz, line 45: comment all fields? some might be obvious but, just for the sake of uniformity

retis-events/src/file/rotate.rs
```diff
@@ -0,0 +1,181 @@
...
+        // Create the new file.
+        (self.inner, self.written) = Self::new_file(&self.target, &startup)?;
+
+        Ok(())
+    }
+}
+
+impl Write for RotateWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        if let Some(policy) = &self.policy {
```
amorenoz, line 145: Shorten to if let Some(RotationPolicy::Size(limit))?

retis-events/src/helpers/file_rotate.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// Create a new `RotateWriter` given a file path and a rotation policy.
+    pub fn new<P: AsRef<Path>>(
+        file: P,
+        policy: Option<RotationPolicy>,
+        monotonic_offset: TimeSpec,
+    ) -> Result<Self> {
+        // Generate a startup event.
+        let index = 0;
+        let mut startup = serde_json::to_vec(&startup_event(monotonic_offset, index, policy)?)?;
+        startup.push(b'\n');
```
amorenoz: Maybe move the newline addition to `new_file` and avoid the one in `rotate` as well?

amorenoz: Or even move the entire startup_event generation as well?

retis/src/collect/cli.rs
```diff
@@ -94,6 +94,16 @@ Examples of meta filters:
...
     )]
     pub(super) out: Option<PathBuf>,
+    #[arg(
+        long,
+        requires = "out",
+        value_name = "LIMIT",
+        help = "Rotate the output file (see `--out`) once a given limit is reached. The rotation limit can be:
+- A size: <LIMIT> is in bytes and must be suffixed with a size unit (MB, GB). The file names will be <OUT>.X, with X being a number starting at 0 and increasing over time. e.g. '--out-rotate 64MB'.
+
+Events from the same series might end up on different files. If a previous collection with rotation enabled was not removed only the files required to store the new collection will be overridden. This includes \"retis.data\""
```
amorenoz: `retis.data` or the alternative name provided in `--out`?

retis-events/src/file/rotate.rs
```diff
@@ -0,0 +1,181 @@
...
+
+use anyhow::Result;
+use log::{error, info};
+
+use crate::{helpers::time::*, *};
+
+#[event_type]
+#[derive(Copy, Eq, PartialEq)]
+pub enum RotationPolicy {
+    Size(usize),
```
amorenoz: Have you considered adding a `None` variant and avoid the use of `Option<RotationPolicy>` all over?

retis-events/src/file/file.rs
```diff
@@ -22,23 +18,20 @@ pub enum FileType {
...
 impl FileEventsFactory {
-    pub fn new<P>(file: P) -> Result<Self>
-    where
-        P: AsRef<Path>,
-    {
-        let mut reader = BufReader::new(
-            File::open(&file)
-                .map_err(|e| anyhow!("Could not open {}: {e}", file.as_ref().display()))?,
-        );
+    pub fn new(reader: Box<dyn ReadSeek>) -> Result<Self> {
```
amorenoz, line 46: This changes the API for every user of  `FileEventsFactory` who now have to add `Box::new(File::open()` to their constructors, just to allow one of them to provide the reader directly.
Would you consider another constructor, i.e: leave `new` as it was (building the Boxed File internally) and add a `with_reader` for the rotation case?

retis-events/src/helpers/file_rotate.rs
```diff
@@ -179,3 +184,154 @@ pub(crate) fn startup_event(
...
+/// is detected automatically.
+pub struct RotateReader {
+    inner: File,
+    // Controls how rotation is done.
+    policy: Option<RotationPolicy>,
+    // Target file name for the output. Will be suffixed following the rotation
+    // policy rules.
+    target: PathBuf,
+    // Try opening next split files?
+    follow_policy: bool,
```
amorenoz: nit: variable called "policy" that only has a true/false value is weird.

amorenoz: Plus it gets confusing further below when code deals with rotation policy.

retis-events/src/file/rotate.rs
```diff
@@ -179,3 +184,154 @@ pub(crate) fn startup_event(
...
+        if let Some(startup) = event.startup {
+            if let Some(split) = startup.split_file {
+                match split.policy {
+                    RotationPolicy::Size(_) => {
+                        if path.extension()
+                            == Some(<String as AsRef<OsStr>>::as_ref(&format!("{}", split.id)))
+                        {
+                            path.set_extension("");
+                            return Ok((path, split.id, Some(split.policy)));
+                        }
```
amorenoz, line 305: Maybe we should print an error if the path extension does not match the split id from the startup event?

retis-events/src/common.rs
```diff
@@ -18,6 +23,16 @@ impl EventFmt for StartupEvent {
...
+pub struct SplitFile {
+    /// Split file id in the set of split files; first one is "0".
```
amorenoz, line 57: Why not display this? It could be useful if we're scrolling through a rotated file-set to see the file index were starting.

retis/src/process/cli/print.rs
```diff
@@ -21,10 +24,17 @@ use crate::{
...
-    pub(super) input: PathBuf,
-    #[arg(long, help = "Format used when printing an event.")]
+    #[arg(
+        help = "File from which to read events. By default both the single default filename and its split version are used.
+
+When reading a split file, the next one in the set will be read after EOF if:
+- Not explicitly setting this parameter (aka using the default behavior).
+- Or appending '..' to the file name, e.g. `retis.data.1..`.
+
+[default: retis.data + retis.data.0]"
```
amorenoz: nit: The "+" sign is a bit confusing here.

atenart:
> One question: why not add `RotateReade`

Just because I was waiting on feedback before moving all users, but it's planned.

> (and the SIGPIPE fix) to `sort`?

I think that's the case already?

retis-events/src/helpers/mod.rs
atenart: So maybe the fix would be to rename `helpers`? Any suggestion?

The initial goal was all top-level files are about the events, "helpers" to manipulate events can be in their own sub-crate.

retis-events/src/file/rotate.rs
atenart: I thought about that but the current form ease addition of new rotation policy variants.

retis-events/src/helpers/file_rotate.rs
atenart: > Or even move the entire startup_event generation as well?

I would prefer not to, this is linked to the conversion of the event to a given representation (json) + a specific use (in an event file where events are split by `\n`). I moved it to `new_file`.

retis-events/src/file/rotate.rs
atenart: That was my initial implementation, but IMO `None` is not a rotation policy per say. The absence of one is. Also since I'm directly using the type to embed the policy into events, using this form made more sense.

retis-events/src/file/file.rs
atenart: Good idea! (I'll still rename `new` to something more expressive).

retis-events/src/helpers/file_rotate.rs
atenart: Should we instead set the policy to `None` when we force the policy not to be used?

retis-events/src/file/rotate.rs
atenart: I added a warning.

retis/src/process/cli/print.rs
atenart: I used "then" instead of "+", but suggestions welcomed.

retis-events/src/helpers/mod.rs
amorenoz: Making something in `helpers` one of your main API entrypoints feels weird, that's all. I think it's counter-intuitive.
It just seems strange that `file` which has pretty core functionality that this crate exports is at the same level as `U128` and `TimeSpec`. The two latter do fit pretty well into the "helpers" category. But everything in `file` I think deserves it's own `mod`.

retis-events/src/file/rotate.rs
amorenoz: But there are no plans to add them, right? When it comes, we can change them instead of leaving a more verbose code for an undetermined amount of time? Not a very strong opinion though

retis-events/src/file/rotate.rs
amorenoz: > That was my initial implementation, but IMO None is not a rotation policy per say. The absence of one is. Also since I'm directly using the type to embed the policy into events, using this form made more sense

Maybe..., maybe that's because the type is called `RotationPolicy`, if you remove the `Policy` part or rename it to `Config` for instance, then `None` makes more sense :-)

I didn't get the second part

retis-events/src/helpers/file_rotate.rs
amorenoz: I was not even considering the implementation, although your suggestion would improve clarity IMO.
What I meant was that when I first read "follow_policy" I thought "this is another policy about how to follow something" but it only has "true/false" values....So we have "rotation policy" and another "file following policy" and further below the code is a fight between these two policies. You could also  read it as whether to follow or not the policy specified in `policy`.

Maybe it was just a brain fart form my side. Maybe just "ignore" or "obey" would improve readability, idk

Nevertheless, in terms of implementation, storing `policy: None` depending on the flag and avoid having it as struct member is cleaner.

retis-events/src/helpers/mod.rs
atenart: I see, so splitting what's in `helpers`, e.g. putting the file related things in `file`. That makes sense, I'll update the PR with this.

retis-events/src/file/rotate.rs
amorenoz: I don't have a very strong opinion on this either, just a suggestion.

retis-events/src/common.rs
amorenoz: Based on the discussions on #533 , if we're keeping the newline trick, could we just replace this with:
```rust
/// Split-file information
///
/// Information about a partial ...
```

retis-events/src/file/rotate.rs
atenart: We could also make the `RotationPolicy` enum a bool. That's not wanted to keep things more future proof in the stored events file. The above is IMO more clear in this specific case: we're handling all variants, not just one. (Then there's only one atm). Not a super strong opinion either tbf :)

retis-events/src/file/rotate.rs
atenart: > > That was my initial implementation, but IMO None is not a rotation policy per say. The absence of one is. Also since I'm directly using the type to embed the policy into events, using this form made more sense
> 
> Maybe..., maybe that's because the type is called `RotationPolicy`, if you remove the `Policy` part or rename it to `Config` for instance, then `None` makes more sense :-)

Fair enough :)

> I didn't get the second part

The rotation policy lies into a `SplitFile` (itself an `Option`). Having a `None` variant there would be confusing IMO. I'm open for other ways to encode the split information though.

retis-events/src/helpers/file_rotate.rs
atenart: I see. I think using `None` does solve the above, as it convey the information in a single place and avoid the two concepts live in parallel — as this basically removes the `follow_policy` flag.

atenart: I went a slightly different way: if there's no intent to follow a rotation policy... we just do not use the `RotateReader` :) This makes things more simple and should fix this.

atenart:
Comments should be taken into account. I've also converted the `sort` and `pcap` sub-commands in addition to the `print` one.

It's worth noting the `python` sub-command and Python helpers have not been converted yet. There's no technical blocker (they use a `FileEventsFactory` already), but that requires some more work. Let's see if the current form is acceptable first, and then decide if we tackle this as part of this PR or in a follow-up.

retis/src/helpers/file_rotate.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/// Given an "input" command line option (e.g. from the "print" or "sort"
+/// sub-commands), returns a `FileEventsFactory`. This handles the following
+/// cases:
+/// - No input is given: use the default behavior ("retis.data.." then
+///   "retis.data.0..").
+/// - A range is given: use `RotateReader` as the underlying reader.
+/// - A single file is given: use a non-rotate reader.
+pub(crate) fn factory_from_retis_data(input: Option<&PathBuf>) -> Result<FileEventsFactory> {
```
amorenoz: the name is a bit confusing, as it's not reading retis data at all. It's just the input argument value, i.e: name of file

Also, have you considered encoding this into a custom type e.g: `InputArgument`

retis-events/src/file/rotate.rs
amorenoz: I think mixing both things (how we encode the information in the event file and how we deal with configuration of internal structs) can be a source of confusion, but atm creating another struct  for just a single field seems a bit overkilling

retis/src/helpers/file_rotate.rs
atenart: > the name is a bit confusing, as it's not reading retis data at all. It's just the input argument value, i.e: name of file

`factory_from_path`?

> Also, have you considered encoding this into a custom type e.g: `InputArgument`

I'm not sure to get how that would look like, can you show a (brief) example?

amorenoz:
> > One question: are you planning to add support for the python subcommand now or in an independent PR. I do think we should have it in all commands for consistency.
> 
> Ah right, I forgot that part...
> 
> This comes down to how to support this in the Python readers, e.g. `PyEventReader`. I'm not sure we can do the same as for the cli (e.g. `EventFile("pat/to/retis.data..")`) as the `..` look a bit weird here. Should we default to always using the rotation reader? Or introduce a dedicated constructor (can we do that?)?
> 
> Maybe this needs a dedicated PR? I can also work on this here, as you prefer.

We can tackle it in an independent PR, I just want to avoid forgetting about it.

retis/src/helpers/file_rotate.rs
amorenoz: > > the name is a bit confusing, as it's not reading retis data at all. It's just the input argument value, i.e: name of file
> 
> `factory_from_path`?
> 

Works better.

> > Also, have you considered encoding this into a custom type e.g: `InputArgument`
> 
> I'm not sure to get how that would look like, can you show a (brief) example?

I was thinking in something such as this

The following encapsulates our argument logic:
```rust
/// A custom argument type for retis data files that handles rotation logic.
/// This encapsulates the file factory creation logic for use with clap.
#[derive(Debug, Clone)]
pub struct RetisDataFile {
    path: PathBuf,
    use_rotation: bool,
    try_split: bool
}

impl RetisDataFile {
    pub fn factory(self) -> Result<FileEventsFactory> {
        if self.use_rotation {
            FileEventsFactory::new(Box::new(RotateReader::new(self.path, self.try_split)?))
        } else {
            let path_str = self
                .path
                .to_str()
                .ok_or_else(|| anyhow!("Cannot convert path to str"))?;
            FileEventsFactory::from_path(path_str)
        }
    }

    fn from_str(path_str: &str) -> Result<Self> {
        let use_rotation = path_str.ends_with("..");
        let base_path = if use_rotation {
            PathBuf::from(path_str.strip_suffix("..").unwrap())
        } else {
            PathBuf::from(path_str)
        };
        Ok(Self {
            path: base_path,
            use_rotation,
            try_split: false,
        })
    }
}

impl Default for RetisDataFile {
    fn default() -> Self {
        // Default behavior
        Self {
            path: PathBuf::from("retis.data"),
            use_rotation: true,
            try_split: true,
        }
    }
}

impl fmt::Display for RetisDataFile {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        if self.use_rotation {
            write!(f, "{} (with rotation)", self.path.display())
        } else {
            write!(f, "{}", self.path.display())
        }
    }
}
```

The following makes it parsable by clap:
```rust
/// Custom clap value parser for RetisDataFile
#[derive(Clone, Debug)]
pub struct RetisDataFileParser;

impl TypedValueParser for RetisDataFileParser {
    type Value = RetisDataFile;

    fn parse_ref(
        &self,
        _cmd: &Command,
        _arg: Option<&Arg>,
        value: &std::ffi::OsStr,
    ) -> Result<Self::Value, Error> {
        let path_str = value
            .to_str()
            .ok_or_else(|| Error::new(ErrorKind::InvalidValue))?;

        // Create using the constructor (and validate?)
        let retis_file = RetisDataFile::from_str(path_str).map_err(|e| {
            Error::raw(
                ErrorKind::InvalidValue,
                format!("Invalid retis data file: {}", e),
            )
        })?;

        Ok(retis_file)
    }
}
```

Then print.rs does:

```rust
#[derive(Parser, Debug, Default)]
#[command(name = "print", about = "Print stored events to stdout.")]
pub(crate) struct Print {
    #[arg(value_parser = RetisDataFileParser, help = "File from which to read events:
- If a file name is given, it is read and processing stops at EOF. E.g. 'retis.data'.
- If '..' is appended to a file name, it is read and if it is a split file following ones will be read at EOF (if any). E.g. 'retis.data.2..'.
[default: 'retis.data..', then 'retis.data.0..']")]
    pub(super) input: Option<RetisDataFile>,
   // ...
}

impl SubCommandParserRunner for Print {
    fn run(&mut self, _: &MainConfig) -> Result<()> {
        // Create running instance that will handle signal termination.
        let run = Running::new();
        run.register_term_signals()?;

        // Create event factory.
        // NOTE Have to clone because self is &mut, maybe SubcommandParser should own self?
        let mut factory = self.input.clone().unwrap_or_default().factory()?;
       // ...
```

Before I wrote this I thought I was going to be  able to just remove the `Option` in the argument, and just let clap initialize the `RetisDataFile` with its `default()` implementation if the user does not provide an argument. Unfortunately, clap does not do that. It turns out clap calls `RetisDataFileParser::parse_ref` with `RetisDataFile::default().display()`. Since our default logic cannot be expressed in the cli, this does not work...

In any case, my dump example does not do any validation but we could very well do it in `RetisDataFile::from-str` (e.g check the path exists) and that would be kept in clap argparsing which yields nicer errors.
It's just another way (maybe simpler?) of encapsulating the logic

atenart: Thanks for the very detailed example! Although it's slightly more complex and larger than the proposed helper function, I think that makes a lot of sense and will help maintenance. I'll use something along the lines for the next version.

amorenoz: Aha, FromStr is simpler than custom `TypedValueParser`!

PR #565, opened 2025-08-29
pcap: insert VLAN accel data into the payload

atenart:
When using AF_PACKET utilities, if the VLAN information is hw accelerated it is inserted in the kernel before the packet reaches userspace. Because of this, those tools display VLAN data as part of the packet even when accelerated.

This is not the case in Retis, as we report the packet as-is and the VLAN hw acceleration data as part of the event. When using Retis, both cases work well as the VLAN data is printed one way or another (in the packet or as an event section). But when converting the packet to the PCAP format, the result is hw acceleration data is not inserted and the result differs from what we can expected when using AF_PACKET utilities.

Insert the VLAN tag when accelerated by default when converting events to PCAP data in Retis, to match the other utilities expectations. This can be disabled with a cli flag.

---

Note that with #533 the hw accel data will be part of the packet in the metadata. But that won't cover all cases (e.g. working directly with tcpdump) and IMO following the expectation of such tools by default make sense.

### DIFF
--- a/retis/src/process/cli/pcap.rs
+++ b/retis/src/process/cli/pcap.rs
@@ -79,7 +79,7 @@ impl EventParser {
     }
 
     /// Parse & process a single Retis event.
-    fn parse(&mut self, event: &Event) -> Result<Vec<Block<'_>>> {
+    fn parse(&mut self, event: &Event, fixups: bool) -> Result<Vec<Block<'_>>> {
         // Having a common & a kernel section is mandatory for now, seeing a
         // filtered event w/o one of those is bogus.
         let common = event
@@ -150,15 +150,34 @@ impl EventParser {
             }
         };
 
-        // Add the packet itself.
+        // Add the packet itself. If VLAN h/w offload information is there,
+        // insert it in the payload.
+        let mut data = packet.data.0.clone();
+        let mut len = packet.len;
+        match &event.skb {
+            Some(skb) if fixups && skb.vlan_accel.is_some() => {
+                // Unwrap as we just made sure the section is there.
+                let vlan_accel = skb.vlan_accel.as_ref().unwrap();
+                let tag = (vlan_accel.proto as u32) << 16
+                    | ((vlan_accel.pcp & 0x7) as u32) << 13
+                    | (vlan_accel.dei as u32) << 12
+                    | (vlan_accel.vid & 0xfff) as u32;
+
+                data.splice(12..12, tag.to_be_bytes());
+
+                len += 4;
+            }
+            _ => (),
+        }
+
         v.push(
             EnhancedPacketBlock {
                 interface_id: id,
                 timestamp: Duration::from_nanos(i64::from(
                     TimeSpec::new(0, common.timestamp as i64) + self.ts_off.unwrap_or_default(),
                 ) as u64),
-                original_len: packet.len,
-                data: Cow::Borrowed(&packet.data.0),
+                original_len: len,
+                data: Cow::Owned(data),
                 options: vec![EnhancedPacketOption::Comment(Cow::Owned(format!(
                     "probe={}:{}",
                     &kernel.probe_type, &kernel.symbol
@@ -215,6 +234,12 @@ pub(crate) struct Pcap {
         help = "Write the generated PCAP output to a file rather than stdout"
     )]
     pub(super) out: Option<PathBuf>,
+    #[arg(
+        long,
+        help = "Post-processing might be performed on events when generating a PCAP output. This disable such fixups to keep the PCAP output entirely in sync with the events from the input. List of fixups:
+- VLAN hardware acceleration: if available (the vlan_accel section of the skb event section is present), a VLAN header is inserted in the packet payload."
+    )]
+    pub(super) no_fixups: bool,
     #[arg(default_value = "retis.data", help = "File from which to read events")]
     pub(super) input: PathBuf,
 }
@@ -275,6 +300,7 @@ impl SubCommandParserRunner for Pcap {
 
         handle_events(
             self.input.as_path(),
+            !self.no_fixups,
             &filter,
             &mut EventParser::new(),
             write_block,
@@ -286,6 +312,7 @@ impl SubCommandParserRunner for Pcap {
 /// Internal logic to retrieve our events to feed the parser.
 fn handle_events<F>(
     input: &Path,
+    fixups: bool,
     filter: &dyn Fn(&str, &str) -> bool,
     parser: &mut EventParser,
     mut writer_callback: F,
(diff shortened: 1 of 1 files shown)


retis/src/process/cli/pcap.rs
```diff
@@ -150,15 +150,34 @@ impl EventParser {
             }
         };
 
-        // Add the packet itself.
+        // Add the packet itself. If VLAN h/w offload information is there,
```
amorenoz, line 153: Does it really have to do with hardware? h/w offload brings a lot of other stuff to my mind, isn't this just vlan-acceleration?

retis/src/process/cli/pcap.rs
```diff
@@ -150,15 +150,34 @@ impl EventParser {
...
+        // insert it in the payload.
+        let mut data = packet.data.0.clone();
```
amorenoz, line 155: how about checking "fixups" before cloning?

retis/src/process/cli/pcap.rs
```diff
@@ -150,15 +150,34 @@ impl EventParser {
...
+        let mut len = packet.len;
+        match &event.skb {
+            Some(skb) if fixups && skb.vlan_accel.is_some() => {
+                // Unwrap as we just made sure the section is there.
+                let vlan_accel = skb.vlan_accel.as_ref().unwrap();
+                let tag = (vlan_accel.proto as u32) << 16
+                    | ((vlan_accel.pcp & 0x7) as u32) << 13
+                    | (vlan_accel.dei as u32) << 12
+                    | (vlan_accel.vid & 0xfff) as u32;
```
amorenoz, line 164: Better to centralize packet bit manipulation in pnet_packet to ease maintenance.

retis/src/process/cli/pcap.rs
```diff
@@ -215,6 +234,12 @@ pub(crate) struct Pcap {
         help = "Write the generated PCAP output to a file rather than stdout"
     )]
     pub(super) out: Option<PathBuf>,
+    #[arg(
+        long,
+        help = "Post-processing might be performed on events when generating a PCAP output. This disable such fixups to keep the PCAP output entirely in sync with the events from the input. List of fixups:
+- VLAN hardware acceleration: if available (the vlan_accel section of the skb event section is present), a VLAN header is inserted in the packet payload."
```
amorenoz, line 240: Considering #533 , this may become confusing....

retis/src/process/cli/pcap.rs
atenart: This is referenced as "VLAN h/w offload" or "VLAN acceleration" in the kernel and in the ml. If that is confusing in Retis we can use "VLAN acceleration" as this is the name of the event field.

retis/src/process/cli/pcap.rs
atenart: I didn't find a way to only clone if we fixup the packet, but might have missed something.

But anyway this has to be seen in link with how that is transformed into a PCAP block. We either have:

- `clone()` then `Cow::Owned`
- or `Cow::Borrowed` then `into_owned()`, which will `clone()` it.

retis/src/process/cli/pcap.rs
atenart: I'm not sure what is the proposal here?

retis/src/process/cli/pcap.rs
atenart: Yeah, that's a fair point as we could end up populating the VLAN acceleration data into the packet and still providing it as part of the metadata attached to the packet.

We could make this opt-in and issue a warning about the duplication. Or decide Retis is about what is really happening in the kernel and drop this.

amorenoz: Right, I think "Retis is about what is really happening in the kernel" is a fair statement, and deviating from it could cause confusion as users would not know if the packet has been fixed up by retis in any way.

retis/src/process/cli/pcap.rs
amorenoz: Sorry, the proposal is to use a `pnet_packet::MutableVlanPacket` and populate the fields using the helper functions.

atenart: I'm not sure this would make things easier, we still need to move data, there's Ethernet data before the VLAN packet, etc. Also the VLAN tag is quite simple to construct.

retis/src/process/cli/pcap.rs
atenart: So, not even as an opt-in option?

amorenoz: I'm not 100% sure.
Are there corner cases in the kernel where for example the vlan acceleration metadata is present in skb _and_ in the packet?

amorenoz: @vlrpl what do you think? Generally, should we fixup the packet to _try_ to make it easier to understand or should we remain faithful to what we saw in the kernel and use #533 to also report things like vlan accel?

atenart: > Are there corner cases in the kernel where for example the vlan acceleration metadata is present in skb and in the packet?

In case of stacked VLANs it's possible to have one in the metadata and the one in the packet.

I agree that by default we should make things look like what they are in the kernel and not the wire. So it's a question of providing this as an option or not. I don't have strong opinions here tbf.

vlrpl: > @vlrpl what do you think? Generally, should we fixup the packet to _try_ to make it easier to understand or should we remain faithful to what we saw in the kernel and use #533 to also report things like vlan accel?

I guess it's under the prerogatives of post processing to fix things up, or to say it differently, post-process things up. There's a but, though :D 
For pcap as it is now, makes sense to repush it into the header, as we're essentially adapting our events to a specific representation, and AF_PACKET utilities repush it.

Given (and depending on) the future direction, though, where we're adapting the representation to kind of accomodate our events, my preference is to avoid hidden behaviors as they will turn out to be misleading.

atenart: > Given (and depending on) the future direction, though, where we're adapting the representation to kind of accomodate our events

Keeping in mind this will only be true for only one of the pcap consumers, wireshark, not e.g. tcpdump.

vlrpl: > > Given (and depending on) the future direction, though, where we're adapting the representation to kind of accomodate our events
> 
> Keeping in mind this will only be true for only one of the pcap consumers, wireshark, not e.g. tcpdump.

true, and speaking about wireshark, pushing vs not pushing has an impact on display filters as I guess we could potentially have different events having the same header placed in the meta data and others in the data (depending on the probe point), but the display filters are expressive enough to not make this a blocker (the cost is probably a little more verbosity in the expression for such [rare] cases).

On a general note, if we're using something existing it's absolutely desirable to honor other consumer's behavior, whereas with a different representation, IMO, the perspective changes and it seems more natural maybe even more robust (last famous words :) to stick with a different and "closer to the kernel" way.

amorenoz: > Keeping in mind this will only be true for only one of the pcap consumers, wireshark, not e.g. tcpdump.

Unfortunately, libpcap does not seem to report even the comments back to the client.

PR #567, opened 2025-09-01
Container related improvements

atenart:
A few container related improvements; mainly the build and availability of an arm64 version. The build was tested here: https://github.com/retis-org/retis/actions/runs/17374493237/job/49317674484 and the generated images are available under the `test-multi` tag.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -15,7 +15,7 @@ $ retis --help
 
 ### Container image
 
-We provide a script to run Retis in a container,
+We provide a script to run Retis in a container for x86_64 and aarch64 targets,
 [retis_in_container.sh](https://raw.githubusercontent.com/retis-org/retis/main/tools/retis_in_container.sh).
 The current directory is mounted with read-write permissions to the container
 working directory. This allows Retis to read and write files (eg. events, pcap).
--- a/.github/workflows/build_push_image.yaml
+++ b/.github/workflows/build_push_image.yaml
@@ -15,6 +15,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Install qemu dependency
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y qemu-user-static
+
       - name: Checkout code
         uses: actions/checkout@v4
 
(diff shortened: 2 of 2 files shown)


docs/install.md
```diff
@@ -15,7 +15,7 @@ $ retis --help
 
 ### Container image
 
-We provide a script to run Retis in a container,
+We provide a script to run Retis in a container on x86_64 and aarch64 targets,
```
amorenoz: nit:  s/container on/container for/?

tools/retis_in_container.sh
```diff
@@ -64,9 +63,10 @@ if [ -f ${OVS_RUNDIR:-/var/run/openvswitch}/ovs-vswitchd.pid ]; then
 fi
 
 # Run the Retis container.
-exec $runtime run $extra_args $term_opts --privileged --rm --pid=host \
+exec $runtime run $extra_args $term_opts --rm --pid=host \
+      --cap-drop all --security-opt no-new-privileges --read-only --net none \
+      --cap-add SYS_ADMIN --cap-add BPF --cap-add CAP_PERFMON --cap-add SYSLOG --cap-add DAC_OVERRIDE \
```
amorenoz: For `--ovs-track` to work we need `SYS_PTRACE`, and for `--allow-system-changes` we need `CAP_NET_ADMIN`

atenart: Thanks for testing! I forgot to test with those optional modes...

More generally, this is the tricky commit from this PR. I think there's value in restricting the container as much as we can, but that also comes with potential issues and it's not like it'll run without having wide rights. There's a balance between making it a bit more contained and usability, so feel free to push back on this.

amorenoz: I thought of it, but having a known list of needed capabilities is useful, even if it comes with some extra maintenance cost of increasing them when we add a new feature.
Maybe we could use this in our CI tests?

atenart: I'd like to see a container image build + use in our runtime CI, testing both the image generation and the script; which should cover most of the capability testing (given the right options are used). That's on my TODO list for later already.

PR #569, opened 2025-09-04
docs: remove limitation on early probes

atenart:
This is no longer true since we're parsing the packet in userspace. Packets will be captured from the MAC or network offset in all cases up until the limit or skb len, not metadata from in-kernel parsing.


amorenoz:
LGTM, thanks!

PR #570, opened 2025-09-05
collect: explicit the automatic collectors selection

atenart:
Auto loading and not failing on missing prerequisites for collectors was relying on the --collectors option not being set. This is an issue when using profiles and especially when combining them as reported by Adrián.

Making such logic explicit allows to rely on the behavior in profiles. It has the nice consequence of also allowing to combine required and optional collectors, e.g. "--collectors auto,ovs" where all collectors can fail to load but the ovs one.


retis/src/collect/collect.rs
```diff
@@ -367,7 +366,7 @@ impl Collectors {
         //   might add probes and we could be interested in getting those only.
         if main_config.profile.is_empty()
             && collect.probes.is_empty()
-            && collect.collectors.is_none()
+            && (collect.collectors.len() == 1 && collect.collectors[0] == "auto")
```
amorenoz: Maybe simplify by doing `collect.collectors.eq(&["auto"])`?

amorenoz (commented):
Also, documentation needs to be update.

amorenoz (approved):
LGTM, thanks.

PR #573, opened 2025-09-09
Bump libbpf-sys to version 1.6.1

### LINKED ISSUE #572: Build error with clang version 21.1.0
A build failure was observed in our CI (Rawhide):

https://cirrus-ci.com/task/4979628978208768

The problem is in the libbpf which in turn appear in our build because we use the vendored headers shipped with libbpf-sys bindings.

Further details about the issue can be found in the fix
https://github.com/libbpf/libbpf/commit/d88ca951338a765bcb28672dc7b1221fde8e4ae4

A new version of libbpf-sys containing the fix is already available, but it's not yet consumed by libbpf-rs.
For the time being the plan is to wait for libbpf-sys bump in libbpf-rs as this is very likely to happen soon.

vlrpl:
This will fix #572 and our CI.
We'll be using the beta releases of libbpf-{rs,cargo} until the non beta ones are released, but that's fine for main as they will not be part of any release and in the future will be replaced by the non beta versions.


retis/Cargo.toml
```diff
@@ -38,8 +38,8 @@ clap_complete = "4.4"
 elf = "0.8"
 events = {version = "1.5", path = "../retis-events", package="retis-events" }
 flate2 = "1.0"
-libbpf-rs = "0.25"
-libbpf-sys = "1.5"
+libbpf-rs = "=0.26.0-beta.1"
+libbpf-sys = "1.6.1"
```
atenart: I don't think we need to specify `.1` here, do we?

vlrpl: Do you mean for libbpf-sys, right?
I think it's ok having 1.6 ony. I was a bit unsure because of the `=...` in libbpf-rs.
We should be fine without it as well.

atenart: Yes, libbpf-sys, sorry.

PR #575, opened 2025-09-24
add timestamp to bpf log + bindings update

vlrpl:


atenart:
IMO we should report the timestamp as part of `retis_log_event` instead. This would defer formatting to the Rust side and allow for more flexibility.

retis/src/core/events/bpf.rs
```diff
@@ -254,7 +254,7 @@ impl BpfEventsFactory {
             };
 
             match raw_to_string!(&log_event.msg) {
-                Ok(msg) => log!(log_level, "[eBPF] {msg}"),
+                Ok(msg) => log!(log_level, "[eBPF] {} {msg}", log_event.ts),
```
atenart: It's worth noting this won't follow the time format specified by the user (e.g. `--utc`). My impression is this shouldn't be too hard to add support for. WDYT?

vlrpl: honestly, this is a debug feature and there's no requirement for it to be tied to UX.
That being said, we can do this, of course it would require BpfEventsFactory to store the monotonic_offset and consume it.
Having a look, though, I noticed something I missed and is that the monotonic_offsets are detached from each other (e.g. initial event and printer). This doesn't seem right even if we're talking about ns, does it?
One may argue that we may want to remove the initial event if we don't store, but still we can run --print when storing.
I think that should be fixed and we can store it at `Collectors::new()` level once and for all (or alternatively, if we want to be fancy, we can lazily store it at `Collect` level with something like `#[skip_arg]`).
The point is that those offsets should not be taken as we go, but once, or am I missing something?

atenart: > honestly, this is a debug feature and there's no requirement for it to be tied to user UX. That being said, we can do this, of course it would require BpfEventsFactory to store the monotonic_offset and consume it. 

I was thinking being able to see when an error is thrown wrt to events might be wanted, otherwise you're not really sure which one triggered it (if the formats can't be compared). But your point also makes sense; it's mostly here to report seriously wrong things, and it's not conveyed at post-processing time.

> Having a look, though, I noticed something I missed and is that the monotonic_offsets are detached from each other (e.g. initial event and printer). This doesn't seem right even if we're talking about ns, does it? One may argue that we may want to remove the initial event if we don't store, but still we can run --print when storing. I think that should be fixed and we can store it at `Collectors::new()` level once and for all (or alternatively, if we want to be fancy, we can lazily store it at `Collect` level with something like `#[skip_arg]`). The point is that those offsets should not be taken as we go, but once, or am I missing something?

The theoretical offset should be the same, both clocks are running forward. In practice as the offset computation relies on doing two syscalls one after the other a slightly different offset might be computed.  But also in practice, the UTC formatting won't go down that level so the time being printed should be the same at collect and post-processing time.

vlrpl: > > honestly, this is a debug feature and there's no requirement for it to be tied to user UX. That being said, we can do this, of course it would require BpfEventsFactory to store the monotonic_offset and consume it.
> 
> I was thinking being able to see when an error is thrown wrt to events might be wanted, otherwise you're not really sure which one triggered it (if the formats can't be compared). But your point also makes sense; it's mostly here to report seriously wrong things, and it's not conveyed at post-processing time.
> 
> > Having a look, though, I noticed something I missed and is that the monotonic_offsets are detached from each other (e.g. initial event and printer). This doesn't seem right even if we're talking about ns, does it? One may argue that we may want to remove the initial event if we don't store, but still we can run --print when storing. I think that should be fixed and we can store it at `Collectors::new()` level once and for all (or alternatively, if we want to be fancy, we can lazily store it at `Collect` level with something like `#[skip_arg]`). The point is that those offsets should not be taken as we go, but once, or am I missing something?
> 
> The theoretical offset should be the same, both clocks are running forward. In practice as the offset computation relies on doing two syscalls one after the other a slightly different offset might be computed. But also in practice, the UTC formatting won't go down that level so the time being printed should be the same at collect and post-processing time.

that was my point. I didn't look into UTC formatting precision, though, and if this is visible (if not, it's not a problem indeed), otherwise, with this, we'd end up having 3.

atenart: Yeah :) Not having a practical issue doesn't mean we can't make things more robust though.

vlrpl:
Note about the linter patch.
Although #[default] was stabilized in 1.62, that should be no issue for us, given the following:
https://documentation.ubuntu.com/ubuntu-for-developers/reference/availability/rust/

If we still have doubts about having that minimum version (assuming it's not already so), the patch must be turned in lint suppression as the CI is going to fail sooner or later.

retis-events/src/time.rs
```diff
@@ -79,6 +79,12 @@ impl From<TimeSpec> for i64 {
     }
 }
 
+pub fn format_date_time(timestamp: u64, offset: TimeSpec) -> String {
```
atenart: This is formatting the date in UTC. Either we should append `_utc` to the name or make the time zone a generic in the function. The later will require users to include `chrono` to access its `TimeZone` trait (which is fine IMO).

Another option could be to factorize the whole time formatting logic into a function taking the timestamp and the `TimeFormat` as an argument; which would avoid each user having to re-implement the logic.

retis/Cargo.toml
```diff
@@ -33,6 +33,7 @@ btf-rs = "1.1"
 byteorder = "1.5"
 caps = "0.5"
 cargo-platform = "0.3"
+chrono = "0.4"
```
atenart: This is not needed (unless the format function takes a timezone as a generic param).

retis/src/core/events/bpf.rs
```diff
@@ -254,7 +264,13 @@ impl BpfEventsFactory {
             };
 
             match raw_to_string!(&log_event.msg) {
-                Ok(msg) => log!(log_level, "[eBPF] {msg}"),
+                Ok(msg) => {
+                    let timestamp_str = match monotonic_offset {
+                        Some(offset) => format_date_time(log_event.ts, offset),
+                        None => log_event.ts.to_string(),
+                    };
+                    log!(log_level, "[eBPF] {} {msg}", timestamp_str);
```
atenart: nit: `{timestamp_str}`.

vlrpl:
> > Note about the linter patch. Although #[default] was stabilized in 1.62, that should be no issue for us, given the following: https://documentation.ubuntu.com/ubuntu-for-developers/reference/availability/rust/
> > If we still have doubts about having that minimum version (assuming it's not already so), the patch must be turned in lint suppression as the CI is going to fail sooner or later.
> 
> We require the latest version of Rust/Cargo so this should be fine?

that's right, I was considering non-CI builds, but from the lints point of view they don't matter much (if not at all).
Yes, we should be fine with #[default].

retis/src/core/events/bpf.rs
```diff
@@ -134,9 +136,16 @@ impl BpfEventsFactory {
             handle: None,
             log_handle: None,
             run_state: Running::new(),
+            monotonic_offset: None,
         })
     }
 
+    /// Configure logger timestamp formatting
+    pub(crate) fn config_logger(&mut self, utc: bool, monotonic_offset: TimeSpec) -> Result<()> {
```
atenart: nit: I don't think the `utc` check should be in the logger. Instead an `Option<TimeSpec>` can be given, which also allows disabling the feature if needed.

atenart (commented):
A few nits, otherwise looks good!

retis/Cargo.toml
vlrpl: it was initially open coded in retis before `format_date_time()`, then this remained.
I'll take a look at your generic param suggestion.

retis/src/core/events/bpf.rs
vlrpl: That's what happens already. I guess you mean to move the utc check one level up (`Collectors::config`), right?
`Result<()>` also can be dropped here (made sense in the previous patch, but keeping it doesn't add anything now).

atenart: > I guess you mean to move the utc check one level up

That's what I had in mind however I'm seeing an issue right now. Doing this would mean `TimeSpec` == UTC formatting. However this is not true. We're actually missing a conversion from `--utc` to a `TimeFormat` representation and feeding this into the logger; liek what we do for other consumers of that cli option.

retis-events/src/time.rs
vlrpl: Factoring everything passing timestamp + offset + formatter looks the best approach as we simply decouple the consumers from the TimeZone logic (and the formatting one), a thing that the function w/ generic doesn't "solve".

retis/src/core/events/bpf.rs
vlrpl: Makes sense, although for the logger, whose lifetime is all at collect time, there should be no logical difference. Still respecting the decoupling remains valid, e.g. if one day a new time formatter gets introduced.

PR #576, opened 2025-09-24
doc: contributing: remove issue links

atenart:
None of those two labels exist anymore.


atenart:
The unit test CI failure can be safely ignored, this PR is only modifying the documentation.

PR #577, opened 2025-09-26
Makefile: Instruct the linker to not gc the sections for the test target

vlrpl:
Other approaches are possible but all come with downsides:
- The officially suggested option is not supported by rustc < 1.90 
- .cargo/config.toml doesn't discriminate between different build profiles/types (e.g. test vs release)
- using gcc as linker in `.cargo/config.toml` seems to fix the issue, but pinning the linker doesn't seem the best approach


atenart (approved):
LGTM, thanks for the investigation and the fix!

PR #580, opened 2025-10-02
docs: add filtering blog post

vlrpl:
A new blog post is available, add it to the list.


vlrpl:
> The new link should be on top, they are ordered by reverse date.

whoops, didn't notice :)
"Make way for the youth!" :)

atenart (approved):
Thanks for the article!

PR #582, opened 2025-10-07
retis_in_container: fix mounting fs when using Docker and AppArmor

atenart:
When using Docker as the runtime on an AppArmor enabled system, if --allow-system-changes is enabled Retis can mount tracefs/debugfs. This will fail with EACCES as the default Docker AppArmor profile does not allow mount operations (as we run without --privileged).

Fix this by disabling the AppArmor profile when using Docker as the runtime, limiting the behavior to when --allow-system-changes is enabled.

Fixes: a6dba52316d4 ("retis_in_container: be less privileged")


tools/retis_in_container.sh
```diff
@@ -41,7 +41,7 @@ for kconfig in /proc/config.gz \
         kconfig_map="$kconfig_map -v ${kconfig}:${kconfig}:ro"
 	# Map the first item to /kconfig to support older versions of the container
 	# image.
-	[[ -z $kconfig_legacy_map ]] && kconfig_legacy_map="-v ${kconfig}:/kconfig"
+	[[ -z $kconfig_legacy_map ]] && kconfig_legacy_map="-v ${kconfig}:/kconfig:ro"
```
vlrpl, line 44: I guess in the commit message `explicitly setting their rw rights` should be "access rights" (or ro, not sure why rw).

Unrelated, what's the advantage of this check `[[ -z $kconfig_legacy_map ]]`?

atenart: > I guess in the commit message `explicitly setting their rw rights` should be "access rights" (or ro, not sure why rw).

Will do!

> Unrelated, what's the advantage of this check `[[ -z $kconfig_legacy_map ]]`?

The idea is to set it to the first valid mapping found, as they are defined in preferred order.

vlrpl: yeah, of course

PR #583, opened 2025-10-10
Next test infra

atenart:
As discussed offline, we have an interest to improve the testing infrastructure, especially for what we can "runtime tests". This is currently living in `tests/` and using `pytest`.

While `pytest` is well-known and provides good cleanup and testing capabilities, and while using Python to check test outputs and results provided solid results, it had disadvantages:

- Medium/high entry bar to write tests, especially as the whole setup and topology had to be done in Python.
- Low flexibility, for the same reasons.
- Reproducing errors and inspecting them locally was not the easiest.

This is an attempt at solving this. The main goals were:

- Providing good cleanup capabilities, whether or not a test succeeded.
- Allowing flexibility by allowing any command to be used (during test setup, run and processing).
- Lowering the entry bar by removing one abstraction layer (not requiring Python, especially during setup).
- While still benefiting from the versatility and power of Python, especially while checking test outputs.

**How it works — writing tests**

Cleanup relies on registering commands to be run if the test fails or if it completes. Cleanup is run between every tests or on failure. Every command failure is considered a test failure, unless manually handled (e.g. `expected_fail || true`).

The core library aims at providing common helpers to ease test writing and in case where cleanup is always required, to make it transparent. But it's not designed to replace each and every shell command, if they are simple enough. E.g. setting an ip address to a device in a netns can be done manually.

As before, common topologies are provided.

Tests are each ran in a dedicated temp dir. All test output is stored in `$tempdir/log` and displayed on failure (or accessible on inspection, see below).

Example:

```bash
#!/bin/bash
source include/lib.sh
source include/helpers.sh

test_ping() {
    two_ns_simple
    ip netns exec ns0 ping -c1 10.0.42.2
}

test_dummy() {
    daemonctl start
    cleanup "daemonctl stop"
    ....
}

run_tests test_ping test_dummy
```

For more complex testing needs, Python can be used (e.g. at post-processing). The preferred way is to use `retis python`, although calling Python code directly  would work too. Common helpers are provided in `include/helpers.py` and can be used.

```bash
test() {
    ...
    echo >test.py <<EOF
from helpers import *
...
EOF
    retis python test.py
}
```

The `assert_events_present` is converted already as part of this PR.

**How it works — using tests**

The goal will be to have a simple `make runtime-tests` target for running all tests in the CI or locally (not done yet).

But tests can also be run individually (from any directory). The specific sub-test to run can be specified too.

```bash
$ sudo ./tests/next/skb.sh
$ sudo TESTS=foo ./tests/next/skb.sh
```

A shell can be started on failure, within the workdir and with some capabilities exported (`$retis` to access the right binary, config for Python to import the `helpers`, etc). Upon closing the inspecting shell, the test is cleaned up.

```bash
$ sudo INSPECT_ERR=1 ./tests/next/skb.sh
```

**Next steps**

This is only a proposal. If we agree on a direction, we can have the two tests infra in parallel for a while:

- Add new tests to the new infra only.
- Converting existing tests over time (so it can be done as a common effort, to benefit from learning writing tests).

It's also a good opportunity to improve the `retis python` sub-command by using it more and relying on its behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ __pycache__/
 retis.data
 retis-events/.tox
 tools/wireshark/retis.so
+.vagrant


vlrpl:
> > * Range run. It can be nice to run group of tests. It's true tests are already split logically in files, but it would be nice to run subsets of different files as the split forces. Like `TESTS=04-10 ...`.
> 
> I'm not pushing back on this but I'm wondering how this could look like. For context there are three ways of running tests (first one is not part of the PR yet):
> 
>     * Running all tests, e.g. `make runtime-test`.
> 
>     * Running a set of tests, `./tests/ovs.sh`.
> 
>     * Running a specific test or collection of tests within a single tests set, `TESTS="test1 test2 test3" ./tests/skb.sh`. One improvement might be to be able to run a specific test w/o knowing the exact file where it is defined (and mixing tests), e.g. `TESTS=test0 make runtime-test`.
> 

Having `TESTS=test0 make runtime-test` would be nice, I agree.
Another thing, though, is listing the test names. That is pretty standard and would be useful either ways. It should not change things up much.

I'm proposing the range run because I'm an optimist and I expect we'll have loads of tests that will require a little flexibility to call a significant scattered subset without using lengthy invocation. 
OTOH, I hate to push in a direction that requires significant rework just because "it may be useful, but I don't really know". I prefer the "we need it, we do it" approach. That to say I'm sharing random idea here, more than having a strong opinion/preference about it.

> 
> Tests are not defined in a single location (which is handy for the 2nd point above). Also there is no current notion of test number and while we could skip the first n tests and then run the next x ones, this doesn't map to any user fronting information. With that I'm wondering if a number would be the correct way of expressing the range, as it doesn't really map to anything (for now)?
> 

In my mind it would have been something like:

```
tests/run_tests.sh --list
01 - file_basename - function_name
02 - file_basename - function_name
[...]
N - file_basename - function_name

TESTS="20-32 45-61" tests/run_tests.sh
```

I guess this would always require us to call `run_tests.sh` even for a single file execution:
`tests/run_tests.sh skb`

But as I said, I'm not sure this will give us a great gain (hence I hate to advocate for it :D) in terms of usage, especially if the whole test set is quick to run as I expect.

> But if we do not use numbers, how to express a range? A solution in between could be to allow wildcards, for running all tests matching an expression? E.g. `TESTS="*tcp*" make runtime-tests`.
> 
> To sum up I'm mainly wondering how to express this the best way. Thoughts?
> 
> > * AFAIU, we can inspect on error. This is great to jump in and deal with failures, but sometimes we simply want to stop the execution at some point and jump in (e.g. while developing a feature, or while writing a new test). Having something like that would be nice. I was initially thinking to something like `-bp` flag in rpmbuild, but might well be too much. It would be enough to have something as simple as some `BREAK` that stops the execution allowing one to work in the environment. I consider resuming the test once optional.
> 
> That's a great idea. It's actually possible already, by just calling `false`. But maybe having something a little more user friendly would be better, e.g. the proposed `BREAK`.

yep, that will again stop on error. I mean, "BREAK" can be an alias for `false`, or, if we find value in resuming once done, I guess `false` can be replaced by some kind of mechanism (should not be too bad to implement).

atenart:
> > > * Range run. It can be nice to run group of tests. It's true tests are already split logically in files, but it would be nice to run subsets of different files as the split forces. Like `TESTS=04-10 ...`.
> > 
> > 
> > I'm not pushing back on this but I'm wondering how this could look like. For context there are three ways of running tests (first one is not part of the PR yet):
> > ```
> > * Running all tests, e.g. `make runtime-test`.
> > 
> > * Running a set of tests, `./tests/ovs.sh`.
> > 
> > * Running a specific test or collection of tests within a single tests set, `TESTS="test1 test2 test3" ./tests/skb.sh`. One improvement might be to be able to run a specific test w/o knowing the exact file where it is defined (and mixing tests), e.g. `TESTS=test0 make runtime-test`.
> > ```
> 
> Having `TESTS=test0 make runtime-test` would be nice, I agree. 

:+1:

> Another thing, though, is listing the test names. That is pretty standard and would be useful either ways. It should not change things up much.

Good idea!

> I'm proposing the range run because I'm an optimist and I expect we'll have loads of tests that will require a little flexibility to call a significant scattered subset without using lengthy invocation. OTOH, I hate to push in a direction that requires significant rework just because "it may be useful, but I don't really know". I prefer the "we need it, we do it" approach. That to say I'm sharing random idea here, more than having a strong opinion/preference about it.
> 
> > Tests are not defined in a single location (which is handy for the 2nd point above). Also there is no current notion of test number and while we could skip the first n tests and then run the next x ones, this doesn't map to any user fronting information. With that I'm wondering if a number would be the correct way of expressing the range, as it doesn't really map to anything (for now)?
> 
> In my mind it would have been something like:
> 
> ```
> tests/run_tests.sh --list
> 01 - file_basename - function_name
> 02 - file_basename - function_name
> [...]
> N - file_basename - function_name
> 
> TESTS="20-32 45-61" tests/run_tests.sh
> ```
> 
> I guess this would always require us to call `run_tests.sh` even for a single file execution: `tests/run_tests.sh skb`
> 
> But as I said, I'm not sure this will give us a great gain (hence I hate to advocate for it :D) in terms of usage, especially if the whole test set is quick to run as I expect.

So maybe we can wait a bit to see if we miss that functionality and add it if we feel it's missing.

> > But if we do not use numbers, how to express a range? A solution in between could be to allow wildcards, for running all tests matching an expression? E.g. `TESTS="*tcp*" make runtime-tests`.
> > To sum up I'm mainly wondering how to express this the best way. Thoughts?
> > > * AFAIU, we can inspect on error. This is great to jump in and deal with failures, but sometimes we simply want to stop the execution at some point and jump in (e.g. while developing a feature, or while writing a new test). Having something like that would be nice. I was initially thinking to something like `-bp` flag in rpmbuild, but might well be too much. It would be enough to have something as simple as some `BREAK` that stops the execution allowing one to work in the environment. I consider resuming the test once optional.
> > 
> > 
> > That's a great idea. It's actually possible already, by just calling `false`. But maybe having something a little more user friendly would be better, e.g. the proposed `BREAK`.
> 
> yep, that will again stop on error. I mean, "BREAK" can be an alias for `false`, or, if we find value in resuming once done, I guess `false` can be replaced by some kind of mechanism (should not be too bad to implement).

:+1:

We can start by aliasing `BREAK` to false and if we find resuming would be needed we can add it then.

tests/next/include/lib.sh
```diff
@@ -0,0 +1,96 @@
...
+
+	# Start a shell to inspect failures if configured to.
+	[ "$INSPECT_ERR" = "1" ] && {
+		echo "Starting shell for inspection..."
+		PS1="inspect> " bash --norc
+	}
+
+	rm -f log
+}
+exec 21>&1; exec 22>&1
```
vlrpl: a small suggestion/nit. It would be nice to add a small comment here to explain the reason of the exec here. WDYT?

atenart: Yep. Also it seems there's a c/p error :)

Makefile
```diff
@@ -201,15 +201,7 @@ check-rust:
...
-	COMMIT_RANGE=$$(git rev-list --no-merges --reverse $${base_hash}..HEAD); \
-	for commit in $$COMMIT_RANGE; do \
-	    pretty_commit=$$(git -P show -s --format="%h (\"%s\")" --abbrev=12 $${commit}); \
-	    echo "Commit $${pretty_commit}"; \
-	    printf -- '-%0.s' $$(seq 1 $${#pretty_commit}); \
-	    echo "-------"; \
-	    git show --format=email $${commit} -- '*/bpf/*' | ./tools/checkpatch/checkpatch.pl --no-tree --ignore=SPDX_LICENSE_TAG,FILE_PATH_CHANGES,EMAIL_SUBJECT,VOLATILE -q || restyle=$$?; \
-	    echo; \
-	done; \
+	git diff $${commit} -- '*/bpf/*' | ./tools/checkpatch/checkpatch.pl --no-tree --ignore=SPDX_LICENSE_TAG,FILE_PATH_CHANGES,EMAIL_SUBJECT,VOLATILE -q || restyle=$$?; \
```
vlrpl: I'm missing where $${commit} is defined (I guess it's a typo and should be base_hash), but that aside, this is not the same thing.
The previous was intentionally implemented to show what patch has what issue. This way we just point the issue out.

atenart: > I'm missing where $${commit} is defined (I guess it's a typo and should be base_hash),

Good catch!

> but that aside, this is not the same thing. The previous was intentionally implemented to show what patch has what issue. This way we just point the issue out.

The change here is intended (and can be discuss of course), maybe the commit message could be more clear.

Having an unified output has the following advantages:
- It is aligned with all other format / lint tests we have; they all check the current state of the tree.
- It helps integrating the eBPF checks in wider targets and making a simple "check all" target.
- It takes into account uncommited changes (like other fmt/lint checks).

The drawback is (AFAICT):
- It's not checking intermediate states. But that's also what other fmt/lint checks do.

The biggest selling point for me being unification between check targets. WDYT?

vlrpl: > > I'm missing where $${commit} is defined (I guess it's a typo and should be base_hash),
> 
> Good catch!
> 
> > but that aside, this is not the same thing. The previous was intentionally implemented to show what patch has what issue. This way we just point the issue out.
> 
> The change here is intended (and can be discuss of course), maybe the commit message could be more clear.
> 
> Having an unified output has the following advantages:
> 
>     * It is aligned with all other format / lint tests we have; they all check the current state of the tree.
> 
>     * It helps integrating the eBPF checks in wider targets and making a simple "check all" target.
> 
>     * It takes into account uncommited changes (like other fmt/lint checks).
> 
> 
> The drawback is (AFAICT):
> 
>     * It's not checking intermediate states. But that's also what other fmt/lint checks do.
> 
> 
> The biggest selling point for me being unification between check targets (although I'm not sure how to integrate the eBPF check into the wider one w/o that change). WDYT?

Does it fail when used as prerequisite?

My remarks about the matter:

What I find extremely annoying (especially in big series) about rust check tooling is exactly the lack of the git-aware feature :) But this pairs with the lack of diff support (another bad thing for multiple reasons, IMO).
Most of the times, if `make check-ebpf` complains, I do:
```
# with all the series applied
# fix patch1
stg refresh -p patch1
# fix patch2
stg refresh -p patch2
[...]
```
or:
`stg edit --diff patch1`

With Rust tooling it's easier to fix the style issue in the wrong patch (I know, a bit of extra care avoids this :)

I prefer the per-commit version as it's way more convenient. If we really want to check against the uncommitted changes, we can add the `git diff | ...` flavour for the uncommitted changes.

atenart: > > The biggest selling point for me being unification between check targets (although I'm not sure how to integrate the eBPF check into the wider one w/o that change). WDYT?
> 
> Does it fail when used as prerequisite?

Not sure I got this. If checkpatch issue a warning or an error, it is displayed and if another target was depending on this (e.g. `checks`) it does fail.

> With Rust tooling it's easier to fix the style issue in the wrong patch (I know, a bit of extra care avoids this :)

You mean as an external configuration in your text editor?

> What I find extremely annoying (especially in big series) about rust check tooling is exactly the lack of the git-aware feature :) But this pairs with the lack of diff support (another bad thing for multiple reasons, IMO).

I'm not a fan of making similar targets behave differently as the expectations are then different; even if I understand it's easier to deal with the eBPF fmt issues in the current implementation (outside uncomitted ones) for you.

If I got the tooling comment right, this feels like relying on the implementation for some parts while having a custom one for others. IMO the checks should work standalone, if one wants to tweak things locally that's perfectly fine. (This, regardless of the default behavior, as long as it's consistent).

I understand breaking something that works for your use atm case is frustrating though... :(

> I prefer the per-commit version as it's way more convenient. If we really want to check against the uncommitted changes, we can add the `git diff | ...` flavour for the uncommitted changes.

Yep, and in addition capture the checkpatch output and only print it if warnings/errors are found.

Let's ask for a third opinion; @amorenoz WDYT?

vlrpl: > > > The biggest selling point for me being unification between check targets (although I'm not sure how to integrate the eBPF check into the wider one w/o that change). WDYT?
> > 
> > 
> > Does it fail when used as prerequisite?
> 
> Not sure I got this. If checkpatch issue a warning or an error, it is displayed and if another target was depending on this (e.g. `checks`) it does fail.
> 
> > With Rust tooling it's easier to fix the style issue in the wrong patch (I know, a bit of extra care avoids this :)
> 
> You mean as an external configuration in your text editor?
> 

no, no custom confs involved.
Let's say patch3 has a style issue, but patch 6 and 7 modify lines close to that, you have to rebase manually and check each patch to get where the issue is. Also, if patch6 moves patch3 code, you fix it in patch6 but you still have it in 3, and so on. 

> > What I find extremely annoying (especially in big series) about rust check tooling is exactly the lack of the git-aware feature :) But this pairs with the lack of diff support (another bad thing for multiple reasons, IMO).
> 
> I'm not a fan of making similar targets behave differently as the expectations are then different; even if I understand it's easier to deal with the eBPF fmt issues in the current implementation (outside uncomitted ones) for you.

They will still behave differently as `cargo fmt` works with files. For checkpatch we agreed to work with patches.

> 
> If I got the tooling comment right, this feels like relying on the implementation for some parts while having a custom one for others. IMO the checks should work standalone, if one wants to tweak things locally that's perfectly fine. (This, regardless of the default behavior, as long as it's consistent).
> 

All works standalone as of now. No custom things involved.
What I meant is that I prefer not to discard a better behavior and use a sub-optimal one only because the others only have the "sub-optimal". Being a dev thing, I struggle to think a contributor may get confused by it (I'm obviously talking about any casual contributor passing by, not the ones used to it).

> I understand breaking something that works for your use atm case is frustrating though... :(
> 

not just my use, logs are better and CI benefits from it

> > I prefer the per-commit version as it's way more convenient. If we really want to check against the uncommitted changes, we can add the `git diff | ...` flavour for the uncommitted changes.
> 
> Yep, and in addition capture the checkpatch output and only print it if warnings/errors are found.
> 
> Let's ask for a third opinion; @amorenoz WDYT?

atenart: You probably already know this but as a side note you can run any check target per commit using `git rebase main -x 'make check-<flavor>'`. I'm not sure how to do this with `stgit`, maybe a script is needed there :(

vlrpl: > You probably already know this but as a side note you can run any check target per commit using `git rebase main -x 'make check-<flavor>'`. I'm not sure how to do this with `stgit`, maybe a script is needed there :(

yep. 
Not that I always only use stgit :)  but I usually have `[stgit.alias]` with `!exec` for things like that.
Regardless of the tooling, the bad thing is that logs in CI should be considered "unreliable" (you always have to double check through the series). This will be extended to ebpf.

Makefile
```diff
@@ -192,6 +192,12 @@ endef
 
 $(foreach tgt,check clippy,$(eval $(call analyzer_tmpl,$(tgt))))
 
+check-rust:
+	$(call build, fmt --check, check format)
+	# No need to run `check` in addition to `clippy`, it's a superset.
+	$(call build, clippy --quiet, running clippy)
```
vlrpl: it's a pity having the target (theoretically both) and not being able to reuse them.
"clippy" does essentially the same and with `CARGO_CMD_OPTS=-F benchmark` should be enough for benchmarks as well.
Did you consider to unify them in some way?

atenart: We currently only have the `clippy` target supported. The issue is to set the right flags when calling these different commands (e.g. `cargo fmt --quiet` does not print anything on error), and it turned out reusing the targets ended up requiring complex logic and more lines. Those targets are also just a wrapper for handling `RA=1` which is not providing value nor wanted in `check-rust` I believe.

vlrpl: yeah, `RA` is not what we want.
I was hoping we could reuse them by simply using it as a prerequisite, but I agree, it's hairy (i.e. I don't have an quick solution to easily reuse it).
I'll try to take a second look at it.
For the time being, I noticed that the output of `make check-rust V=1` differs from `make V=1`. It seems this is because the clippy subcommand doesn't honor the `-q` option, but requires its own --quiet per-subcommand. Argh :)

atenart: > I'll try to take a second look at it.

Thanks!

> For the time being, I noticed that the output of `make check-rust V=1` differs from `make V=1`. It seems this is because the clippy subcommand doesn't honor the `-q` option, but requires its own --quiet per-subcommand. Argh :)

Ah that's not great, we should support V=1.

vlrpl (commented):
I will try to go through one more round of tests/reviews, but in general the PR looks good apart from the check-ebpf thingy :)
Some changes might be split in a different PR (e.g. some on Makefile) and could be merged straight away, but that's up to you.
Just a small heads up (most probably you already noticed), there are some conflicts.

atenart:
> Some changes might be split in a different PR (e.g. some on Makefile) and could be merged straight away, but that's up to you. Just a small heads up (most probably you already noticed), there are some conflicts.

I'll keep the all the patches in there, the Makefile changes are not urgent. I'll update once we settle on the checkpatch question.

tests/next/include/lib.sh
```diff
@@ -0,0 +1,128 @@
...
+# Allow importing custom Python modules.
+export PYTHONPATH="${PYTHONPATH}:$(git rev-parse --show-toplevel)/tests/next/include"
+
+# Create network namespaces. Take a list of network namespace names as the
+# input.
+add_ns() {
+	for ns in $@; do
+		ip netns add $ns
+		ip -n $ns link set lo up
+		cleanup "ip netns del $ns | xargs -r kill"
```
amorenoz: AFAICS "ip netns del" does not print any pid that can be then killed. Did you mean something like:
```
ip netns pids $ns | xargs -r kill
ip netns del $ns
```

tests/next/include/helpers.py
amorenoz, line 1: I wonder if this could be a python script that gets called by retis embedded python interpreter directly, .e.g: `retis python contains --args "expected.json`.

Makefile
```diff
@@ -206,6 +212,14 @@ check-ebpf:
 	done; \
 	[ -z "$$restyle" ]
 
+check-python:
+	$(call out_console,FLAKE8,checking Python coding style ...)
+	python3 -m flake8 tests
+	$(call out_console,BLACK,checking Python coding style ...)
+	python3 -m black --check $(if $(VERBOSITY),,--quiet) --diff tests/*.py tests/next/include/*.py
+
+checks: check-rust check-ebpf check-python
```
amorenoz: nit: "checks" is  a bit confusing for me. too similar to "check" with a completely different behavior. "lint"?

tests/next/include/lib.sh
```diff
@@ -0,0 +1,128 @@
...
+		# Do not error on unknown tests to allow `TESTS` use with
+		# `make functional-tests`.
+		typeset -f $t >/dev/null || continue
+		[ "$LIST_TESTS" == "1" ] && { echo $t; continue; }
+
+		tmpdir=$(mktemp -d -t retis-test-XXXX)
+		cleanup "rm -rf $tmpdir"
+		cd $tmpdir
+
+		echo -n "Running $t test..."
```
amorenoz, line 88: Have you considered printing stuff in a standard-ish format? e.g: tap or ktap?

Makefile
amorenoz: Just manged to carve out some time to look at this.
I haven't thought this very thoroughly but my first impression is that we should leave the more advanced output we currently have.
- Even if output is different from other "checks", this is for devs, not users, I think they'll understand what is printed. Besides, the current output is pretty clear, I don't think there is room for confusion.
- The current behavior gives you more information as it saves you one step (when I get a lining error I immediately `rebase -i main --exec "make ..."` to find the culprit).  The only information that is missing is the uncommitted changes but if we do the ` git diff | ...` trick that kind of covers everything.

tests/next/include/lib.sh
atenart: The number of times I made this error on my testing scripts... Thanks will fix!

tests/next/include/lib.sh
atenart: I have not but that's something we can do.

tests/next/include/helpers.py
atenart: This should work (there's an example of using `retis python` already in the skb sanity check). This was not done for checking a collectiong contains specific events since the events are directly read from the event file, using `retis python` would not add value here (and parse the events for not using them in their Python representation).

Makefile
atenart: Makes sense, will update.

Makefile
atenart: So the direction would be to keep the existing behavior and add a step for uncommited changes?

Just to make my point clear; that would make the ebpf check the only one behaving differently. The diff against `main` is just a shortcut IMO and I would even update this to diff against the empty tree (there's a special hash for that). If no other comment I'll follow the direction though :)

atenart: I went with reverting back to the old behavior while also adding support for uncommitted changes.

tests/next/include/lib.sh
amorenoz: I think tap is a fairly easy and flexible format. I was planning to output the result of the current functional tests in junit for cirrus to consume them directly and show each individual test pass/fail in the GH interface. Doing junit in bash might be an overkill but there might be tap-2-junit converters (have not looked).

atenart: As discussed offline, we'll consider using a known format for the output, in a follow-up.

tests/next/include/lib.sh
```diff
@@ -0,0 +1,128 @@
...
+	veth0_ns=$2	# Optional ("" to skip)
+	veth1_name=$3
+	veth1_ns=$4	# Optional ("" to skip)
+
+	[ ! -z "$veth0_ns" ] && ns0="netns $veth0_ns" || true
+	[ ! -z "$veth1_ns" ] && ns1="netns $veth1_ns" || true
+
+	ip link add $veth0_name $ns0 type veth peer name $veth1_name $ns1
+	ip -net ns0 link set $veth0_name up
+	ip -net ns1 link set $veth1_name up
```
amorenoz: netns names are hardcoded to "ns0" and "ns1".  We should use "$veth0_ns" and "$veth1_ns" if set or omot the "-net" option entirely.

tests/next/include/helpers.sh
```diff
@@ -0,0 +1,34 @@
...
+	ip -net ns0 address add 10.0.42.1/24 dev veth01
+	ip -net ns1 address add 10.0.42.2/24 dev veth10
+	ip -net ns0 address add 1111::1/64 dev veth01
+	ip -net ns1 address add 1111::2/64 dev veth10
+}
+
+# On top of two_ns_simple adds one VLAN interface per network namespace, as
+# follow:
+# ns0 - veth01.123 - 10.0.43.1/24 - 2222::1/64
+# ns1 - veth10.123 - 10.0.43.2/24 - 2222::2/64 - VLAN h/w offload off
```
amorenoz, line 19: Why the asymmetric VLAN h/w offload switching? We don't have any existing test that requires this. While I do see it as a good scenario for a test, I don't see this setup as being so common to justify it being in "helpers".

tests/next/include/lib.sh
atenart: Nice catch, I'll fix it!

tests/next/include/helpers.sh
atenart: We do, `two_ns_vlan` does this.

I'm not against making the default no use offload. On the other hand having this asymmetric configuration helps testing both VLANs flavors.

WDYT?

PR #584, opened 2025-10-17
Include the kernel version to the initial event

vlrpl:
`uname -rvm` was used, but `/proc/version` is another interesting one.

### DIFF
--- a/docs/events/core.md
+++ b/docs/events/core.md
@@ -45,10 +45,12 @@ nor disabled.
 
 ```none
 Retis version {version}
+Machine info {kernel_release} {kernel_version} {machine}
 ```
 
-Contains the `version` of the Retis binary that collected the event series as
-well as timing information for being able to display time in UTC at
-post-processing time. Only the `version` is shown in the output.
+Contains the `version` of the Retis binary that collected the event series, some
+machine-related information as in `uname -rvm` from the machine the data was
+collected, and timing information for being able to display time in UTC at
+post-processing time. The timing information is not displayed in the output.
 
 This section is emitted when a collection is started.


retis/src/collect/collect.rs
```diff
@@ -343,12 +343,19 @@ impl Collectors {
 
     // Generate an initial event with the startup section.
     fn initial_event(&mut self) -> Result<()> {
+        let uname =
+            uname().map_err(|e| anyhow!("Failed to get kernel version information: {e}"))?;
+        let release = uname.release().to_string_lossy();
+        let version = uname.version().to_string_lossy();
+        let machine = uname.machine().to_string_lossy();
```
atenart: Not all those fields are strictly speaking the kernel version, e.g. `machine`. They could be used independently at post-processing, e.g. checking the arch. Maybe this should be stored in 3 dedicated fields instead?

Maybe that's an opportunity to move this information to a sub-struct about the machine in general and not only the kernel?

(The display format is fine though).

retis-events/src/compat/compat.rs
```diff
@@ -114,10 +122,7 @@ pub(crate) fn compatibility_fixup(
     let fixups = FIXUPS
         .iter()
         .enumerate()
-        .filter_map(|(i, fixup)| match i {
-            x if x >= boundary => Some(fixup),
-            _ => None,
-        })
+        .filter_map(|(i, fixup)| if i > boundary { Some(fixup) } else { None })
```
atenart, line 136: This is changing the `boundary` check. Is that a fix (could be moved to its own commit) or a c/p error?

retis-events/src/compat/compat.rs
```diff
@@ -44,16 +49,19 @@ pub(crate) enum CompatVersion {
     V0 = 0,
     /* v1.6.x.. */
```
atenart: Please remove the trailing "..".

retis-events/src/common.rs
```diff
@@ -10,11 +10,21 @@ pub struct StartupEvent {
...
 
 impl EventFmt for StartupEvent {
-    fn event_fmt(&self, f: &mut Formatter, _: &DisplayFormat) -> fmt::Result {
-        write!(f, "Retis version {}", self.retis_version)
+    fn event_fmt(&self, f: &mut Formatter, d: &DisplayFormat) -> fmt::Result {
+        let sep = if d.multiline { "\n" } else { " " };
+
+        write!(f, "Retis version {}", self.retis_version)?;
+
+        if !self.kernel_version.is_empty() {
```
atenart: I don't think we should special case the string being empty. It's expected to be filled with a value, except for the backward compatibility output which could use "unknown" instead.

retis-events/src/compat/compat.rs
vlrpl: I don't follow, isn't this a separate commit already?

retis-events/src/common.rs
vlrpl: Uhm, I initially set it to "Unknown", but that require to change to "lazy" the FIXUPS const as we cannot initialize Strings with heap allocate values. I can do that quickly.
In general, we skip printing for other fields, doesn't sound like a special case (of course, one can argue that most of the times those are optional), at least from the user point of view.

retis-events/src/compat/compat.rs
vlrpl: ok, so the one for `v1.6.x..` was not intentional :)

retis-events/src/compat/compat.rs
atenart: Oops that right, I'm sorry, I'll have another coffee... I forgot I was reviewing in the unified diff tab...

vlrpl: > Oops that right, I'm sorry, I'll have another coffee... I forgot I was reviewing in the unified diff tab...

no problem at all.
While at it, I'll take it as a suggestion and make some coffee as well :)

retis-events/src/common.rs
atenart: OK, that works too. One of the thing making me think we should not special case this is except for the compatibility stuff, empty versions would end up being two empty spaces. But with the other comments, maybe that will be fixed.

retis-events/src/compat/compat.rs
atenart: It was, we could make it `v1.y.0..HEAD` if that's more clear. The intent was also so modify the comment to the right range when a new version is added. Here only a single y-stream version is contained, to `1.6.x` does the job. Maybe there are better ways to express this though? :)

vlrpl: no, TBH, it's clear. I guess it was just me glancing too quickly at your comment :)

retis/src/collect/collect.rs
vlrpl: You are suggesting something like this, I suppose.
But what about `CompatValue`? I don't think we're going to have many event sections/types (at least not at once) such that this becomes inconvenient, but I still wonder if this was what you had in mind in the first place in that todo comment.

The transition of FIXUPS to Lazy is optional in this case.

```diff
diff --git a/retis-events/src/common.rs b/retis-events/src/common.rs
index 39c354d2c..d96a45c90 100644
--- a/retis-events/src/common.rs
+++ b/retis-events/src/common.rs
@@ -2,6 +2,14 @@ use std::fmt;
 
 use crate::*;
 
+#[event_type]
+#[derive(Default)]
+pub struct MachineInfo {
+    pub kernel_release: String,
+    pub kernel_version: String,
+    pub machine: String,
+}
+
 /// Startup event section. Contains global information about a collection as a
 /// whole, with data gathered at collection startup time.
 #[event_section]
@@ -10,8 +18,8 @@ pub struct StartupEvent {
     pub retis_version: String,
     /// CLOCK_MONOTONIC offset in regards to local machine time.
     pub clock_monotonic_offset: TimeSpec,
-    /// Kernel release version while collecting events.
-    pub kernel_version: String,
+    /// Machine information retrieved while collecting events.
+    pub machine_info: MachineInfo,
 }
 
 impl EventFmt for StartupEvent {
@@ -20,9 +28,13 @@ impl EventFmt for StartupEvent {
 
         write!(f, "Retis version {}", self.retis_version)?;
 
-        if !self.kernel_version.is_empty() {
-            write!(f, "{sep}Kernel version {}", self.kernel_version)?;
-        }
+        write!(
+            f,
+            "{sep}Machine info {} {} {}",
+            self.machine_info.kernel_release,
+            self.machine_info.kernel_version,
+            self.machine_info.machine
+        )?;
 
         Ok(())
     }
diff --git a/retis-events/src/compat/compat.rs b/retis-events/src/compat/compat.rs
index 226602c41..0bfa96354 100644
--- a/retis-events/src/compat/compat.rs
+++ b/retis-events/src/compat/compat.rs
@@ -1,29 +1,35 @@
 #![allow(dead_code)] // Not everything is used in all versions.
 
 use anyhow::{anyhow, bail, Result};
+use once_cell::sync::Lazy;
 use semver::{Version, VersionReq};
 
+use crate::MachineInfo;
+
 /// When making breaking changes in the event, fixups should be added.
 use CompatFixup::*;
-const FIXUPS: &[&[CompatFixup]] = &[
-    /* CompatVersion::V0 */
-    &[],
-    /* CompatVersion::V1 */
-    &[
-        Add("ct/ct_status", CompatValue::Uint(0)),
-        Move("skb/packet/packet", "skb/packet/raw"),
-        Move("skb/ns/netns", "skb/ns/inum"),
-        Move("skb/packet", "packet"),
-        Move("skb/dev", "dev"),
-        Move("skb/ns", "netns"),
-        Move("packet/raw", "packet/data"),
-    ],
-    /* CompatVersion::V2 */
-    &[Add(
-        "startup/kernel_version",
-        CompatValue::String(String::new()),
-    )],
-];
+
+static FIXUPS: Lazy<Vec<Vec<CompatFixup<'static>>>> = Lazy::new(|| {
+    vec![
+        /* CompatVersion::V0 */
+        vec![],
+        /* CompatVersion::V1 */
+        vec![
+            Add("ct/ct_status", CompatValue::Uint(0)),
+            Move("skb/packet/packet", "skb/packet/raw"),
+            Move("skb/ns/netns", "skb/ns/inum"),
+            Move("skb/packet", "packet"),
+            Move("skb/dev", "dev"),
+            Move("skb/ns", "netns"),
+            Move("packet/raw", "packet/data"),
+        ],
+        /* CompatVersion::V2 */
+        vec![Add(
+            "startup/machine_info",
+            CompatValue::MachineInfo(MachineInfo::default()),
+        )],
+    ]
+});
 
 enum CompatFixup<'a> {
     Remove(&'a str),
@@ -47,7 +53,7 @@ pub(crate) enum CompatStrategy {
 pub(crate) enum CompatVersion {
     /* v1.5.x; we start the backward compatibility effort with v1.5.0 */
     V0 = 0,
-    /* v1.6.x.. */
+    /* v1.6.x */
     V1,
     /* v1.7.x.. */
     V2,
@@ -159,6 +165,7 @@ pub(crate) enum CompatValue {
     Int(i64),
     Uint(u64),
     String(String),
+    MachineInfo(MachineInfo),
     // TODO: event sections?
 }
 
diff --git a/retis-events/src/compat/json.rs b/retis-events/src/compat/json.rs
index 6fa25ef2c..8361adc19 100644
--- a/retis-events/src/compat/json.rs
+++ b/retis-events/src/compat/json.rs
@@ -69,6 +69,7 @@ impl EventCompatibility for serde_json::Value {
                     .ok_or_else(|| anyhow!("Failed to convert {val} to a serde_json Number"))?,
             ),
             CompatValue::String(val) => serde_json::Value::String(val),
+            CompatValue::MachineInfo(val) => serde_json::to_value(val)?,
         };
 
         match target {
diff --git a/retis/src/collect/collect.rs b/retis/src/collect/collect.rs
index 1f2fa7bff..a0810e7e2 100644
--- a/retis/src/collect/collect.rs
+++ b/retis/src/collect/collect.rs
@@ -355,7 +355,11 @@ impl Collectors {
                     .unwrap_or("unspec")
                     .to_string(),
                 clock_monotonic_offset: self.monotonic_offset,
-                kernel_version: format!("{release} {version} {machine}"),
+                machine_info: MachineInfo {
+                    kernel_release: release.to_string(),
+                    kernel_version: version.to_string(),
+                    machine: machine.to_string(),
+                },
             });
             Ok(())
         })

```

atenart: > You are suggesting something like this, I suppose. But what about `CompatValue`? I don't think we're going to have many event sections/types (at least not at once) such that this becomes inconvenient, but I still wonder if this was what you had in mind in the first place in that todo comment.

Initially yes but I don't think that's necessary. Adding support for types in `CompatValue` is for leaf types. Other intermediate ones can be represented in an intermediate generic type (`Value::Object(Map)` for json, dict for python). So maybe we should just add support for something like `CompatValue::Section`.

```rust
/* CompatVersion::V2 */
vec![
    Add("startup/machine_info", CompatValue::Section),
    Add("startup/machine_info/kernel_release", CompatValue::String("unknown")),
    Add("startup/machine_info/kernel_version", CompatValue::String("unknown")),
    Add("startup/machine_info/machine", CompatValue::String("unknown")),
],
```

It's a bit verbose in the fixup definitions, but simplicity is nice too.

WDYT?

vlrpl: Sounds a little hard believe now :) but this was what I thought in the first place.
I was too afraid about the verbosity (also I had some doubts this was the intention given the explicit reference to event sections). I guess, even though it is verbose, in practice it should not end up being that much.
I'm personally fine with it, it's probably nice to be descriptive here.
Also, I'm wondering if it's better at all even in terms of verbosity, as in case we set an `event_type` for `vX.(Y-2)`, but we then change the layout of that `event_type` in `vX.Y`, we have to carry the compat version of `vX.(Y-2)`, and doesn't really sound fun :)

retis-events/src/compat/compat.rs
```diff
@@ -1,24 +1,44 @@
...
-        Move("skb/packet/packet", "skb/packet/raw"),
-        Move("skb/ns/netns", "skb/ns/inum"),
-        Move("skb/packet", "packet"),
-        Move("skb/dev", "dev"),
-        Move("skb/ns", "netns"),
-        Move("packet/raw", "packet/data"),
-    ],
-];
+
+static FIXUPS: Lazy<Vec<Vec<CompatFixup<'static>>>> = Lazy::new(|| {
```
atenart: Have you considered making `CompatValue::String` hold a `&'static str` instead? TBF I'm not sure if this would be better.

(Although the `Lazy` construction might be useful for specific leaf types in the future).

vlrpl: I considered it, but admittedly I skipped it for no particular reason, I essentially kept the String thinking that maybe other type could benefit from it (but TBH, as of now, I can't think of any type other than String).
There's no big difference for our current use, but I guess we can do that and convert to Lazy if we really need it. WDYT?

atenart: Yeah, both work for me. Strictly speaking it's probably better to introduce such feature when we really need it, but I'm not against having it already. Your call :)

vlrpl: ok, in such case let's keep what we use for the time being and go with `&'static str`

retis/src/collect/collect.rs
```diff
@@ -343,12 +343,23 @@ impl Collectors {
 
     // Generate an initial event with the startup section.
     fn initial_event(&mut self) -> Result<()> {
+        let uname =
+            uname().map_err(|e| anyhow!("Failed to get kernel version information: {e}"))?;
```
atenart: nit: s/kernel version/system/

retis-events/src/common.rs
```diff
@@ -2,6 +2,13 @@ use std::fmt;
 
 use crate::*;
 
+#[event_type]
+pub struct MachineInfo {
+    pub kernel_release: String,
+    pub kernel_version: String,
+    pub machine: String,
```
atenart: As per the `uname` man page this is the h/w name. IMO it would be less confusing to name it this way, also avoiding `machine_info.machine`. WDYT?

retis-events/src/common.rs
```diff
@@ -10,11 +17,25 @@ pub struct StartupEvent {
     pub retis_version: String,
     /// CLOCK_MONOTONIC offset in regards to local machine time.
     pub clock_monotonic_offset: TimeSpec,
+    /// Machine information retrieved while collecting events.
+    pub machine_info: MachineInfo,
```
atenart: nit: I'm wondering if just calling it `machine` would be better? Most things in the events are "information" and this sould make using this easier, `event.machine.<>`.

retis-events/src/common.rs
vlrpl: the name comes from uname(2) man page :)
```
           struct utsname {
               [...]
               char machine[];    /* Hardware type identifier */
               [...]
           };
```
but I'm fine to rename it to hw_name. At this point, maybe hw_type? (a little less vague than hw_name(?))

atenart: Yeah the naming of this option in the interface and in uname options and docs is not the best; the above comment says "hardware type", the man page "hardware name", the command option "machine" and we might find other variants :sweat_smile: To me it even could be name "arch" or something. I guess there's no perfect choice here, so all mentioned options (except "machine" IMO) are good, your call :)

vlrpl: let's go with your suggestion, then

PR #591, opened 2025-11-28
Support parsing and printing AH and ESP packets

atenart:
This improves debugging IPsec setups. ESP support is only partial as the payload is encrypted.

### DIFF
--- a/retis-events/src/packet.rs
+++ b/retis-events/src/packet.rs
@@ -206,7 +206,7 @@ impl RawPacket {
                     None => Err(PacketFmtError::Truncated),
                 }
             }
-            _ => self.format_l3(f, format, etype, payload),
+            _ => self.format_etype(f, format, etype, payload),
         }
     }
 
@@ -232,7 +232,7 @@ impl RawPacket {
         Ok(())
     }
 
-    fn format_l3(
+    fn format_etype(
         &self,
         f: &mut Formatter,
         format: &DisplayFormat,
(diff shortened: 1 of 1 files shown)


retis-events/src/packet.rs
```diff
@@ -179,13 +179,7 @@ impl RawPacket {
...
-
-        if format.print_ll {
-            write!(f, " ")?;
-        }
-
-        self.format_l3(f, format, etype, payload)
+        self.format_l3(f, format, eth.get_ethertype(), eth.payload())
     }
 
     fn format_l3(
```
vlrpl: just a nit, is it still fair to suffix this with `_l3`?
I understand ethernet has its own dedicated format helper, but isn't `_l2l3` a little more fitting here?

atenart: Because of the VLAN traversal being in `format_l3` now?

atenart: I reworked the VLAN handling (in a dedicated commit), which makes things way better. The logic is more aligned with the other helpers, other protocols can chose to go back to Ethernet parsing or the VLAN directly depending on the needs, and even the extra space before l3 formatting is more logical now.

atenart: > Because of the VLAN traversal being in `format_l3` now?

Or is that because it handles things like ARP or MACsec? One way to make things less confusing would be to:
- s/format_l3/format_etype/
- s/format_l4/format_protocol/

vlrpl: some parts were arguably L2-only, others were kind of 2.5 :)  so I was wondering if we can be more descriptive in the naming.
I personally don't find it confusing as it's clear looking at the code, but it doesn't completely match the expectations.
Your proposal works for me, AFAIC.

PR #592, opened 2025-11-28
collect: add command line to the startup event

vlrpl:
RFC only to double-check for potential objections.
This is normally useful when:

-  the capture is performed by a third party
-  when captures are revisited after a long time

### DIFF
--- a/retis-events/src/common.rs
+++ b/retis-events/src/common.rs
@@ -15,6 +15,8 @@ pub struct MachineInfo {
 pub struct StartupEvent {
     /// Retis version used while collecting events.
     pub retis_version: String,
+    /// Expanded command line used to invoke the application.
+    pub cmdline: String,
     /// CLOCK_MONOTONIC offset in regards to local machine time.
     pub clock_monotonic_offset: TimeSpec,
     /// Machine information retrieved while collecting events.
@@ -26,6 +28,8 @@ impl EventFmt for StartupEvent {
         let sep = if d.multiline { "\n" } else { " " };
 
         write!(f, "Retis version {}", self.retis_version)?;
+
+        write!(f, "{sep}Command line {}", self.cmdline)?;
 
         write!(
             f,
--- a/retis/src/cli/cli.rs
+++ b/retis/src/cli/cli.rs
@@ -172,6 +172,9 @@ pub(crate) struct MainConfig {
         help = "Path to an additional directory with custom profiles. Takes precedence over built-in profiles directories."
     )]
     pub(crate) extra_profiles_dir: Option<PathBuf>,
+    /// Expanded command line that was used to invoke retis
+    #[arg(skip)]
+    pub(crate) cmdline: String,
 }
 
 const HELP_TEMPLATE: &str = "\
@@ -338,13 +341,13 @@ impl RetisCli {
         RetisCli::enhance_profile(&main_config, subcommand.name().as_str(), &mut args)
             .map_err(|err| command.error(ErrorKind::InvalidValue, format!("{err}")))?;
 
-        debug!(
-            "Resulting CLI arguments: {}",
-            args.iter()
-                .map(|o| o.as_os_str().to_str().unwrap_or("<encoding error>"))
-                .collect::<Vec<&str>>()
-                .join(" ")
-        );
+        let cmdline = args
+            .iter()
+            .map(|o| o.as_os_str().to_str().unwrap_or("<encoding error>"))
+            .collect::<Vec<&str>>()
+            .join(" ");
+
+        debug!("Resulting CLI arguments: {cmdline}");
 
         // Final round of parsing.
         let matches = match cfg!(test) {
@@ -364,6 +367,9 @@ impl RetisCli {
                 .update_from_arg_matches(matches)
                 .unwrap_or_else(|e| e.exit()),
         }
+
+        // Store the command line for use in initial event
+        main_config.cmdline = cmdline;
 
         Ok(CliConfig {
             command,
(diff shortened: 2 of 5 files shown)


retis/src/collect/collect.rs
```diff
@@ -348,11 +349,16 @@ impl Collectors {
         let version = uname.version().to_string_lossy();
         let machine = uname.machine().to_string_lossy();
 
+        let command_line: String = env::args_os()
+            .map(|arg| arg.to_string_lossy().into_owned())
+            .collect();
```
atenart: If a profile is used we might not know what's inside exactly, if it was modified or if it's a custom one. We do have however a `debug!` statement in our cli module that display the resulting arguments, after profile processing. Should we use it instead?

Having said that I'm also wondering if what we get from `env` has some extra information than what we can get from `clap` (e.g. env variables) and if that matters?

retis-events/src/common.rs
```diff
@@ -27,6 +29,8 @@ impl EventFmt for StartupEvent {
 
         write!(f, "Retis version {}", self.retis_version)?;
 
+        write!(f, "{sep}Command line {}", self.command_line)?;
```
atenart: This might be a long line, especially if we use the result command from `clap` after profile processing (see the other comment). Is that an issue? Should we have a verbose flag?

retis-events/src/common.rs
```diff
@@ -15,6 +15,8 @@ pub struct MachineInfo {
 pub struct StartupEvent {
     /// Retis version used while collecting events.
     pub retis_version: String,
+    /// Command line used to invoke the application.
+    pub command_line: String,
```
atenart: super nit: we could call it `cmdline` as it's the common name in `/proc` for similar things? Feel free to ignore this :)

retis/src/collect/collect.rs
vlrpl: > If a profile is used we might not know what's inside exactly, if it was modified or if it's a custom one. We do have however a `debug!` statement in our cli module that display the resulting arguments, after profile processing. Should we use it instead?
> 

I looked at that before before opening the discussion, but I was considering to keep profiles out. OTOH, and to be fair, the profiles inclusion is a good point, though.

I guess one way to store that is by the means of the following in `MainConfig`:
```
#[arg(skip)]
pub(crate) cmdline: String,
```

Which I think logically makes sense, otherwise, passing it around from `parse_from`  doesn't sound very nice. If you had something different in mind feel free to point it out.

> Having said that I'm also wondering if what we get from `env` has some extra information than what we can get from `clap` (e.g. env variables) and if that matters?

nothing comes to mind, but can't we always read the `env` and include it if it ispotentially missing?

atenart: > If you had something different in mind feel free to point it out.

I was mostly asking questions :) Not sure how practical that is.

retis-events/src/common.rs
vlrpl: It's true this can become fairly long, but it should not be the common case, I suppose. 
I guess we can add the option. The bad thing about it is that this would be the only consumer and of course, we would have an option just to avoid the truncated line.
Also I was wondering how `--verbose` might appear WRT `--log-level ...`. They are unrelated, but they kind of seem related :)

atenart: IMO `--log-level` and `--verbose` are quite different. The former is for "status" messages about how the processing is going on, while the latter is about the purpose of a command and its actual output. The former is implemented at the top level while the latter would be only for a subset of sub-commands.

Having said that I'm not pushing one way or the other wrt if we need to introduce `--verbose` or not.

vlrpl: > IMO `--log-level` and `--verbose` are quite different. The former is for "status" messages from the tool itself about how the processing is going on, while the latter is about the purpose of a command and its actual output.
> 

of course they are. I was considering how they would appear to the user (but that's a nit, honestly).

> Having said that I'm not pushing one way or the other wrt if we need to introduce `--verbose` or not.

sure, and same on my side. It's perfectly legitimate to ask whether it may make sense or not, I was just trying to think about upsides and downsides, if any :)

atenart: We can always go down the safe path, not have a verbose flag right now and add it later if we see the cmdline print is too verbose in real cases.

retis-events/src/common.rs
```diff
@@ -15,6 +15,8 @@ pub struct MachineInfo {
 pub struct StartupEvent {
     /// Retis version used while collecting events.
     pub retis_version: String,
+    /// Command line used to invoke the application.
```
atenart: nit: + "expanded"

retis/src/cli/cli.rs
```diff
@@ -172,6 +172,9 @@ pub(crate) struct MainConfig {
         help = "Path to an additional directory with custom profiles. Takes precedence over built-in profiles directories."
     )]
     pub(crate) extra_profiles_dir: Option<PathBuf>,
+    /// Command line that was used to invoke retis
```
atenart: nit: + "expanded"

retis/src/cli/cli.rs
```diff
@@ -338,13 +341,13 @@ impl RetisCli {
...
-                .collect::<Vec<&str>>()
-                .join(" ")
-        );
+        let cmdline = args
+            .iter()
+            .map(|o| o.as_os_str().to_str().unwrap_or("<encoding error>"))
+            .collect::<Vec<&str>>()
+            .join(" ");
+
+        debug!("Resulting CLI arguments: {}", cmdline);
```
atenart: `{cmdline}`

I'm wondering why Clippy did not flag this.

atenart (approved):
LGTM, thanks!

Issue #599, opened 2026-02-09
ARM64: tracepoint events silently dropped due to ELF mapping symbols in kallsyms


takaotsutomu:
On aarch64 (Ubuntu 24.04, kernel 6.8.0-90), tracepoint events from kernel modules are silently dropped. Tested on v1.6.3 and current main (aa241456) — the affected code has not changed across versions.

To reproduce: load the openvswitch module on an ARM64 host, run `sudo retis collect -c ovs,skb -f icmp`, and send ICMP traffic through an OVS bridge. Events with symbol `openvswitch:ovs_do_execute_action` are missing. Stderr shows:

```
ERROR Could not parse raw event: Factory 2 failed to parse section: Symbol $d does not exist or isn't traceable
```

ARM64 ELF mapping symbols (`$d`, `$x`, etc.) share addresses with real symbols in `/proc/kallsyms`:

```
ffff80007bfb6d60 d $d                                 [openvswitch]
ffff80007bfb6d60 d __tracepoint_ovs_do_execute_action  [openvswitch]
```

The kallsyms parser stores symbols in a bimap keyed by address, so `$d` overwrites the real symbol. `Symbol::from_addr()` then resolves the address to `$d` and fails the traceable-symbol check.

x86_64 is unaffected (no mapping symbols in ELF). The bug can affect any module tracepoint on ARM64, not just OVS.

PR #600, opened 2026-02-09
core: inspect: skip ARM64 ELF mapping symbols in kallsyms

takaotsutomu:
On ARM64, ELF mapping symbols (`$d`, `$x`, `$t`, `$a`) in `/proc/kallsyms` can share addresses with real kernel symbols and shadow them in the bimap, causing events to be silently dropped. Filter them out during parsing.

Fixes: #599

### DIFF
--- a/retis/src/core/inspect/kernel.rs
+++ b/retis/src/core/inspect/kernel.rs
@@ -473,6 +473,25 @@ mod tests {
     }
 
     #[test]
+    fn module_mapping_symbols_filtered() {
+        // The test data must contain a $d mapping symbol at a module
+        // symbol address. If test_data/kallsyms gets regenerated
+        // without it, this assert will catch it.
+        let kallsyms =
+            std::fs::read_to_string("test_data/kallsyms").expect("test_data/kallsyms readable");
+        assert!(
+            kallsyms.contains("ffffffffc0a25120 d $d\t[openvswitch]"),
+            "test_data/kallsyms must contain a $d mapping symbol \
+             at ffffffffc0a25120",
+        );
+
+        assert_eq!(
+            inspector().get_symbol_name(0xffffffffc0a25120).unwrap(),
+            "__tracepoint_ovs_do_execute_action",
+        );
+    }
+
+    #[test]
     fn kernel_modules() {
         assert_eq!(inspector().is_module_loaded("zram"), Some(true));
         assert_eq!(inspector().is_module_loaded("openvswitch"), Some(false));


retis/src/core/inspect/kernel.rs
```diff
@@ -84,6 +84,16 @@ impl KernelInspector {
                 .next()
                 .ok_or_else(|| anyhow!("Couldn't get symbol name for {}", data[0]))?;
 
+            // ARM64 ELF mapping symbols can share addresses with real
+            // symbols in kallsyms, shadowing them in the bimap.
```
atenart: nit: Could you add "since Linux commit ff09f6fd2972", to make understanding the context easier in a few years.

retis/test_data/kallsyms
```diff
@@ -96217,6 +96217,7 @@ ffffffff99d1d9d0 T __kfree_skb
 ffffffff99d1da10 T __pfx_kfree_skb_partial
 ffffffff99d1da20 T kfree_skb_partial
 ffffffff99d1da70 T __pfx_consume_skb
+ffffffff99d1da80 d $d
```
atenart: Adding a test is great (thanks!) however I'm concerned future updates of that test file will just remove the added line.

We can't only rely on reviewers and there's no easy way to create an `Inspector` using custom kallsyms data from a buffer. Maybe the best way would be to add another assert (+ comment) in the `arm64_mapping_symbols_filtered` test to make sure `$d` is found in `kallsyms`? This way we'll notice if an update does not have such problematic lines.

atenart (changes requested):
Thanks for the bug report and pull request! This makes sense and I only had minor comments.

(Also don't worry about the Python lint test failing)

retis/src/core/inspect/kernel.rs
takaotsutomu: Thanks for the pointer. I looked at the commit and it seems to affect the kernel's internal symbol lookup but not `/proc/kallsyms` output itself. Am I reading this right, or is there more context I'm missing? I'll update the comment once I understand the right framing.

retis/test_data/kallsyms
takaotsutomu: That's a good point. I agree with you. I added an assert that checks if the test file has the `$d` line to the existing `symbol_name` test so we'll notice, and removed the separate `arm64_mapping_symbols_filtered` test to avoid duplication.

retis/src/core/inspect/kernel.rs
atenart: My limited understanding of that code is that such symbols are already excluded from the core ones (see `e0109042cc4e`, `a2833d1b07ab` and `scripts/mksysmap`). That would align with the issue being see on an OvS symbol.

If we want to replicate the issue 1:1 the test case could be updated to use an OvS symbol too, instead of `consume_skb`. I guess the entry would be `ffffffffc0a25120 d $d   [openvswitch]` to conflict with `openvswitch:ovs_do_execute_action`.

takaotsutomu: Thanks for the explanation and the pointers to `e0109042cc4e`, `a2833d1b07ab`, and `scripts/mksysmap`. I was able to dig into the details and it all makes sense now. `mksysmap` already strips mapping symbols from vmlinux at build time, so the issue only affects module symbols in `/proc/kallsyms`.

I've moved the `$d` test entry to `ffffffffc0a25120` to conflict with `openvswitch:ovs_do_execute_action`, matching the real scenario. The test is now `module_mapping_symbols_filtered` to make the module context clear. I also added the `ff09f6fd2972` reference to the code comment and simplified the filter to `symbol.starts_with('$')` to match the kernel's own `is_mapping_symbol()`.

atenart: Thanks!

PR #601, opened 2026-02-11
Fix binding generation

vlrpl:
Binding generation is currently not working because a newly introduced prerequisite is missing in the gen-bindings target since 89885ae48079 ("Makefile: set ebpf variables on demand")
The patch restores the working behavior.


vlrpl:
> LGTM, thanks for the fix (and the lint commit message 😆)!

I didn't repush the version including the linter message /o\ 
Fortunately, it's simple enough to not require more than the current commit message 😆

PR #602, opened 2026-02-16
Misc improvements

atenart:


retis/src/profiles/cli.rs
```diff
@@ -66,7 +66,11 @@ impl ProfileCmd {
                     }
                 }
                 Err(err) => {
-                    warn!("Skipping invalid file {}: {err}", entry.path().display())
+                    let entry_path = entry.path();
+
+                    if entry_path.ends_with(".yaml") || entry_path.ends_with(".yml") {
```
amorenoz: Why not skip those files altogether before even trying to parse them?

retis/src/collect/collector/nft/nft.rs
```diff
@@ -33,7 +33,7 @@ pub(crate) struct NftCollectorArgs {
         long,
         value_parser=PossibleValuesParser::new(["all", "continue", "break", "jump", "goto", "return", "drop", "accept", "stolen", "queue", "repeat"]),
         value_delimiter=',',
-        default_value="drop,accept",
+        default_value="drop",
```
amorenoz: Is this because of some user report? I'm afraid changing this default behavior might make users think packets are not traversing nft when they are.

docs/events/packet.md
```diff
@@ -89,3 +89,25 @@ vxlan [{flags}] vni {vni}
...
+
+```none
+spi {spi} seq {sequence number}
+```
+
+## AH
+
+```none
+spi {spi} seq {sequence number} icv {icv}
+```
```
amorenoz, line 129: Maybe put these two last sections under an "## IPSec" section?

tools/retis_in_container.sh
```diff
@@ -12,29 +14,43 @@ errcho() {
...
-# EOL char is added (see commit 9f3361ac39c3).
-[[ ! $@ =~ "pcap" ]] && term_opts="-it"
+# Parse (w/o enforcing validity) Retis arguments to allow custom actions.
+while [ $# -gt 0 ]; do
+	case $1 in
+	collect)
+		# Retis cannot run in an unprivileged Podman container to collect events.
+		if [[ $runtime == "podman" && $(id -u) -ne 0 ]]; then
+			errcho "Error: Retis collection cannot run in an unprivileged container."
+			exit -1
```
amorenoz: I don't quite enjoy having ccli parsing at this level, Is it time to consider removing some of it? i.e: maybe we should just let the command fail with a pretty expressive:

```
Error: Failed to create events map: Operation not permitted (os error 1)
```

And avoid:
- users using "collect" as output filename, netdev name (in meta-filter) or other weird but still valid cases
- users not having the same experience as in the retis cli where you can collect with "retis c" retis "co", retis "coll", etc

tools/retis_in_container.sh
```diff
@@ -12,29 +14,43 @@ errcho() {
...
+		# but is needed to mount tracefs/debugfs.
+		[ $runtime == "docker" ] && extra_args="--security-opt apparmor=unconfined"
+		;;
+	pcap)
+		# We can't use a pseudo-tty (see `-t` option in `man podman
+		# run`) when using a command outputting a specific format to
+		# stdout (which could be piped into another utility parsing
+		# it), like the pcap command. This is because an extra EOL char
+		# is added (see commit 9f3361ac39c3).
+		term_opts=""
```
amorenoz: Following the same idea (trying to minimize the amount of hacks we do here). Why don't remove "-t" altogether and remove this case?

tools/retis_in_container.sh
```diff
@@ -25,6 +25,13 @@ fi
 # Parse (w/o enforcing validity) Retis arguments to allow custom actions.
 while [ $# -gt 0 ]; do
 	case $1 in
+	-P|--extra-profiles-dir)
```
amorenoz: I have the same concerns, "-P" could be an argument to a python script.

I don't have a good solution for this ... other than allowing users to add the mount point manually?

`RETIS_CONTAINER_ARGS="-v my/path:/root/somewhere" -P /root/somewhere collect`

retis/src/profiles/cli.rs
atenart: That would be a lot better, will do in next version.

retis/src/collect/collector/nft/nft.rs
atenart: > Is this because of some user report?

Yes, me :D

> I'm afraid changing this default behavior might make users think packets are not traversing nft when they are.

I get that point, however issuing an event for each accept verdict is *really* verbose on a simple system like my laptop. On a server with more complex rules things get unreadable quickly. It's worth mentioning `accept` is not just the final rule hit a packet (not being dropped), it's every single chain traversed.

Another thing to consider is we don't often see the issue and I'm expecting users as well, because `--allow-system-changes` is required for the `nft` collector to run, which I assume many (including me) don't set unless required. The point here is not that the issue is a minor one, but that we might have a wrong impression in or daily use.

tools/retis_in_container.sh
atenart: > I don't quite enjoy having ccli parsing at this level

Me neither and as you noted it's impossible to get this right w/o having false positives. I agree with removing the checks when the underlying error is understandable (like here).

tools/retis_in_container.sh
atenart: Or just using a profile accessible from the current directory (or within its tree) as it's automatically mounted. That already works.

Maybe we can still try to detect and issue a warning if we see the option used with a path that won't be accessible? But this could still collide, especially with `-P`...

tools/retis_in_container.sh
atenart: Not using `-t` prevents the pager from running. I guess the cons of having a false positive here is acceptable (no pager), WDYT?

amorenoz: Ugh... I mean ... is it that bad if the user has to add "| less"  if using retis_in_container?
OTOH, is it **that** common to use `retis_in_container.sh pcap | tcpdump -r -` .... we can write it to a file and then read it with tcpdump, wireshark or whatever ...

atenart: > Ugh... I mean ... is it that bad if the user has to add "| less" if using retis_in_container? OTOH, is it **that** common to use `retis_in_container.sh pcap | tcpdump -r -` .... we can write it to a file and then read it with tcpdump, wireshark or whatever ...

Sure. But while I do agree on trying to limit the fixups this one has minimal cons while nice pros IMHO and no one complained so far. So maybe let's wait for an issue to be raised?

amorenoz: I'm not sure about the pros/cons balance, given the user might want to pipe the output of any command to any other tool that that won't work.

How about we add this instead:
```bash
[[ -t 1 ]] || term_opts=""
```
This will remove the pseudo-tty if stdout is not a terminal, so any pipe will work without extra characters (including to tcpdump) and pager should also work if user did not pipe the output of retis_in_container.sh. WDYT?

atenart: I didn't know about `-t`, that sounds a lot better! Let me add this to the PR.

retis/src/collect/collector/nft/nft.rs
amorenoz: OK, I haven't used nft that much. @vlrpl are you ok with this?

vlrpl: I agree this changes expectation. It's very hard to say whether this is a real problem or not, that is, someone is actually relying on this.
nft is about drops, as much as retis is. Not just that. Filtering through the existing option allows to just limit to a single verdict, so a minimal profile can help.
That being said, if this default happens to change, `nft_verdicts: drop` in the drop profile becomes redundant and should be removed (which further highlights the fact the change turns nft in a drop module by default).

atenart: I'm not sure what is the way forward, so I removed `nft_verdicts` from the profile. If you'd prefer to drop the whole patch instead let me know (please try `retis c --allow-system-changes` first, even on your laptop).

vlrpl: me neither, but I'm not against dropping it (accept verdict, not the patch :). I guess the need, though, comes mostly from the automatic behavior of enabling everything (supported) while collecting, otherwise someone investigating nft chains would normally enable it.
Just thinking out loud as a last resort, would it help/make sense to move policy in the configuration knob?

atenart: > I guess the need, though, comes mostly from the automatic behavior of enabling everything (supported) while collecting, 

Right, the default in Retis is currently only only enabling some features to get an output not being too verbose for the only purpose of having a first look at what's going on. We could even argue that a packet being dropped by a nft rule is perfectly normal there... which could point to the question of wheter we should enable all collectors by default. I guess that's a question for another day :)

> Just thinking out loud as a last resort, would it help/make sense to move policy in the configuration knob?

Not sure to get it, do you have an example of what it would look like?

vlrpl: > > Just thinking out loud as a last resort, would it help/make sense to move policy in the configuration knob?
> 
> Not sure to get it, do you have an example of what it would look like?

as of today, we carry in cfg->verdict what we want to send up through
the ringbuf. The idea was to add policy and check against it as well.
It may sound a bit an intruder (given it's not a verdict, but more a
superset of verdicts), but it depends on how one looks at it.
The semantic should be: for policies return allow listed verdicts, otherwise,
skip it. The new policy knob will remain disabled by default, of course.

Note: I'm not pushing toward it, just considering a different angle.

atenart: Ah I see thanks! Yeah that's a good idea, an explicit allow or deny is probably more interesting that a default one. We can always create a new issue for this and see if it gets traction?

vlrpl: > Ah I see thanks! Yeah that's a good idea, an explicit allow or deny is probably more interesting that a default one. We can always create a new issue for this and see if it gets traction?

sgtm. In the meantime, just to be explicit, I'm ok with this change. If someone complains, we'll revert this through zstream.

amorenoz: >  (please try retis c --allow-system-changes first, even on your laptop).

The verbosity of this command is equivalent to the verbosity of enabling the ovs module in a system where OVS handles traffic. IMHO, ovs and nft are both generic packet processing frameworks and when you enable them you should expect to see many packets being processed.

atenart: This makes me wonder if we should enable all collectors by default. Anyway for now I removed that commit.

retis/src/profiles/cli.rs
```diff
@@ -51,11 +51,17 @@ impl SubCommandParserRunner for ProfileCmd {
 impl ProfileCmd {
     fn list_path(path: &Path) -> Result<()> {
         for entry in path.read_dir()? {
-            let entry = entry?;
-            match Profile::from_file(entry.path()) {
+            let entry = entry?.path();
+
+            // Skip non-YAML files.
+            if !entry.extension().is_some_and(|e| e == "yaml" || e == "yml") {
+                continue;
```
vlrpl, line 59: genuinely asking, does it make sense to add a debug message here with what we're skipping?

Not sure this will ever happen, but a yaml without an extension is always a yaml file. Not super sure how much the debug will help in such a case, but at least we try to tell the user :)

atenart: I'm not sure the message would provide much, one can use `retis profile list` to see all valid profiles and figure out something's not right if one is missing. So I'm wondering if we should not instead make it clear in the doc & help as the behavior is not dynamic?

vlrpl: a note in the doc would work for me

atenart: I consolidated the `retis --help`, `retis profile --help` and documentation wrt to yaml/yml filtering. I also added the extension filtering logic in `Profile::from_dir` to have a consistent behavior.

PR #603, opened 2026-02-16
Introduce exit callbacks

### LINKED ISSUE #527: Early termination combined with --allow-system-changes may not release all the resources
If the debugfs has not been mounted previously, terminating very early (i.e., right after the debugfs gets mounted) leaves the `debugfs` mounted (if the tool applies system changes, they are normally supposed to be cleaned up on termination).

This is probably a little broader issue (not strictly related to the `debugfs`).

atenart:
Callbacks can be registered to be run at either termination signal time or exit time:

- `Running` is converted to use exit callbacks.
- Auto-mounted paths are unmounted at exit time using this new mechanism.

Fixes #527.

### DIFF
--- a/retis/src/collect/collect.rs
+++ b/retis/src/collect/collect.rs
@@ -121,15 +121,12 @@ impl Collectors {
         let factory = BpfEventsFactory::new()?;
         let probes = ProbeManager::new()?;
 
-        let run = Running::new();
-        run.register_term_signals()?;
-
         Ok(Collectors {
             collectors: HashMap::new(),
             probes,
             factory,
             known_kernel_types: HashSet::new(),
-            run,
+            run: Running::new()?,
             tracking_gc: None,
             tracking_config_map: None,
             stack_tracking_config_map: None,
--- a/retis/src/helpers/signals.rs
+++ b/retis/src/helpers/signals.rs
@@ -23,27 +23,16 @@ pub(crate) struct Running {
 }
 
 impl Running {
-    pub(crate) fn new() -> Running {
-        Self {
-            condition: Arc::new(AtomicBool::new(false)),
-            callbacks: Arc::new(Mutex::new(Vec::new())),
-        }
-    }
-
-    /// Add a new callback to be run on `Drop`.
-    pub(crate) fn add_drop_cb<F>(&mut self, cb: F)
-    where
-        F: FnOnce() + Send + Sync + 'static,
-    {
-        self.callbacks.lock().unwrap().push(Box::new(cb))
-    }
-
-    /// Register termination signals so the current Running instance will stop
-    /// upon receiving one of those signals (SIGTERM, etc). This can only work
-    /// from the main thread.
-    pub(crate) fn register_term_signals(&self) -> Result<()> {
+    // Create a new Running instance.
+    //
+    // - Helps handling loops in various threads to synchonize termination.
+    // - This instance will also stop upon receiving one of the termination
+    //   signals (e.g. SIGTERM).
+    pub(crate) fn new() -> Result<Running> {
         let mut sigs = Signals::new(signal_hook::consts::TERM_SIGNALS)?;
-        let condition = Arc::clone(&self.condition);
+
+        let run = Self::ignore_signals();
+        let condition = Arc::clone(&run.condition);
 
         thread::spawn(move || {
             sigs.wait();
@@ -51,7 +40,24 @@ impl Running {
             info!("Received signal, terminating...");
         });
 
-        Ok(())
+        Ok(run)
+    }
+
+    // Same as `new()` but without handling termination signals. Termination
+    // *must* be manually handlded here.
+    pub(crate) fn ignore_signals() -> Running {
+        Self {
+            condition: Arc::new(AtomicBool::new(false)),
+            callbacks: Arc::new(Mutex::new(Vec::new())),
+        }
+    }
+
+    // Add a new callback to be run on `Drop`.
+    pub(crate) fn add_drop_cb<F>(&mut self, cb: F)
+    where
+        F: FnOnce() + Send + Sync + 'static,
+    {
+        self.callbacks.lock().unwrap().push(Box::new(cb))
     }
 
     pub(crate) fn running(&self) -> bool {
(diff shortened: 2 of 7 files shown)


retis/src/helpers/signals.rs
```diff
@@ -5,15 +5,77 @@
...
+        let callbacks = Callbacks(Mutex::new(Vec::new()));
+
+        thread::spawn(move || {
+            sigs.wait();
+            info!("Received signal, terminating...");
+            run_exit_callbacks();
+        });
+
+        unsafe {
+            libc::atexit(run_exit_callbacks);
```
amorenoz: Instead of implementing an external atexit callback, have you considered implementing everything within `Running`? Since it's already ubiquitous and used for these kind of things, it could hold the internal callback list and call it during drop().

retis/src/helpers/signals.rs
```diff
@@ -5,15 +5,77 @@
...
+    fn run(&self) {
+        self.0.lock().unwrap().drain(..).for_each(|cb| cb())
+    }
+}
+
+impl Drop for Callbacks {
+    fn drop(&mut self) {
+        self.run()
+    }
+}
```
amorenoz: Isn't this dead code? Callbacks is static, I think it won't get dropped.

retis/src/helpers/signals.rs
atenart: I have but did not for two reasons:

- `Running` goal is to handle processing loops while the exit callbacks are for cleaning up on termination. I feel like this would be mixing two different concepts.
- Not all sub-commands are using `Running`, e.g. when not processing things in a loop. Having the two concepts split allows for all code paths to register an exit cb if needed.

retis/src/helpers/signals.rs
atenart: I implemented the `Drop` trait for completeness but you're right this is not used atm. Without it I feel `Callbacks` would be missing something and it's only a one-liner implementation. Also `Callbacks` could be used outside the exit cb logic, not that this is the main reason for implementing `Drop`.

retis/src/helpers/signals.rs
amorenoz: Right, we would convert `Running` to an object that handles program lifetime as it already controls when the program is interrupted. I don't think it's too much of a stretch. The usecase is interrupting the program before we actually start using `Running` so enhancing its functionality to be usable earlier sounds reasonable to me. OTOH, simple sub-commands not using `Running` do not need this cleanup callback mechanism anywyas.

I'm a bit concerned about direct use of atexit as there has been quite some issues in the past regarding the program being already half crumbling when atexit is called, so I think running during global drop destructors might be safer.
AFAICS, `libc::atexit` runs in the C-level cleanup phase, after all the rust stuff has been cleaned up, i.e: i/o has been already flushed so prints can be missed, thread-local storage has been destroyed, many global resources might have been (partially) cleaned up (e.g: logger?), there are potential thread-safety issues ([here](https://internals.rust-lang.org/t/rusts-exit-forwards-to-cs-exit-but-doesnt-warn-about-ub/19458/2) or [here](https://github.com/rust-lang/rust/issues/126600))...

I'm not sure this use-case justifies this amount of uncertainty (clearly expressed in the required `unsafe` keyword) when a simpler RAII implementation is easily available.

retis/src/helpers/signals.rs
amorenoz: In the (rare) case this struct is ever used in the future with some other purpose, I would prefer a comment warning future devs that they _may_  need to add a Drop implementation than to add dead code at this moment.

retis/src/helpers/signals.rs
atenart: > Right, we would convert `Running` to an object that handles program lifetime as it already controls when the program is interrupted. I don't think it's too much of a stretch. The usecase is interrupting the program before we actually start using `Running` so enhancing its functionality to be usable earlier sounds reasonable to me. OTOH, simple sub-commands not using `Running` do not need this cleanup callback mechanism anywyas.
> 
> I'm a bit concerned about direct use of atexit as there has been quite some issues in the past regarding the program being already half crumbling when atexit is called, so I think running during global drop destructors might be safer. AFAICS, `libc::atexit` runs in the C-level cleanup phase, after all the rust stuff has been cleaned up, i.e: i/o has been already flushed so prints can be missed, thread-local storage has been destroyed, many global resources might have been (partially) cleaned up (e.g: logger?), there are potential thread-safety issues ([here](https://internals.rust-lang.org/t/rusts-exit-forwards-to-cs-exit-but-doesnt-warn-about-ub/19458/2) or [here](https://github.com/rust-lang/rust/issues/126600))...
> 
> I'm not sure this use-case justifies this amount of uncertainty (clearly expressed in the required `unsafe` keyword) when a simpler RAII implementation is easily available.

OK, that does not sound good. You're right, let's convert `Running` to also handle exit cbs on ctrl+c.

retis/src/collect/collect.rs
```diff
@@ -203,11 +203,19 @@ impl Collectors {
         if collect.allow_system_changes {
             if let Some(mounted) = Self::try_mount("tracefs", "/sys/kernel/tracing") {
                 if mounted {
-                    self.mounted = Some(PathBuf::from("/sys/kernel/tracing"));
+                    self.run.add_drop_cb(|| {
+                        if let Err(e) = umount("/sys/kernel/tracing") {
+                            error!("Could not umount /sys/kernel/tracing: {e}");
+                        }
+                    });
```
amorenoz, line 210: nit The ultra-paranoid would say it's still possible to mount the fs and panic before we add the callback, maybe the callback should `try_unmount` and be registered first?

atenart: I don't think that would work as we don't want to umount those filesystems if Retis was not the one to mount them (in many systems they are mounted by default). We register the umount cb after a successful mount operation as this is a way to know if it was already mounted or not.

The only way to register the cb before would be to read and parse `/proc/mounts` first and use this info to trigger or mounting tracefs and debugfs. I can add this to the inspector and rework that part if you want.

amorenoz: Oh, that's right. Well I guess reading `/proc/mounts` is the right way to do it, but we can do that later

PR #604, opened 2026-02-19
Add more tests in the next CI

atenart:
Convert the following tests to the next CI:
- Drop.
- Tracking.
- Pcap.

The filter, nft and ovs tests have been purposely left out to allow others to have a first experience with the next CI framework.

Add a test for the `--out-rotation` option.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -22,6 +22,7 @@ dnf install -y \
     python3-pip \
     python3-devel \
     socat \
+    tcpdump \
     nftables \
     make \
     jq \
@@ -185,6 +186,7 @@ Vagrant.configure("2") do |config|
           python3-dev \
           openvswitch-switch \
           socat \
+          tcpdump \
           nftables \
           make \
           jq


tests/next/include/helpers.sh
```diff
@@ -32,3 +32,30 @@ two_ns_vlan_simple() {
...
+
+# Create three network namespaces, with veth pairs. Configuration is the
+# following:
+# ns0 - veth01 - 10.0.42.1/24 - 1111::1/64
+# ns1 - veth10 - 10.0.42.2/24 - 1111::2/64
+# ns1 - veth12 - 10.0.43.1/24 - 2222::1/64
+# ns2 - veth21 - 10.0.43.2/24 - 2222::2/64
+#
+# The correct routes are installed in ns0 and ns2. Forwarding is enabled on ns1.
+three_ns_simple() {
```
amorenoz: I'm wondering at what point we should drop the "simple" part ... Given we don't have a "complex" version, should we just rename them all to "tree/two_ns" .. etc?

tests/next/include/helpers.sh
```diff
@@ -32,3 +32,30 @@ two_ns_vlan_simple() {
 	ip -net ns0 address add 2222::1/64 dev veth01.123
 	ip -net ns1 address add 2222::2/64 dev veth10.123
 }
+
+# Create three network namespaces, with veth pairs. Configuration is the
+# following:
+# ns0 - veth01 - 10.0.42.1/24 - 1111::1/64
+# ns1 - veth10 - 10.0.42.2/24 - 1111::2/64
```
amorenoz, line 38: mention VLAN h/w offload off?

tests/next/pcap.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+source $(dirname $0)/include/lib.sh
+source $(dirname $0)/include/helpers.sh
+
+pcap_tcp_cc() {
+	two_ns_simple
+
+	ip netns exec ns1 socat TCP-LISTEN:80 /dev/null &
```
amorenoz: can't we leak the process if retis or the connectivity fails?

tests/next/include/helpers.sh
atenart: Right :smile: I can s/simple// for next version.

tests/next/include/helpers.sh
atenart: There are no VLAN in there, it extends `two_ns` not the vlan version.

PR #607, opened 2026-03-02
Support reading split files in Python

atenart:
Some pre-requisites required to convey the rotation config down to the Python helpers. This is actually improving things for all users.

### DIFF
--- a/retis-events/src/file/rotate.rs
+++ b/retis-events/src/file/rotate.rs
@@ -20,9 +20,6 @@ pub struct EventFile {
     pub path: PathBuf,
     /// Read split event files in sequence, if any.
     pub use_rotation: bool,
-    /// Allow to fallback reading the event file using a known split file
-    /// extension.
-    pub try_split: bool,
 }
 
 /// Rotation policy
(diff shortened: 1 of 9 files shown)


retis-events/src/file/rotate.rs
```diff
@@ -13,6 +13,14 @@ use nix::sys::utsname::uname;
 
 use crate::{compat::json, file::guess_version, helpers::time::*, *};
 
+/// Represent an event file and how it should be used wrt rotation.
+#[derive(Clone, Debug)]
+pub struct EventFile {
+    pub path: PathBuf,
+    pub use_rotation: bool,
+    pub try_split: bool,
```
amorenoz: Please document public members.

amorenoz: in particular, when you come back to this code after a while, the meaning of "try_split" gets blurry (split, similar to rotation and "use_rotation" already present).

amorenoz (commented):
s/Add support from/Add support for" in commit 2's  message

Apart from those minor nits, lgtm

retis-events/src/file/rotate.rs
atenart: Fixed.

atenart:
I had to rebase to resolve a conflict.

atenart:
CI is not really failing, it just refused to launch two of the runtime tests for now...

amorenoz (approved):
LGTM

Issue #153, opened 2023-05-16
Remove manual installation of Vagrant for runtime tests


atenart:
We manually install Vagrant for our runtime tests as a workaround for https://bugs.launchpad.net/vagrant/+bug/2017828. Remove it when we can.

amorenoz:
As per https://github.com/retis-org/retis/pull/616#issuecomment-4083250820, we might need to also find an alternative to vagrant or find suitable images.

PR #616, opened 2026-03-18
fix: remove manual installation of Vagrant for runtime tests

Deepak8858:
Removes the manual installation of Vagrant from the Hashicorp repository in the Cirrus CI configuration. Since the original issue (broken box downloads in Ubuntu's Vagrant version) has been addressed in newer releases, we can simplify the setup and use the system-provided package.\n\nAlso fixes a small typo in the environment variable:  -> .\n\nFixes #153

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -152,12 +152,7 @@ functional_task:
   # Initialization of the test environment
   setup_script: &vagrant_setup
     - grep -q vmx /proc/cpuinfo # Ensure nested virtualization is enabled
-      # The version of vagrant shipped with ubuntu fails to download some boxes (e.g: f38).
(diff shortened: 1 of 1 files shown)


amorenoz:
Not only it has not been fixed, but Canonical has dropped support for vagrant altogether.
The package has been dropped (only old archives left): https://documentation.ubuntu.com/public-images/public-images-explanation/vagrant/#support
And images will no longer be created and maintained: https://documentation.ubuntu.com/public-images/public-images-explanation/vagrant/#support

atenart:
This cannot be fixed this way and should at the very least be tested before submission. The timeline is also surprising. Closing this.

PR #617, opened 2026-03-24
Ftrace opt

vlrpl:
Introduce ftrace option.

Might still be revisited, but initial feedback is welcome


retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -10,16 +10,24 @@ struct skb_tracking_event {
 	u64 skb;
 } __binding;
 
-DEFINE_HOOK(F_AND, RETIS_ALL_FILTERS,
+DEFINE_HOOK(F_OR, RETIS_ALL_FILTERS | RETIS_F_STACK_PASS,
```
vlrpl: note to self: this has to remain F_AND. Alternatively, the OR has to be between the whole filtering and stack.

retis/src/collect/cli.rs
```diff
@@ -51,6 +51,7 @@ Wildcards (*) can be used, eg. \"kprobe:tcp_*\" or \"tp:skb:*\".
 
 OPTIONS can be used to configure probes on a per-probe basis. Options are a list of keywords separated by '/' (e.g. TARGET/opt1/opt2). Valid OPTIONS:
 - stack: enables stack traces retrieval (same as \"--stack\", on a per-probe basis).
+- ftrace: enables ftrace-style tracking for kprobes on functions whose arguments do not include any trackable object directly. It's only valid for kprobes. A matching kretprobe is automatically added to close the tracking window. Within the window, probes on functions with known types will propagate the tracking context as if they had been hit by skb-aware functions.
```
atenart: I do understand the description but it's quite low level, maybe we could find a more user driven one? E.g. `The function tracer option allows reporting events from functions called within the ftrace-enabled probe target, even those not reporting events under normal conditions (e.g. no known type in their arguments). Those functions should be explicitly probed too ("--probe").`

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -10,16 +10,24 @@ struct skb_tracking_event {
...
 	skb = retis_get_sk_buff(ctx);
-	if (!skb)
-		return 0;
+	if (skb) {
+		ti = skb_tracking_info_by_skb(skb);
+	} else {
+		refp = stack_get_skb_ref(ctx->stack_base);
+		if (!refp || !(*refp & FTRACE_WINDOW))
+			return 0;
+		ref = *refp & ~(FTRACE_SENTINEL | FTRACE_WINDOW);
```
atenart: `#define FTRACE_MASK (FTRACE_SENTINEL | FTRACE_WINDOW)` ?

retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -346,9 +344,15 @@ static __always_inline int chain(struct retis_context *ctx)
...
-		track_skb_start(ctx);
-	else if (skb)
+	if (RETIS_TRACKABLE(ctx)) {
+		cur_ref = stack_get_skb_ref(ctx->stack_base);
+		in_window = cur_ref &&
+			(*cur_ref & (FTRACE_SENTINEL | FTRACE_WINDOW));
+		ref = track_skb_start(ctx);
+		if (ref && (in_window || (cfg->ftrace &&
+			    ctx->probe_type == KERNEL_PROBE_KPROBE)))
+			*ref |= FTRACE_WINDOW;
```
atenart: IMO the additional lines could go into track_skb_start.

docs/collectors/overview.md
```diff
@@ -48,6 +48,22 @@ This tracking information, which is basically a unique "id", can be used at
...
+
+- If the outer function takes a known type (e.g. `struct sk_buff`), the tracking
+  context is established immediately at entry and tagged so that inner probes can
+  find it. Inner probes that fire within the window will emit a tracking event
+  with the tracking id inherited from the outer; the skb address in those events
+  will be `0` since no known type is directly available at the inner probe site.
+- If the outer function does not take any known type, a sentinel is stored
+  instead to mark the window as open. If an inner SKB-aware probe fires within
+  that frame it replaces the sentinel with a real tracking reference; if no such
+  probe fires the window closes without producing any tracking event.
```
atenart: (Sorry if I forgot previous discussions)

I'm not sure about supporting this specific case:

```
fn0()
  fn1()
  fn2(skb)
  fn3()
```

We would see an event for `fn2` & `fn3`, not `fn1`. That's on purpose but might be very confusing for users?

Going back a little bit, it makes sense to allow attaching a non-skb function to an skb tracking id when called from within.

On the other hand, allowing to generate an skb tracking id mid-function and attaching only subsequent calls seems like a partial solution. In such case I'm wondering if reporting a stack tracking id (new event section) would make more sense? We could then track all calls from non-skb functions, and reconstruct the timeline in post-processing in all cases. The counter argument being we do operate on networking data and we're not ftrace I guess. Also not saying we should do this right now.

retis/src/core/probe/kernel/bpf/include/common.h
atenart, line 377: If `RETIS_F_STACK_PASS` is set and `FTRACE_WINDOW` is not set, can we skip the hooks too?

retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -73,38 +69,37 @@ enum {
...
 	SEC("ext/hook")								\
 	int hook_name(struct retis_context *ctx, struct retis_raw_event *event) \
 	{									\
 		/* Let the verifier be happy */					\
 		if (!ctx || !event)						\
 			return 0;						\
-		if (!((fmode == F_OR) ?						\
-		      (ctx->flags & (fflags)) :				\
-		      ((ctx->flags & (fflags)) == (fflags))))		\
+		if (!((ctx->flags & RETIS_F_STACK_PASS) ||			\
```
atenart: Why not set `RETIS_F_STACK_PASS` in `RETIS_ALL_FILTERS`?

retis/src/core/probe/kernel/utils.rs
```diff
@@ -94,7 +98,16 @@ where
...
+            Kprobe => {
+                let p = Probe::kprobe(symbol.clone())?;
+                // Paired kretprobe to close the ftrace window.
+                if options.contains(&ProbeOption::Ftrace) {
+                    let mut kret = Probe::kretprobe(symbol)?;
+                    kret.set_option(ProbeOption::Ftrace)?;
+                    probes.push(kret);
+                }
+                p
+            }
```
atenart, line 117: This will attach a kprobe twice on the function entry. I think it's time to remove the early `if (kretprobe)` return in `probe_kprobe` and have a probe config (e.g. `cfg->enabled`) bool to control whether to return early or continue processing from `chain`. Kretprobes will then have a per-probe ability to allow processing the kprobe part. We can control this using `ProbeOption::Ftrace` for now.

With that the above needs to be changed to `Kretprobe || options.contains(&ProbeOption::Ftrace) => ...`  instead.

atenart (commented):
Nice additional cleanups :)

retis/src/collect/cli.rs
amorenoz: I like @atenart 's wording better but that still does not convey the fact that we can probe a non-skb functions but events will only be attached after an skb-function (that satisfies filter) gets hit. Not saying we should put all this into the cli help, but this is a complex feature and maybe we have to point users to the documentation?

retis/src/core/tracking/bpf/include/stack_tracking.h
```diff
@@ -63,36 +71,31 @@ static __always_inline u64 get_stack_base(void *ctx, enum kernel_probe_type type
 	return get_base_addr(&stack_addr);
 }
 
-static __always_inline u64 track_stack_update(u64 key, u64 value)
+static __always_inline u64 *track_stack_update(u64 key, u64 value, u64 flags)
```
amorenoz: Why do we need to pass `flags`? AFAICS we always pass `BPF_ANY`

docs/collectors/overview.md
amorenoz: Right. I am also struggling to find an example for the sentinel feature. Most of the stack is skb-driven and other, typically smaller, functions are mostly inlined.
So I wonder how much extra usecase are we trying to cover in exchange of the added complexity.

retis/src/core/probe/kernel/utils.rs
vlrpl: good catch! Fairly bad having it twice. I recall I spotted this at some point before writing the patch, but eventually I clearly missed it. Kind of "I remembered in one dream and I forgot in another" :)

retis/src/core/tracking/bpf/include/stack_tracking.h
vlrpl: I locally changed it over time and this is a leftover. Thanks!

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: because in such case all the flags have to be set at once to pass.
We want disjunction between STACK (it's a logical superset once set) and the others (ideally never set for the non skb case).

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: kept outside to avoid passing in_window (or cur_ref) and keep some separation. We can move it in if you have a preference, though.

docs/collectors/overview.md
vlrpl: I considered the sentinel fine because, although it may not be straightforward it's explicit. My assumption is that an user not understanding more or less advanced features should not use them expecting them being self explanatory. We can reduce the complexity, but additional control requires some.
A codepath I tested is `-p "k:ip_push_pending_frames/ftrace" -p "k:__ip_local_out" -p "k:ip_send_check"`.
Sometimes we might want to take a look at how a caller behaves on more packets (in relation to a list/batch containing our[s]). That's the idea behind it. Not saying it is perfect :)
What do you guys think?

About the section, in one of the initial iterations (older PR) I tentatively added a field in the kernel section (not exactly a separate section, but could have been an empty kernel section, so still decoupled from skb_tracking's), but we then agreed to generate skb_tracking events given we always tie the feature to skbs. We can consider a separate section, maybe as part of a follow up, if there are strong preference/advantages.

retis/src/collect/cli.rs
vlrpl: I'm fine changing the wording + pointing to the doc.

retis/src/core/probe/kernel/bpf/include/common.h
atenart: I don't have a huge preference, just thought since the is linked to skb tracking and there's already stack tracking logic in `track_skb_start` we could consolidate there.

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: Just a note. With the sentinel removal this should not be a case anymore

retis/src/core/probe/kernel/utils.rs
vlrpl: On a second thought, it's not that bad :)
Reusing kretprobe/kprobe infrastructure is nice, but would require at least to freplace on the kprobe as well (didn't investigate much, but I'm not sure nhooks for the kprobe/kretprobe case is always consistent for the kprobe).
This would require a rework that can be tackled in multiple ways (e.g. with the simple "always freplace for kprobes as well and do some minor adjustments on the ebpf side"), or better simply extend/generalize the behavior to better accommodate all the users (like a non-automatic kretprobe[?]). I think this could dilute the scope of this PR and slow it a bit down.
We can speak more about this offline, though (unless you prefer to carry on the discussion here :)

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
vlrpl: note: this is a good one, but with the sentinel removal does not apply anymore.

retis/src/collect/collect.rs
```diff
@@ -397,20 +400,38 @@ impl Collectors {
...
-            let ok = self.known_kernel_types.iter().any(|t| {
+        let filter = |symbol: &Symbol, options: &HashSet<ProbeOption>| -> Result<bool> {
+            let has_known_type = self.known_kernel_types.iter().any(|t| {
                 symbol
                     .parameter_offset(t)
                     .is_ok_and(|offset| offset.is_some())
             });
-            if !ok {
+
+            if options.contains(&ProbeOption::Ftrace) && !has_known_type {
```
vlrpl: note for the discussion:

This is a stretch as known_types are broader than the one we should enforce (like known_packet_types). Left it here just to discuss about what we may want (e.g. per known_packet_types or simply open code sk_buff and nft_traceinfo).

retis/src/core/probe/kernel/utils.rs
atenart: I had in mind the hooks were shared, but you're right it's a different loaded program. We can keep things as-is for now, that's an optimization we can do for later.

retis/src/collect/collect.rs
atenart: Fixing this properly is not so simple. A few options:

- We know there's a little gap in here but the documentation explicitly state that only skb-aware functions are supported. Other use cases are undefined behavior.
- We can hardcode some types in the filter function setting `has_known_type` looking for any type we know we can get an skb from, and use a new bool to denote that. Then we can use it in the above condition.
- We can use BTF to explore known types in the filter function looking for an skb. This is similar to the above point, but dynamic.

retis/src/collect/collector/skb_tracking/bpf/tracking_hook.bpf.c
```diff
@@ -10,16 +10,24 @@ struct skb_tracking_event {
...
-	if (!skb)
-		return 0;
+	if (skb) {
+		ti = skb_tracking_info_by_skb(skb);
+	} else {
+		refp = stack_get_skb_ref(ctx->stack_base);
+		if (!refp || !(*refp & FTRACE_WINDOW))
+			return 0;
+		ref = *refp & ~FTRACE_WINDOW;
+		ti = skb_tracking_info(&ref);
```
atenart: nit: I would add an helper instead (e.g. `skb_tracking_info_by_stack`) to avoid leaking stack tracking logic in hooks.

retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -437,6 +434,16 @@ static __always_inline int chain(struct retis_context *ctx)
...
 
+	/* Close the ftrace window by stripping the ftrace tag bits so that
+	 * potentially subsequent consumers that rely on stack tracking are
+	 * not affected.
+	 */
+	if (cfg->ftrace && ctx->probe_type == KERNEL_PROBE_KRETPROBE) {
+		ref = stack_get_skb_ref(ctx->stack_base);
+		if (ref && (*ref & FTRACE_WINDOW))
+			*ref &= ~FTRACE_WINDOW;
+	}
```
atenart: Could we also have this in `track_skb_end` for symmetry? Clearing the `FTRACE_WINDOW` can be done only when `RETIS_TRACKABLE` is true I think, so the call site doesn't need to be updated.

retis/src/core/probe/kernel/bpf/include/common.h
atenart: `RETIS_F_STACK_PASS` is set if the stack is known, not if it is in an `FTRACE_WINDOW`. We might need an extra flag to denote the two possibilities.

I guess the way is to add another helper similar to `stack_is_tracked` for ftrace stacks.

Additional ask: on top of the definition of `RETIS_F_STACK_PASS` and the extra one for ftrace, it would be nice to have a comment explaining the difference between the two and their goals.

retis/src/core/probe/kernel/utils.rs
```diff
@@ -73,28 +73,46 @@ pub(crate) fn parse_cli_probe(input: &str) -> Result<(CliProbeType, &str, HashSe
...
+    }
+
     // Convert the target to a list of matching ones for probe types
     // supporting it.
     let mut symbols = match r#type {
         Kprobe | Kretprobe => matching_functions_to_symbols(target)?,
         RawTracepoint => matching_events_to_symbols(target)?,
     };
 
+    // ftrace cannot be applied to wildcards matching multiple symbols.
```
atenart: Why not?

That makes me think we might not want to allow more than a single ftrace enabled probe, otherwise the following could happen?

```
kprobe_probe0: enable ftrace
  kprobe_probe1: enable ftrace (noop)
  kretprobe_probe1: disable ftrace
  missed_ftrace)event
kretprobe_probe0: noop
```

Another way would be to emit an error from the BPF side if trying to enable an ftrace section on an already enabled stack.

---

I saw later the documentation stating the above is UB. I think we could still issue a warning/error from BPF?

retis/src/core/tracking/bpf/include/stack_tracking.h
```diff
@@ -63,36 +66,31 @@ static __always_inline u64 get_stack_base(void *ctx, enum kernel_probe_type type
...
-static __always_inline u64 track_stack_update(u64 key, u64 value)
+static __always_inline u64 *track_stack_update(u64 key, u64 value)
 {
 	u64 addr = value;
 
-	if (!bpf_map_update_elem(&stack_tracking_map, &key,
-				 &addr, BPF_ANY))
-		return addr;
+	if (!bpf_map_update_elem(&stack_tracking_map, &key, &addr, BPF_ANY))
+		return bpf_map_lookup_elem(&stack_tracking_map, &key);
```
atenart: Not sure we should handle returning an outdated value as part of the update helper. IMO this should instead be handled in the call site if that's the wanted behavior. Bonus point, we can use `cur_ref` there an avoid an extra lookup.

retis/src/core/tracking/bpf/include/stack_tracking.h
```diff
@@ -63,36 +66,31 @@ static __always_inline u64 get_stack_base(void *ctx, enum kernel_probe_type type
...
-	return *val;
+	return bpf_map_lookup_elem(&stack_tracking_map, &stack_base);
 }
 
 static __always_inline bool stack_is_tracked(u64 stack_base)
 {
-	return stack_get_skb_ref(stack_base);
+	u64 *ref = stack_get_skb_ref(stack_base);
+
+	return ref && *ref;
```
atenart: I'm not sure why also checking `*ref` here. Isn't the value in the map a pointer, which if not NULL means an skb is associated with the stack?

docs/collectors/overview.md
```diff
@@ -48,6 +48,19 @@ This tracking information, which is basically a unique "id", can be used at
 post-processing time to reconstruct in-kernel packets flow using the `sort`
 post-processing command.
 
+When a probe is configured with the `ftrace` option (see `retis collect --help`),
+it opens a tracking window for the duration of the probed function call. The
```
atenart: nit: s/probed function call/function execution/

atenart (changes requested):

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: not sure I got exactly what you mean. It's set in both cases, otherwise we'd never reach skb-tracking hook sending the tracking event for the non-skb case.

Maybe you mean there's no straightforward way (FTRACE_WINDOW should be the way) to distinguish them (hence the flag)?

retis/src/core/probe/kernel/utils.rs
vlrpl: enabling ftrace in a wildcard doesn't make much sense, and it's not really meaningful. Nesting is an additional problem in that case. One can always specify the flag for a probe included in the wildcard but separately, right?

Nesting is [intentionally :D] bad. I admittedly didn't consider detecting it from ebpf (meaning, attempting to open the window twice), it might be a good idea.

retis/src/core/probe/kernel/bpf/include/common.h
atenart: I have the impression we only want probes to be called if and only if in an ftrace window, not just because the stack is being tracked, as my understanding is the stack is always tracked. E.g. if in non-skb function is probed and called, we only want events from the ftrace section.

But I might have missed something :)

retis/src/core/probe/kernel/utils.rs
atenart: > enabling ftrace in a wildcard doesn't make much sense, and it's not really meaningful. Nesting is an additional problem in that case. One can always specify the flag for a probe included in the wildcard but separately, right?

I think that should work, overriding with a more specific definition. Although not supporting wildcards is probably a good idea, to limit the scope of a complex mechanism.

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: But I may miss something, and worse, we can both miss something different at the same time :)
My take here is:

- skb tracking sends only ftraced events for the non-skb case (so no spurious events should be generated in the ovs path as things should be gated by in-window checks in the hook)
- ftrace/in-window and the ovs stack path always originates starting from skb-tracking. If a user sets inadvertently or intentionally the flag in a colliding place of the path, things should still follow the intended logic (so potential non-skb skb-tracking events are theoretically what the user intended).

atenart: > * skb tracking sends only ftraced events for the non-skb case (so no spurious events should be generated in the ovs path as things should be gated by in-window checks in the hook)

But why do we do the filtering in the skb-tracking collector? My impression is it would be better to filter before calling hooks, similarly to what is done with the filtering / skb tracking logic.

vlrpl: mostly because we have to separate the skb case from the other case.
Also, we should allow hooks calls for both cases, and RETIS_F_STACK_PASS is common.
```
	if (skb) {
		ti = skb_tracking_info_by_skb(skb);
	} else {
		refp = stack_get_skb_ref(ctx->stack_base);
		if (!refp || !(*refp & FTRACE_WINDOW))
			return 0;
		ref = *refp & ~FTRACE_WINDOW;
		ti = skb_tracking_info(&ref);
	}
```
The point was that, given we have to lookup anyways to get the `ref`, we can determine if this is a RETIS_F_STACK_PASS + FTRACE_WINDOW.
With the extra-flag we can maybe skip an extra lookup (only in the worst case path, that is, RETIS_F_STACK_PASS only), but I guess the check inside the hook can safely remain the same.

retis/src/collect/collect.rs
vlrpl: I guess item 3 is not very different than what we do (besides known_kernel_types). It will require the target type to be the sk_buff.
I wonder if this (regardless the proposed alternatives above) has to include `struct nft_pktinfo *` and `struct nft_traceinfo *`  or we just want to limit this to sk_buff (which is of course better than considering known_kernel_types)

atenart: Ideally I would say any skb that can be accessed directly or indirectly by Retis should be covered here. But this might be something for later when/if we tackle doing the skb retrieval in a more dynamic way.

retis/src/core/probe/kernel/bpf/include/common.h
atenart: Doesn't not filtering at the higher level on the ftrace window mean that all probes will always be called (but not fill an event), regardless of the tracking status?

vlrpl: what happens as of now is that skb-tracked skb aware functions always call hooks, if non-skb symbols are probed go through the hooks whether this is an ftrace_window or not. If we add a new flag I think we'll introduce some corner cases for RETIS_F_STACK_PASS (e.g. a clumsy user "ftraces" the ovs path).

atenart: The current logic is there is no early return based on filtering, however most hooks are only called if and only if filtering was positive. There is a special case allowing raw hooks to be called regardless of this (it also does not depend on anything, e.g. skb being available). But the main point is the filtering is done at the hook calling level and I don't think that should change for stack based filtering. The logic here relies on the skb-tracking collector to filter out non stack window calls.

This also impacts performances as every hook has to reject filling an event in its own way instead on relying on the infrastructure.

I would only call hooks if in a stack window. Is there a use case I'm not seeing in being more liberal?

atenart: > But the main point is the filtering is done at the hook calling level and I don't think that should change for stack based filtering.

To be a bit more clear, I'm concerned this is changing the logic to always call every hooks as we'll most likely be in an skb-aware stack most of the time.

vlrpl: > The current logic is there is no early return based on filtering, however most hooks are only called if and only if filtering was positive. There is a special case allowing raw hooks to be called regardless of this (it also does not depend on anything, e.g. skb being available). But the main point is the filtering is done at the hook calling level and I don't think that should change for stack based filtering. The logic here relies on the skb-tracking collector to filter out non stack window calls.
> 
> This also impacts performances as every hook has to reject filling an event in its own way instead on relying on the infrastructure.
> 
> I would only call hooks if in a stack window. Is there a use case I'm not seeing in being more liberal?

just to clarify, we're not changing any current behavior. If we're doing it, it's a bug.
`RETIS_F_STACK_PASS` is a superset of `FTRACE_WINDOW`.

`RETIS_F_STACK_PASS` is only set for non-skb function related to a tracked packet.
So, when the the pre-hook condition with RETIS_F_STACK_PASS is true, we're only talking about probed non skb functions.
`FTRACE_WINDOW` is just a way to determine whether this is in_window or not, but again it only serves its purpose on non-skb functions.

Something like ((ctx->flags & RETIS_F_STACK_PASS) || (ctx->flags & RETIS_F_WINDOW_PASS) is irrelevant, as RETIS_F_STACK_PASS will still be a superset of RETIS_F_WINDOW_PASS.

As of today, we cannot probe non-skb functions. With this patch, we can iff `ftrace` is set. We cannot statically sanitize whether that function is legitimately a callee, but that really is on the user. What we can do is generate an event iif it is a callee (in_window).

vlrpl: to summarize, the performance hit can happen when `ftrace` is set and a non-callee non-skb function in the packet path gets probed. 

Only those non-skb (and technically wrong) probes are affected by the extra cost (while generating nothing) due to hooks early exiting. Nothing happens as not a single hook can get its type plus skb-tracking detecting it to not be `in_window`.

atenart: > `RETIS_F_STACK_PASS` is only set for non-skb function related to a tracked packet. So, when the the pre-hook condition with RETIS_F_STACK_PASS is true, we're only talking about probed non skb functions. `FTRACE_WINDOW` is just a way to determine whether this is in_window or not, but again it only serves its purpose on non-skb functions.
> 
> Something like ((ctx->flags & RETIS_F_STACK_PASS) || (ctx->flags & RETIS_F_WINDOW_PASS) is irrelevant, as RETIS_F_STACK_PASS will still be a superset of RETIS_F_WINDOW_PASS.

Right but why calling all hooks only because we're in an skb-tracked stack? The condition could just be `ctx->flags & RETIS_F_WINDOW_PASS`, unless the hook is special and a raw probe can be used.

The current situation is:
1. Hooks are called only if the filtering logic was positive,
1. Unless special handling is needed and a raw hook can be used.

Unless I'm missing something this would make 1. only true for skb-aware probes and 2. matching both raw probes and stack tracked ones. `RETIS_F_STACK_PASS` to me is just an internal thing that is used as a building block for different features, namely kretprobe and `FTRACE_WINDOW` which is the real stack filtering condition.

The current situation would make any non-skb probe to call all hooks even if not in an ftrace window, and I'm not seeing why this should be the case. If special handling (looking at stack vs ftrace window tracking and/or skb tracking and/or misc things) a raw hook can be used.

atenart: I get that in practice there are not a lot of corner cases for now but why having a different logic for stack tracking and what is the benefit? (vs using raw hooks when flexibility is needed)

vlrpl: there's no different logic, it's just piggying back the old one.\
Probably, to do something like that we have to reduce the scope of `RETIS_F_STACK_PASS`, detach it from `in_window` and give up for good on `--ftrace` global option.

TBH, we have already the same with something as simple as (but not limited to):

```
sudo ./target/debug/retis collect -p "k:dev_get_by_index"
[...]

220: kprobe_multi  prog 1284
        kprobe.multi  func_cnt 1
        addr             cookie           func [module]
        ffffffffa184d210 ffffffffa184d210 dev_get_by_index
[...]


201: tracing  prog 1288
        prog_type ext  attach_type cgroup_inet_ingress
        target_obj_id 1284  target_btf_id 89
        pids retis(3419849)
202: tracing  prog 1290
        prog_type ext  attach_type cgroup_inet_ingress
        target_obj_id 1284  target_btf_id 91
        pids retis(3419849)
203: tracing  prog 1292
        prog_type ext  attach_type cgroup_inet_ingress
        target_obj_id 1284  target_btf_id 93
        pids retis(3419849)
204: tracing  prog 1294
        prog_type ext  attach_type cgroup_inet_ingress
        target_obj_id 1284  target_btf_id 95
        pids retis(3419849)
205: tracing  prog 1296
        prog_type ext  attach_type cgroup_inet_ingress
        target_obj_id 1284  target_btf_id 97
        pids retis(3419849)
206: tracing  prog 1298
        prog_type ext  attach_type cgroup_inet_ingress
        target_obj_id 1284  target_btf_id 99
        pids retis(3419849)
207: tracing  prog 1303
        prog_type ext  attach_type cgroup_inet_ingress
```

vlrpl: I think, considering the fact some time passed, and the approach changed, I ended up mixing older and newer approaches. It's not a big deal, but I think things can be improved.
I got your point, RETIS_F_STACK_PASS might not be needed at all. I'll revisit this part.

atenart: Thanks!

retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -45,66 +46,73 @@ struct {
...
+ *
+ *   F_GROUPS(A)       ->  F_ALL(A)
+ *   F_GROUPS(A, B)    ->  F_ALL(A) || F_ALL(B)
+ */
+#define _F_GROUPS1(a)			F_ALL(a)
+#define _F_GROUPS2(a, b)		F_ALL(a) || F_ALL(b)
+#define _F_GROUPS3(a, b, c)		F_ALL(a) || F_ALL(b) || F_ALL(c)
+#define _F_GROUPS4(a, b, c, d)	F_ALL(a) || F_ALL(b) || F_ALL(c) || F_ALL(d)
+#define _F_GROUPS_SEL(_1, _2, _3, _4, N, ...)	N
+#define F_GROUPS(...)	_F_GROUPS_SEL(__VA_ARGS__, _F_GROUPS4, _F_GROUPS3, _F_GROUPS2, _F_GROUPS1)(__VA_ARGS__)
```
atenart: I do like the `F_GROUPS` addition because it fits nicely in the filter flags logic and does not special case anything. This is good for maintenance and opens potential doors for later.

However I think all hooks should be called for `RETIS_F_WINDOW_PASS` by default in the `DEFINE_HOOK` definition, otherwise hooks like `dev` or `ns` won't fill their section on non-skb probes in an ftrace section. That or `RETIS_ALL_FILTERS` should include `F_GROUPS(RETIS_F_WINDOW_PASS)` in its definition, it is essentially a bypass of the other filter results.

retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -434,8 +445,8 @@ static __always_inline int chain(struct retis_context *ctx)
 	/* Cleanup stage while tracking an skb. If no skb is available this is a
 	 * no-op.
 	 */
-	if (RETIS_TRACKABLE(ctx))
-		track_skb_end(ctx);
+	if (RETIS_TRACKABLE(ctx) || (cfg->ftrace && ctx->probe_type == KERNEL_PROBE_KRETPROBE))
```
atenart: Can `cfg->ftrace && ctx->probe_type == KERNEL_PROBE_KRETPROBE` be true if `RETIS_TRACKABLE(ctx)` is false? I don't think so since ftrace sections must be on skb-enabled probes and since the ftrace logic only triggers if the skb is tracked. (Point is, the additional condition might not be needed).

retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -336,8 +344,10 @@ static __always_inline int chain(struct retis_context *ctx)
 	skb = retis_get_sk_buff(ctx);
 	if (skb)
 		ctx->flags = filter(skb);
-	else if (stack_is_tracked(ctx->stack_base))
-		ctx->flags = RETIS_F_STACK_PASS;
+
+	if ((RETIS_TRACKABLE(ctx) || !skb) &&
+	    stack_in_window(ctx->stack_base))
+		ctx->flags |= RETIS_F_WINDOW_PASS;
```
atenart: I'm not sure about restricting this to `RETIS_TRACKABLE(ctx) || !skb`. Wouldn't that make non-matching skb function to be left out, while still producing events for non-skb functions that might be called in the context of those non-matching skbs? Producing an event for all probes within an ftrace section is probably fine, we do report skb tracking + stack tracking information that can be used to understand what was the flow (and if the skb did not match). Also IMO any probe within an ftrace section should report events, that's the contract.

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: uhm, that's correct, it's just needed inside `track_skb_end()`. Probably remained moving things around.

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: Aren't those raw hooks? They should follow F_ALWAYS, don't they?

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: > I'm not sure about restricting this to `RETIS_TRACKABLE(ctx) || !skb`. Wouldn't that make non-matching skb function to be left out, while still producing events for non-skb functions that might be called in the context of those non-matching skbs? Producing an event for all probes within an ftrace section is probably fine, we do report skb tracking + stack tracking information that can be used to understand what was the flow (and if the skb did not match). Also IMO any probe within an ftrace section should report events, that's the contract.

here it depends on the assumption and what a non matching skb represents.
We know an skb along the path doesn't always represent the packet we are tracking but may contain something different, but in the common case it should. That said, I'm fine loosening this, the cost is more look ups, though.

retis/src/core/probe/kernel/bpf/include/common.h
atenart: That's right, those hooks should behave fine. The question remains for skb-aware probes in ftrace windows, which is discussed in another thread too.

retis/src/core/probe/kernel/bpf/include/common.h
atenart: > here it depends on the assumption and what a non matching skb represents.

Wrt this discussion here that could be any skb really. The case I have in mind could be:

```
ftraced' func(skb) {
    loop {
        fn1
        skb = dequeue_skb
    }
}
```

And not seeing:

```
skb event
fn1 event    # related to first skb
fn1 event    # related to second skb
```

But instead:

```
skb event
fn1 event    # related to first skb
second skb event
fn1 event    # related to second skb
```

vlrpl: I agree, we have to remove that check. We can have "weird" event sequences

retis/src/core/probe/kernel/bpf/include/common.h
vlrpl: we agreed offline that this should be fine, but @atenart feel free to point other cases out wrt this :)

retis/src/core/probe/kernel/bpf/include/retis_context.h
```diff
@@ -84,6 +84,17 @@ struct retis_context {
...
+#define RETIS_F_PASS(f, v)			\
+	RETIS_F_##f##_PASS_SH = v,		\
+	RETIS_F_##f##_PASS = 1 << v
+
+/* Defines the bit position for each filter */
+enum {
+	RETIS_F_PASS(PACKET, 0),
+	RETIS_F_PASS(META, 1),
+	RETIS_F_PASS(WINDOW, 2),
+};
```
atenart: Moving this here makes sense but I think `RETIS_ALL_FILTERS` should then be moved as well. Then there's the question of where all the `F_ALL`/`F_GROUPS` definitions should be.

retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -45,66 +46,73 @@ struct {
...
+/* OR of AND-groups: each argument is a bitmask that must be fully set (AND),
+ * and the groups are OR'd together. Up to 4 groups are supported.
+ *
+ *   F_GROUPS(A)       ->  F_ALL(A)
+ *   F_GROUPS(A, B)    ->  F_ALL(A) || F_ALL(B)
+ */
+#define _F_GROUPS1(a)			F_ALL(a)
+#define _F_GROUPS2(a, b)		F_ALL(a) || F_ALL(b)
+#define _F_GROUPS3(a, b, c)		F_ALL(a) || F_ALL(b) || F_ALL(c)
+#define _F_GROUPS4(a, b, c, d)	F_ALL(a) || F_ALL(b) || F_ALL(c) || F_ALL(d)
```
atenart: nit: the indentation is not the same for all 4 definitions.

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -164,11 +178,12 @@ static __always_inline int track_skb_start(struct retis_context *ctx)
...
-	if (!track_stack_update(ctx->stack_base,
-			       inv_head
-			       ? (u64)skb
-			       : (u64)head))
+	cur_ref = stack_get_skb_ref(ctx->stack_base);
+
+	if (track_stack_update(ctx->stack_base, inv_head ? (u64)skb : (u64)head))
 		log_error("While tracking stack. Unable to update the entry");
+	else
+		ref = cur_ref ? cur_ref : stack_get_skb_ref(ctx->stack_base);
```
atenart: Suggestion to save a map lookup:

```c
// Get the ref before the stack update.
ref = stack_get_skb_ref(ctx->stack_base);
new_ref = inv_head ? (u64)skb : (u64)head;

if (track_stack_update(ctx->stack_base, new_ref)
		log_error("While tracking stack. Unable to update the entry");
else if (!ref)
		ref = new_ref;
```

I think we could also go further and do something like to save another map access:

```c
// Get the ref before the stack update.
ref = stack_get_skb_ref(ctx->stack_base);
new_ref = inv_head ? (u64)skb : (u64)head;

if ((ctx->flags & RETIS_F_WINDOW_PASS) || (ftrace && ctx->probe_type == KERNEL_PROBE_KPROBE))
		new_ref |= FTRACE_WINDOW;

if (ref != new_ref) {
		if (track_stack_update(ctx->stack_base, new_ref)
				log_error("While tracking stack. Unable to update the entry");
		else if (!ref)
				ref = new_ref;
}
```

With that we can also not return a pointer from `stack_get_skb_ref` and also simplify things eg. in `stack_in_window`.

nit: we can probably do all the above at the end of `track_skb_start` to avoid interleaving the skb and the stack tracking logics.

tests/next/tracking.sh
```diff
@@ -97,4 +97,75 @@ EOF
...
+		-p "tp:net:netif_receive_skb/ftrace" \
+		--cmd "ip netns exec ns0 ping -c1 10.0.42.2" 2>/dev/null
+
+	! $retis collect -c skb,skb-tracking \
+		-p "kr:__ip_local_out/ftrace" \
+		--cmd "ip netns exec ns0 ping -c1 10.0.42.2" 2>/dev/null
+
+	! $retis collect -c skb,skb-tracking \
+		-p "k:ip_send_check/ftrace" \
+		--cmd "ip netns exec ns0 ping -c1 10.0.42.2" 2>/dev/null
```
atenart, line 116: nit: can we have a comment before each case expecting to fail?

atenart (commented):
That looks quite nice already and I see no major issue, only a few minor ones.

atenart:
I did some tests locally and found a minor thing, which we might not want to "fix": `retis sort` uses its own tracking section which makes non-skb window events having a non-0 skb:

```
33509760677600 (11) [ping] 56285 [k] __ip_local_out #1e7a1961b6e0ffff8db0cce039c0 (skb ffff8db0c96e1c00) n 0
  192.168.99.17 > 1.1.1.1 tos 0x0 ttl 64 id 39730 off 0 [DF] proto ICMP (1)... (truncated or incomplete packet)
  ns 0x7/4026531833
  ↳ 33509760689550 (11) [ping] 56285 [k] ip_send_check #1e7a1961b6e0ffff8db0cce039c0 (skb ffff8db0c96e1c00) n 1
  ↳ 33509760703343 (11) [ping] 56285 [kr] __ip_local_out #1e7a1961b6e0ffff8db0cce039c0 (skb ffff8db0c96e1c00) n 2
      192.168.99.17 > 1.1.1.1 tos 0x0 ttl 64 id 39730 off 0 [DF] len 84 proto ICMP (1) type 8 code 0
      ns 0x7/4026531833
      ct_state NEW status 0x100 icmp orig [192.168.99.17 > 1.1.1.1 type 8 code 0 id 35619] reply [1.1.1.1 > 192.168.99.17 type 0 code 0 id 35619] zone 0 mark 0
```

atenart:
Another thing I noticed while doing tests locally is a function can consume an skb. Tracking then ends before the end of the execution and (at least) the kretprobe won't be linked to prior ftrace events. Eg.

```
$ retis collect -p __dev_queue_xmit/ftrace -p netdev_core_stats_inc
$ retis sort
...
33458329274772 (11) [avahi-daemon] 1645 [k] __dev_queue_xmit #1e6e1fd4c994ffff8db0cf4063c0 (skb ffff8db0c95e6500) n 0
  172.16.42.1.5353 > 224.0.0.251.5353 tos 0x0 ttl 255 id 41052 off 0 [DF] len 201 proto UDP (17) len 173
  ↳ 33458329283995 (11) [avahi-daemon] 1645 [k] netdev_core_stats_inc #1e6e1fd4c994ffff8db0cf4063c0 (skb ffff8db0c95e6500) n 1

33458329287661 (11) [avahi-daemon] 1645 [kr] __dev_queue_xmit
  172.16.42.1.5353 > 224.0.0.251.5353 tos 0x0 ttl 255 id 41052 off 0 [DF] len 201 proto UDP (17) len 173
```

`__dev_queue_xmit` consumed the skb. This is fine but might worth a word in the documentation?

vlrpl:
> Another thing I noticed while doing tests locally is a function can consume an skb. Tracking then ends before the end of the execution and (at least) the kretprobe won't be linked to prior ftrace events. Eg.
> 
> ```
> $ retis collect -p __dev_queue_xmit/ftrace -p netdev_core_stats_inc
> $ retis sort
> ...
> 33458329274772 (11) [avahi-daemon] 1645 [k] __dev_queue_xmit #1e6e1fd4c994ffff8db0cf4063c0 (skb ffff8db0c95e6500) n 0
>   172.16.42.1.5353 > 224.0.0.251.5353 tos 0x0 ttl 255 id 41052 off 0 [DF] len 201 proto UDP (17) len 173
>   ↳ 33458329283995 (11) [avahi-daemon] 1645 [k] netdev_core_stats_inc #1e6e1fd4c994ffff8db0cf4063c0 (skb ffff8db0c95e6500) n 1
> 
> 33458329287661 (11) [avahi-daemon] 1645 [kr] __dev_queue_xmit
>   172.16.42.1.5353 > 224.0.0.251.5353 tos 0x0 ttl 255 id 41052 off 0 [DF] len 201 proto UDP (17) len 173
> ```
> 
> `__dev_queue_xmit` consumed the skb. This is fine but might worth a word in the documentation?

makes sense. I also added a note for the sort/skb addr

retis/src/core/tracking/bpf/include/skb_tracking.h
```diff
@@ -179,31 +187,56 @@ static __always_inline int track_skb_start(struct retis_context *ctx)
...
+
+	if ((ftrace && ctx->probe_type == KERNEL_PROBE_KPROBE) ||
+	    (ctx->flags & RETIS_F_WINDOW_PASS))
+		new_ref |= FTRACE_WINDOW;
+
+	if (!ref || *ref != new_ref) {
+		if (track_stack_update(ctx->stack_base, new_ref))
+			log_error("While tracking stack. Unable to update the entry");
+		else if (!ref)
+			ref = (u64 *)new_ref;
```
atenart: This would mean returning a reference from a local variable instead of the one in the map.

I think the fix is actually to not return a reference to the map entry from `track_skb_start` as it is not used.

vlrpl: yes definitely, but I was sure I removed the retval. I must have messed with my worktrees

atenart (approved):
Thanks for the hard work Paolo, that's a great addition!

PR #618, opened 2026-03-25
ct: add vtag to orig and reply direction

vlrpl:
Add in the ct_event a per-protocol data. SCTP uses it to carry orig/reply vtags.

### DIFF
--- a/docs/events/ct.md
+++ b/docs/events/ct.md
@@ -27,9 +27,12 @@ for the bitset representing the corresponding values.
 This starts by a protocol specific part. For TCP, UDP and SCTP,
 
 ```none
-{protocol name} ({proto state if any}) orig [{src ip}.{src port} > {dst ip}.{dst port}]
-    reply [{src ip}.{src port} > {dst ip}.{dst port}] mark {mark} labels {labels}
+{protocol name} ({proto state if any}) orig [{src ip}.{src port} > {dst ip}.{dst port} {proto specific}]
+    reply [{src ip}.{src port} > {dst ip}.{dst port} {proto specific}] mark {mark} labels {labels}
 ```
+
+where `{proto specific}` is optional protocol-specific information. For SCTP,
+the verification tags are reported as `vtag {vtag}` in hex format.
 
 For ICMP,
 
--- a/retis-events/src/ct.rs
+++ b/retis-events/src/ct.rs
@@ -53,6 +53,8 @@ pub struct CtSctp {
     pub sport: u16,
     /// Destination port.
     pub dport: u16,
+    /// Verification tag.
+    pub vtag: u32,
 }
 
 #[event_type]
@@ -240,16 +242,18 @@ impl CtEvent {
             (CtProto::Sctp { sctp: sctp_orig }, CtProto::Sctp { sctp: sctp_reply }) => {
                 write!(
                     f,
-                    "sctp ({}) orig [{}.{} > {}.{}] reply [{}.{} > {}.{}] ",
+                    "sctp ({}) orig [{}.{} > {}.{} vtag {:#x}] reply [{}.{} > {}.{} vtag {:#x}] ",
                     conn.proto_state.as_ref().unwrap_or(&"UNKNOWN".to_string()),
                     conn.orig.ip.src,
                     sctp_orig.sport,
                     conn.orig.ip.dst,
                     sctp_orig.dport,
+                    sctp_orig.vtag,
                     conn.reply.ip.src,
                     sctp_reply.sport,
                     conn.reply.ip.dst,
                     sctp_reply.dport,
+                    sctp_reply.vtag,
                 )?;
             }
             _ => (),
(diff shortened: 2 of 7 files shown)


retis/src/collect/collector/ct/bpf/ct_hook.bpf.c
```diff
@@ -65,6 +75,7 @@ struct ct_event {
 	u8 labels[16];
 	u16 zone_id;
 	u8 proto_state;
+	union ct_proto_data proto;
```
atenart, line 78: Not a huge fan of using a union (it has to be `unsafe` from the Rust side and is not well supported) and it adds 64 bits to all ct events while only being used for SCTP. Instead we can have an SCTP specific event section by adding a new `enum ct_sections` member.

docs/events/ct.md
```diff
@@ -27,10 +27,14 @@ for the bitset representing the corresponding values.
...
 ```none
-{protocol name} ({proto state if any}) orig [{src ip}.{src port} > {dst ip}.{dst port}]
-    reply [{src ip}.{src port} > {dst ip}.{dst port}] mark {mark} labels {labels}
+{protocol name} ({proto state if any}) orig [{src ip}.{src port} > {dst ip}.{dst port} {proto specific}]
+    reply [{src ip}.{src port} > {dst ip}.{dst port} {proto specific}] mark {mark} labels {labels}
 ```
 
+where `{proto specific}` is optional protocol-specific information. For SCTP,
+the verification tags are reported as `vtag {vtag}` in hex format: `vtag_orig`
+in the `orig` tuple and `vtag_reply` in the `reply` tuple.
```
atenart: I don't think we refer to the event fields explicitly in the output description, the part about `vtag_orig/replay` could be omitted and it's already clear to which of orig and reply a vtag is related. This could confuse users not knowing the event format.

retis-events/src/ct.rs
```diff
@@ -53,6 +53,10 @@ pub struct CtSctp {
     pub sport: u16,
     /// Destination port.
     pub dport: u16,
+    /// Verification tag (original direction).
+    pub vtag_orig: u32,
+    /// Verification tag (reply direction).
+    pub vtag_reply: u32,
```
atenart: There are already directions at the `CtConnEvent` level (`orig`/`reply`) and it seems both of those two added fields are only valid within the current `CtTupe` direction. AFAIU we should only carry `vtag` here, which would be the orig or reply one depending on the direction.

atenart (changes requested):

retis-events/src/ct.rs
vlrpl: yes, doesn't make much sense. I was changing this, indeed.

retis/src/collect/collector/ct/bpf/ct_hook.bpf.c
vlrpl: I was initially ok with this, but after thinking about it, I'm not 100% sure:

Anonymous unions don't play very well with Rust. Named are still unsafe but fundamentally on the ok side with bindings (bindgen automatically hides whatever boilerplate we'd need). Safety is not really a concern.

In terms of final event size it doesn't change much (we have fixed events), if we're talking about potentially reserved space, makes sense, but I admit I didn't like the idea of having something dedicated to a protocol for a single attribute only (which will probably remain forever as another section for a single attribute). Also placing it in there kept the base/parent structure.
While thinking about it, what about mixed protocol parent/base? With a single section this would probably break that (it would require two). 

As a last remark, this will force to some extent the rust part to either reconcile the sctp additional section after unmarshal_ct() (like the snipped below, parent missing) or pass the section to unmarshal_ct().

```rust
        // Same has to be done for the parent (and possibly orig/reply)
        if let Some(raw_section) = raw_sections
            .iter()
            .find(|s| s.header.data_type as u32 == SECTION_SCTP)
        {
            let raw_sctp = parse_raw_section::<ct_sctp_event>(raw_section)?;

            if let CtProto::Sctp { ref mut sctp } = ct.base.orig.proto {
                sctp.vtag = u32::from_be(raw_sctp.base_vtag[IP_CT_DIR_ORIGINAL]);
            [...]
        }
```

All in all 16 bytes is still a good bargain (unless other protocols starts needing that :)

docs/events/ct.md
vlrpl: let's just skip that. While at it, one thing I was noticing is the base mismatch between the init_tag (dec) and vtags (hex). vtags are hex values in both the packet and here. For init_tag we aligned to tcpdump, but I don't see strong reasons to really do it.
I was tempted to convert that to hex as well. Any preference/opinion?

atenart: Makes sense, let's use hex.

retis/src/collect/collector/ct/bpf/ct_hook.bpf.c
atenart: All right, you convinced me. Keeping things simple for now is good too and there's no (current) issue with using extra bytes. It's also an internal thing so we can update anytime if we run into an issue.

vlrpl: good point (internal detail).
Ok, will keep this, then.

atenart (approved):
LGTM, thanks!

PR #619, opened 2026-03-31
Bump btf-rs & related improvements

atenart:
Bumps btf-rs to the upcoming 2.0 release and take advantage of the new features. Also improve BTF handling in Retis wrt to performances.

The decision on whether or not to use the cache backend in Retis is up for discussion, this was done in the PR but can be removed.

This is marked as an RFC until btf-rs 2.0 is released, in the meantime having a consumer of the new features and improvements can be useful to test them.


atenart:
Runtime tests are failing on older distros because `Backend::Mmap` isn't supported on older kernels and we need to use `Backend::Mmap` as a fallback. I'll fix this if we go in this direction.

atenart:
Removing RFC as `btf-rs` 2.0 is now officially released and available.

vlrpl (approved):
thanks for this work!
I manually triggered some of functional tests (it seems for some reason they were never executed, or the reference of the runs got kind of lost).
Approving this in the meantime waiting for them to complete.

PR #622, opened 2026-04-21
CI migration

vlrpl:
As noticed by @atenart, [Cirrus CI will shut down effective Monday, June 1, 2026](https://cirruslabs.org/).

Initially, we targeted all the CI jobs to Cirrus in order to have a unique place to maintain, but given we already decided to introduce some actions for the "-next" image and some other checks (e.g. stale checks), we can safely start migrating the basic tests/builds to GH actions, even though the functional tests may end up being run on some external service. 

Update: this also includes runtime tests


Makefile
```diff
@@ -210,11 +210,10 @@ endef
...
 $(foreach tgt,check clippy,$(eval $(call analyzer_tmpl,$(tgt))))
 
-lint-rust:
+fmt-rust:
 	$(call build, fmt --check, check format)
-	# No need to run `check` in addition to `clippy`, it's a superset.
-	$(call build, clippy $(if $(VERBOSITY),,--quiet), running clippy)
-	$(call build, clippy $(if $(VERBOSITY),,--quiet) -F benchmark, running clippy on benchmarks)
+
+lint-rust: fmt-rust clippy
```
atenart, line 217: This makes the default `lint-rust` output more verbose when running clippy as `VERBOSE` is not handled in the `clippy` target. I do prefer the current behavior but not sure it's worth the effort to fix the PR?

vlrpl: you're right. I'll change it. I'm going to do the below. If cargo logs are not very meaningful with RA=1 (RA is a special different case), but if someone explicitly wants them it's ok. This will make easy to silence stuff:

```diff
diff --git a/Makefile b/Makefile
index f77259d08..162666076 100644
--- a/Makefile
+++ b/Makefile
@@ -202,10 +202,10 @@ pytest: pytest-deps
        cd retis-events && tox

 define analyzer_tmpl
-  $(1): CARGO_CMD_OPTS ?= $(if $(filter 1,$(RA)),--quiet --message-format=json --all-targets --keep-going,)
+  $(1): CARGO_CMD_OPTS ?= $(if $(filter 1,$(RA)),--message-format=json --all-targets --keep-going,)
   $(1): PRINT +=$(if $(filter 1,$(RA)),>/dev/null,)
   $(1):
-       $$(call build,$$(@), running $$@)
+       $$(call build,$$(@) $(if $(VERBOSITY),,--quiet), running $$@)
 endef

 $(foreach tgt,check clippy,$(eval $(call analyzer_tmpl,$(tgt)))) 
 ```

atenart: I think you also need to fix the `RA` help: `When $(RA) is used $(V) becomes ineffective.`

atenart (approved):
Thanks for taking care of this!

PR #623, opened 2026-04-29
builder: set verifier log level when tracing

vlrpl:
Set the log level for programs in open object to 2 when retis log level is set on tracing. This would result in increased logs providing way more information than the default behavior.

### DIFF
--- a/retis/src/helpers/logger.rs
+++ b/retis/src/helpers/logger.rs
@@ -134,6 +134,24 @@ impl log::Log for Logger {
     }
 }
 
+/// Set verifier log level on all programs in an open object.
+///
+/// The log level is a bitmask defined in include/linux/bpf_verifier.h:
+///   BPF_LOG_LEVEL1 (1): log each instruction's index and mnemonic, but
+///                        discard logs from successfully explored branches.
+///   BPF_LOG_LEVEL2 (2): additionally log the full register state at every
+///                        instruction, precision tracking details, and
+///                        preserve logs from all explored paths.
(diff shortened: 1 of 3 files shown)


retis/src/helpers/logger.rs
```diff
@@ -134,6 +134,13 @@ impl log::Log for Logger {
     }
 }
 
+/// Set verifier log level on all programs in an open object.
+pub(crate) fn set_libbpf_rs_prog_log_level(open_obj: &mut libbpf_rs::OpenObject) {
+    if log::max_level() == LevelFilter::Trace {
+        open_obj.progs_mut().for_each(|mut p| p.set_log_level(2));
```
atenart, line 151: Out of curiosity what the different values (> 0) mean? Why `2` specifically?

vlrpl: Sorry, this fell through the cracks.
Consistency with the log level. I guess 1 is what we see now on failure (libbpf should instrument it and reload on failure). 2 includes all the instructions. Very verbose.
There should be 4 (stats I honestly never looked at) and 8 (disable rotation in the buffer) as well, but they are unrelated to 1 and 2.

atenart: Can you add a comment for this? (Is there a link to a doc?)

vlrpl: I don't think any doc is available. I'll take a look, otherwise I'll add a small comment with no links

atenart (approved):
LGTM, thanks!

PR #626, opened 2026-05-27
Bump deps

atenart:
Bump all dependencies, as a prep work for the next release. This includes the updates of:
- pcap-file.
- pyo3.

The Fedora runtime CI is also upgraded to 44.

### DIFF
--- a/retis/Cargo.toml
+++ b/retis/Cargo.toml
@@ -42,7 +42,7 @@ libbpf-sys = "1.6"
 libc = "0.2"
 log = { version = "0.4", features = ["std"] }
 memoffset = "0.9"
-nix = { version = "0.30", features = ["feature", "mount", "time", "user"] }
+nix = { version = "0.31", features = ["feature", "mount", "time", "user"] }
 once_cell = "1.15"
 ovs-unixctl = { version = "0.1.0" }
 pager = "0.16"
@@ -51,18 +51,18 @@ pcap-file = { git = "https://github.com/courvoif/pcap-file.git", rev = "e60e2f9b
 pest = "2.7"
 pest_derive = "2.7"
 plain = "0.2"
-rbpf = {version = "0.3", optional = true}
+rbpf = {version = "0.4", optional = true}
 regex = "1.7"
 retis-derive = {version = "1.5", path = "../retis-derive"}
 serde = {version = "1.0", features = ["derive"]}
 serde_json = "1.0"
 serde_with = "3.0"
 serde_yaml = "0.9"
-signal-hook = "0.3"
+signal-hook = "0.4"
 termcolor = "1.3"
 time = { version = "0.3", features = ["formatting", "macros"] }
 thiserror = "2.0"
-schemars = "0.9"
+schemars = "1.2"
 
 [build-dependencies]
 bindgen = "0.72"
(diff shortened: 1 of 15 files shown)


retis/Cargo.toml
```diff
@@ -47,7 +47,7 @@ once_cell = "1.15"
 ovs-unixctl = { version = "0.1.0" }
 pager = "0.16"
 pcap = "2.2"
-pcap-file = { git = "https://github.com/courvoif/pcap-file.git", rev = "e60e2f9b614812360fdf6909b90a8b05283adb05" }
+pcap-file = "=3.0.0-rc.2"
```
amorenoz: FYI: https://github.com/courvoif/pcap-file/pull/74

atenart: Nice! Let's wait for the official `3.0.0` release then.

atenart: As discussed, this will be tackled in a dedicated PR later.

atenart:
Removed the `pcap-file` update, which will be handled separately.

amorenoz:
> Removed the `pcap-file` update, which will be handled separately.

https://github.com/retis-org/retis/pull/634

atenart:
Had to include a new fix for the pytest target due to a change in the latest version of Tox.

tests/next/pcap.sh
```diff
@@ -9,7 +9,7 @@ pcap_tcp_cc() {
 	$retis collect -o \
 		-f "tcp port 80 or arp" \
 		-p net:netif_rx -p net:net_dev_start_xmit \
-		--cmd "ip netns exec ns0 socat - TCP:10.0.42.2:80"
+		--cmd "ip netns exec ns0 socat - TCP:10.0.42.2:80; sleep 1"
```
amorenoz, line 12: Ugh, I don't like putting sleeps. It just feels pushing the flakiness a bit, but at some point we'll hit it again.

In the past I've used socat's `linger` option to achieve something similar. Have you considered it?

vlrpl: looks a better alternative. If does the trick, we should probably consider this for the other tests using socat+sleep

atenart: I agree `sleep` is not perfect however using `SO_LINGER` will only fix our issues partially. The connection should be closed on exit but Retis might have events "in flight", which will be dropped and not reported.

I'm experimenting with fixing both issues properly (for later), but that later one turn out to be a bit challenging. In the meantime I think sleeping is probably the best.

atenart:
Updated the deps with today's available versions (first commit) as this is not merged yet.

atenart:
Added a fix for Clippy warnings introduced with Rust 1.97.0.

atenart:
Rebased on top of main to run the CI with the latest version.

tests/next/pcap.sh
vlrpl: makes sense. One way (not really a one liner-like solution :))  would be implementing some kind of mechanism like wait until and event gets received (maybe using python pp)

atenart: That's an option. I was more looking at a solution at the Retis level as this not only affects tests but also users. If that doesn't work that can be something to try though.

vlrpl (approved):
lgtm, thanks!

PR #627, opened 2026-05-27
runtime ci: trigger on PR synchronize events as well

atenart:
If the job is only triggered on for 'labeled' events, it won't run again when the pull request is updated. Also trigger the jobs if the PR is updated. The job then only runs if the run-functional-tests labeled is present.


vlrpl (approved):
of course, it was all good during tests except for the only thing untested :)
Thanks!

PR #628, opened 2026-05-27
ci: next: unify sleep delays

atenart:
Use a 1 second everywhere (as already done in the skb tests) waiting for all events to be captured when using the --cmd flag during collection.

This fixes the flaky rotation test too.

### DIFF
--- a/tests/next/rotation.sh
+++ b/tests/next/rotation.sh
@@ -7,8 +7,9 @@ rotation_by_size() {
 
 	ip netns exec ns1 socat TCP-LISTEN:80 /dev/null &
 	$retis collect -o --out-rotate 1MB \
+		--stop-after 8000 \
 		-f "host 10.0.42.1" \
-		--cmd "head -c10M /dev/zero | ip netns exec ns0 socat - TCP:10.0.42.2:80; sleep 1"
+		--cmd "ip netns exec ns0 socat /dev/zero TCP:10.0.42.2:80"
 
 	# Rotation file names should be used.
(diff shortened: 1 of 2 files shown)


tests/next/rotation.sh
```diff
@@ -8,7 +8,7 @@ rotation_by_size() {
 	ip netns exec ns1 socat TCP-LISTEN:80 /dev/null &
 	$retis collect -o --out-rotate 1MB \
 		-f "host 10.0.42.1" \
-		--cmd "head -c10M /dev/zero | ip netns exec ns0 socat - TCP:10.0.42.2:80; sleep 0.1"
+		--cmd "head -c10M /dev/zero | ip netns exec ns0 socat - TCP:10.0.42.2:80; sleep 1"
```
vlrpl: I'm fine with the change.
Just a random thought. If the number of events is stable across kernels, and the env + filter are not noisy, we could consider expected events as well:
`--stop-after <COUNT>`

and keep the sleep just as a timeout (although for that we could use `timeout(1)`).

atenart: That's a great idea, let me update the PR

atenart: done

vlrpl: just a note unrelated to these changes, we still have a failure:

https://github.com/retis-org/retis/actions/runs/26520189528/job/78108687487#step:9:673

vlrpl (approved):
lgtm, thanks!

PR #629, opened 2026-05-27
ci/next: some fixes

vlrpl:
- The for loop masks the failure of non ending tests. They will still be reported as failing, but the makefile will not trigger the CI failure as the exit value is successful if the last test is successful
- fix flaky pcap test due to a spurious event

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,10 +158,12 @@ else
 endif
 
 functional-tests:
+	ret=0; \
 	for script in tests/next/*.sh; do \
 		[ "$$LIST_TESTS" = "1" ] && echo "$$script:"; \
-		$$script; \
-	done
+		$$script || ret=$$?; \
+	done; \
+	exit $$ret
 
 functional-tests-list: export LIST_TESTS=1
 functional-tests-list: functional-tests
--- a/tests/next/drop.sh
+++ b/tests/next/drop.sh
@@ -5,6 +5,9 @@ source $(dirname $0)/include/helpers.sh
 drop_sanity() {
 	two_ns
 
+	# trigger ARP resolution to avoid failures on older kernels
+	ip netns exec ns0 ping -w 1 -c 1 10.0.42.2
+
 	$retis collect -o -c skb-drop,skb,dev -f tcp \
 		--cmd 'ip netns exec ns0 socat -T1 - TCP:10.0.42.2:443'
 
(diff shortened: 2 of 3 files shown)


vlrpl:
the remaining flaky tests are fixed in #628

vlrpl:
there is one last last :) test failing: `drop_sanity()` on c8s.
Apparently, the test expects one single drop, but on c8s we receive two. Didn't look at the code but it seems the NOT_SPECIFIED is generated by a clone which on more recent kernel goes through the consume path:
 
```
[root@localhost vagrant]# ./target/debug/retis print /tmp/retis-test-RxzE/retis.data
914717707774
  collected with Retis v1.7.0+v1.6.0-cce0d31ed189-dirty
  on machine 4.18.0-553.6.1.el8.x86_64 #1 SMP Thu May 30 04:13:58 UTC 2024 x86_64
  cmdline /vagrant/target/debug/retis collect -o -c skb-drop,skb,dev,skb-tracking -f tcp --cmd ip netns exec ns0 socat -T1 - TCP:10.0.42.2:443

914727680602 (0) [socat] 40118 [tp] skb:kfree_skb #d4fa04ba5affff951706e6e400 (skb ffff95175ab78700) drop (reason NOT_SPECIFIED)
  10.0.42.1.51280 > 10.0.42.2.443 tos 0x0 ttl 64 id 44842 off 0 [DF] len 60 proto TCP (6) flags [S] seq 2345515528 win 29200 [mss 1460,sackOK,TS val 1814081518 ecr 0,nop,wscale 7]
  if 2 (veth01)

914727723382 (0) [socat] 40118 [tp] skb:kfree_skb #d4fa04ba5affff951706e6e400 (skb ffff9518052b2980) drop (reason NO_SOCKET)
  10.0.42.1.51280 > 10.0.42.2.443 tos 0x0 ttl 64 id 44842 off 0 [DF] len 60 proto TCP (6) flags [S] seq 2345515528 win 29200 [mss 1460,sackOK,TS val 1814081518 ecr 0,nop,wscale 7]
  if 2 (veth10) rxif 2
  ```

we should relax that check. @atenart, WDYT?

atenart:
> we should relax that check. @atenart, WDYT?

My understanding is this is the ARP packet (skb is reused), which for some reason fails if no socket is listening on RHEL8 (haven't dig into that). IMO the strict check is good wrt testing the stats command, so maybe we can instead just issue a ping before the actual test to ensure connectivity is all set? (Haven't tested that).

atenart (approved):
LGTM, thanks!

PR #631, opened 2026-06-03
Add fzf wrapping script

vlrpl:
The series includes:
- fix broken pipe handling in collect
- bat sublime syntax file. Most of it has been written with Claude Opus support (I will review it a little more although my sublime syntax file knowledge is close to zero, but feel free to contribute :) 
- introduce iretis.sh

Couldn't decide what the best place for documenting the above was. index.md seemed sane, but feel free to suggest any other.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -269,3 +269,28 @@ Output can then be piped through bat directly:
 $ retis print retis.data | bat --language retis
 ```
 
+## Interactive filtering with fzf
+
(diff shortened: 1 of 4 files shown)


vlrpl:
I think the blocker for this pull request is the coloring part which is not as useful as wrapping fzf.
I'm going to remove it for the time being as I do not have time to spend on coloring at the moment.

Issue #630, opened 2026-06-03
Allow self-rotating retis collection


palonsoro:
Network issues often don't happen at predictable moments when one can start capturing. 

For that reason, `tcpdump` offers options like `-C` and `-W`, which allow creating a rotating buffer of W files of C megabytes each, so that the capture becomes a "self-rotating buffer" and hard drive is never exhausted if the tcpdump capture has to run for too long (older PCAPs are removed to make space for the new ones).

Retis can face the very same problem: We may need to do a retis capture to get insights into a problem that may not occur at predictable moments and just leaving `retis collect` running may result in a very big capture file. Having the same problem means it would make sense to follow the same solution: Add some equivalent to `-C` and `-W` tcpdump options so that you can limit retis capture to W files of C megabytes each (and whenever the limit is reached, the oldest file is removed).

atenart:
We do have support for an equivalent of `tcpdump -C` in the development branch of Retis (added in #564), but not for limiting the number of files created.

Adding a limit to the number of files created makes sense, let's track this in there.

atenart:
@palonsoro please feel free to comment on #632. I also built a custom image for you to test the feature, you can use it with:

```
$ RETIS_TAG=632 ./retis_in_container.sh
```

palonsoro:
Thx @atenart . I dropped a minor comment (sorry for the delay).

Thank you very much for having taken care, as this is going to be extremely helpful in loaded scenarios!

PR #632, opened 2026-06-04
collect: allow limiting the number of rotated output files

atenart:
Adds support for limiting the number of output files when using output rotation. This enables running Retis over long periods of time without filling the disk, e.g. waiting for a once in a while issue to be seen.

This was requested in #630. More explanation in the commit logs.

This is the expected behavior:

```
$ retis collect --out --out-rotate 20MB --out-rotate-count 3
[...]
$ ls retis.data*
retis.data.42
retis.data.43
retis.data.44
$ retis print retis.data.42..       # File name must be explicitly used at post-processing time.
[...]
```

While at it provide a short version of the rotation args, for ease of use. The exact letter to use can be discussed. Since the rotation feature is not yet released this is also the last opportunity to change the long version of those arguments before the release (`--out-rotate`/`-r` and `--out-rotate-count`/`-R`).

### DIFF
--- a/retis-events/src/file/rotate.rs
+++ b/retis-events/src/file/rotate.rs
@@ -40,6 +40,8 @@ pub struct RotateWriter {
     inner: BufWriter<File>,
     // Controls how rotation is done.
     policy: Option<RotationPolicy>,
+    // Optional limit to the number of events file.
+    rotate_count: Option<u32>,
     // Target file name for the output. Will be suffixed following the rotation
     // policy rules.
     target: PathBuf,
@@ -58,6 +60,7 @@ impl RotateWriter {
     pub fn new<P: AsRef<Path>>(
         file: P,
         policy: Option<RotationPolicy>,
+        rotate_count: Option<u32>,
         cmdline: &str,
         monotonic_offset: TimeSpec,
     ) -> Result<Self> {
@@ -71,6 +74,7 @@ impl RotateWriter {
         Ok(Self {
             inner,
             policy,
+            rotate_count,
             target: file.as_ref().to_path_buf(),
             index,
             written,
@@ -136,6 +140,45 @@ impl RotateWriter {
         // Create the new file.
         (self.inner, self.written) = Self::new_file(&self.target, &startup)?;
 
+        // Enforce the output files limit, if any.
+        let limit = match self.rotate_count {
+            Some(limit) => limit,
+            None => return Ok(()),
+        };
+
+        // Check if we should enforce the limit yet.
+        if limit > self.index {
+            return Ok(());
+        }
+
+        let mut target = self.target.clone().into_os_string();
+        target.push(format!(".{}", self.index - limit));
+        fs::remove_file(target)
+    }
+
+    // In case we need to tweak the file names once the collection is done.
+    fn files_fixup(&mut self) -> Result<()> {
+        let range = match self.rotate_count {
+            Some(count) if self.index < count => return Ok(()),
+            Some(count) => count,
+            None => return Ok(()),
+        };
+
+        let mut from = self.index - range;
+        let mut to = 0;
+
+        while to < range {
+            let mut file_from = self.target.clone().into_os_string();
+            let mut file_to = file_from.clone();
+            file_from.push(format!(".{from}"));
+            file_to.push(format!(".{to}"));
+
+            fs::rename(&file_from, &file_to)?;
+
+            from += 1;
+            to += 1;
+        }
+
         Ok(())
     }
 }
@@ -170,9 +213,21 @@ impl Drop for RotateWriter {
             return;
         }
 
+        // In case we have a limit on the number of files generated, rename them
+        // so they start at offset 0.
+        if self.policy.is_some() {
+            if let Err(e) = self.files_fixup() {
+                error!("Could not fixup file names: {e}");
+            }
+        }
+
         info!(
-            "Wrote {} event file(s)",
-            if self.policy.is_none() { 1 } else { self.index }
+            "Wrote {} event file(s){}",
+            if self.policy.is_none() { 1 } else { self.index },
+            match self.rotate_count {
+                Some(count) if self.index > count => format!(", kept {count}"),
+                _ => String::new(),
+            },
         );
     }
 }
--- a/retis-events/src/file/rotate.rs
+++ b/retis-events/src/file/rotate.rs
@@ -1,6 +1,5 @@
 /// # Writer handling file rotation
 use std::{
-    ffi::OsStr,
     fs::{self, File, OpenOptions},
     io::{self, BufRead, BufReader, BufWriter, Read, Seek, SeekFrom, Write},
     ops::Drop,
@@ -8,7 +7,7 @@ use std::{
 };
 
 use anyhow::{anyhow, Result};
-use log::{error, info, warn};
+use log::{error, info};
 use nix::sys::utsname::uname;
 
 use crate::{compat::json, file::guess_version, helpers::time::*, *};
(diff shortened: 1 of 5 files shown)


palonsoro:
It looks great!!!

One dummy suggestion: Would it be possible to default to megabytes instead of failing if no suffix is used? That would make it more similar to tcpdump (but if you feel that current status is more usable, it is ok anyway).

Thanks and regards!

atenart:
> One dummy suggestion: Would it be possible to default to megabytes instead of failing if no suffix is used? That would make it more similar to tcpdump (but if you feel that current status is more usable, it is ok anyway).

That was done on purpose because we intended  to use the same option (`--out-rotate`) regardless of the rotation limit being used. We currently only support rotation by size but we could have other strategies in the future (e.g. by time); it would then be a bit confusing to default to one if no suffix is used IMO. `tcpdump` uses different options for different strategies (`-C`, `-G`) which helps having a default unit.

palonsoro:
> That was done on purpose because we intended to use the same option (`--out-rotate`) regardless of the rotation limit being used. We currently only support rotation by size but we could have other strategies in the future (e.g. by time); it would then be a bit confusing to default to one if no suffix is used IMO. `tcpdump` uses different options for different strategies (`-C`, `-G`) which helps having a default unit.

Understood. It is logical. Then, I'd say it is perfect. Thanks!

retis-events/src/file/rotate.rs
```diff
@@ -41,6 +41,8 @@ pub struct RotateWriter {
     inner: BufWriter<File>,
     // Controls how rotation is done.
     policy: Option<RotationPolicy>,
+    // Controls if there is a limit to the number of event files.
+    rotate_count: Option<u32>,
```
amorenoz, line 44: nit: the word "Controls" is a bit strange here. It's not as `RotationPolicy` that can encode arbitrary logic behind each enum, this is an optional integer. `Optional limit to the number of events file` seems more natural

retis/src/collect/cli.rs
```diff
@@ -104,6 +104,16 @@ Examples of meta filters:
...
     )]
     pub(super) out_rotate: Option<String>,
+    #[arg(
+        long,
+        requires = "out_rotate",
+        value_name = "COUNT",
+        value_parser = clap::value_parser!(u32).range(1..),
+        help = "When rotating the output file (see `--out-rotate`) limit the number of output files created to <COUNT>.
+
+As time goes older files are removed keeping the output files in order. This also means the file name will need to be explicitly provided at post-processing time."
```
amorenoz: Why do we need this again?
Why not always have retis.data.1...N? I don't see any technical limitation and it would be much more user friendly.

atenart: Makes sense, I modified the series to always have files starting with suffix `.0`. To make this work I had to relax a bit how suffixes are handled, but this is safe and actually improves things IMO.

atenart:
CI is failing, fixes are in #626.

atenart:
Rebased on top of main to get the CI fixes.

retis/src/collect/cli.rs
```diff
@@ -98,8 +98,10 @@ Examples of meta filters:
         long,
         requires = "out",
         value_name = "LIMIT",
-        help = "Rotate the output file (see `--out`) once a given limit is reached. The rotation limit can be:
-- A size: <LIMIT> is in bytes and must be suffixed with a size unit (MB, GB). The file names will be <OUT>.X, with X being a number starting at 0 and increasing over time. e.g. '--out-rotate 64MB'.
+        help = "Rotate the output file (see `--out`) once a given limit is reached. The file names will be <OUT>.X, with X being a number starting at 0 and increasing over time.
+
+The rotation limit can be:
+- A size: <LIMIT> is in bytes and must be suffixed with a size unit (MB, GB). e.g. '--out-rotate 64MB'.
```
vlrpl, line 105: not that this is related to the PR, but if feels better (and more aligned to other tools) if it's just M or G (or maybe support M, G, MB, GB)

retis-events/src/file/rotate.rs
```diff
@@ -170,6 +213,14 @@ impl Drop for RotateWriter {
...
+        // so they start at offset 0.
+        if self.policy.is_some() {
+            if let Err(e) = self.files_fixup() {
+                error!("Could not fixup file names: {e}");
+            }
+        }
+
         info!(
             "Wrote {} event file(s)",
             if self.policy.is_none() { 1 } else { self.index }
```
vlrpl: this still results in a non limited version (e.g. sounds a little weird to have "Wrote 42 ..." when it is limited to 3). Is that intentional or we want something like:

```rust
let count = if self.policy.is_none() {
         1
     } else {
         match self.rotate_count {
             Some(c) if self.index > c => c,
             _ => self.index,
         }
     };
     info!("Wrote {count} event file(s)");
```

retis/src/collect/cli.rs
atenart: Some tools have a different option depending on what is the limit (size, lines, etc). We chose to use the same. For this reason we need to provide an unit, otherwise if we support another one later we wouldn't be able to distinguish which one is it.

E.g. if we support both a size limit and a lines limit, what `4M` refers to?

retis-events/src/file/rotate.rs
atenart: Not intentional. I'm wondering if we should just print `wrote 3 files` if we wrote 42 of them, or something like: `wrote 42 files, kept 3`. WDYT? I'm not sure if that is better tbh :)

vlrpl: That's exactly what I was thinking about right after writing the comment, but I wasn't sure about the initial intent. In fact my doubt was you wanted to keep track of the actual index just to make the user aware about the number of discarded files.
I think that probably including both makes sense, especially in the case the writes go over the limit.
It would have a double meaning ("kept" would help checking the correctness of the result, "wrote" gives an idea of how many files were discarded).

retis/src/collect/cli.rs
vlrpl: I see. It's a bit annoying to specify MB, GB. I think that the common case will refer to bytes in the vast majority of the cases, so it looks safe to me if we default M or G to MB, GB. The others (for which we might or might not have an actual implementation/request), like lines, may have to always require the specific selector.

But feel free to keep it as is, if you have a strong preference.

atenart: I get the point but I'm not sure. The likely options for limits are by size, by time but also maybe by number of events. The last one is the one that worries me wrt that proposal.

I think this was discussed in the initial PR too, so maybe there's something to do. I would then rather let you open a dedicated PR so we can discuss things there.

vlrpl: that's ok, as I said, not pushing on this. Opening an hanging, although small, PR doesn't sound compelling :) 
OTOH, I get this doesn't really belong here, but it seemed a good chance to see if we could improve user's "quality of life" in what I consider the largely more used/useful use case WRT files.
There still are some counterarguments to consider, but, yeah, let's not dilute or hold this PR for a detail (at some point they all start to sound like "tabs vs spaces" :)

atenart: I think that's a fair point and for the current situation would improve ux. As said I'm worried this would cause confusion. Another case could be `M` (MB) vs `m` (minutes). At least having the unit makes things less likely to collide (of course things like `mb` vs `m` due to a typo could still be possible).

atenart: Another solution would be to have a per-unit cli option.

vlrpl: > I think that's a fair point and for the current situation would improve ux. As said I'm worried this would cause confusion. Another case could be `M` (MB) vs `m` (minutes). At least having the unit makes things less likely to collide (of course things like `mb` vs `m` due to a typo could still be possible).

agreed, that was what I meant with other things to consider, but I'd rather use "min" over 'm' and 'M' over "MB" (but of course, my assumption that size would be the most used must hold true, and I can't tell for sure).
Having MB is not bad per se, and we could always revisit it at any point in time.

The per-unit option might look a good compromise (at least that was my first impression), but it might end up bloating the pool of options that could already be a little overwhelming for the casual user.
Not sure about it. I guess we can unblock this (lgtm) and defer/move the discussion about this.

atenart: > I guess we can unblock this (lgtm) and defer/move the discussion about this.

Yes, I think that's something we can consolidate later if we get more feedback / after we implement another limit unit / after some time.

vlrpl (approved):
lgtm, thanks!

PR #639, opened 2026-07-06
profiles: generic: improve GRO handling

### LINKED ISSUE #636: Generic profile fails on recent kernels
Using the generic profile fails on a recent Fedora kernel (e.g. 7.0.13-200.fc44.x86_64) with:

```
Error: Could not find a traceable function matching 'tcp6_gro_receive'
```

atenart:
Upstream commit 87737cd76e44 ("gro: inline tcp6_gro_receive()") inlined tcp6_gro_receive, making the generic profile to fail on recent kernels. Instead of just removing it, take the opportunity to improve the GRO handling in the generic profile.

Instead of adding protocol-specific GRO probes, rely on the two GRO entry tracepoints and add a probe on the generic skb_gro_receive helper that is called in every GRO path. This has two major advantages over probes like tcp_gro_receive:
- It gives a hint the packet has been considered for GRO.
- It covers all protocols (e.g. tunnels too).

It's worth noting a GRO'ed TCPv6 packet will still go through a protocol specific probe later (e.g. tcp_v6_rcv).

Fixes #636.


atenart:
CI test are failing, fixes are in #626. I'll rebase once merged.

atenart:
Rebased on top of main to get the CI fixes.

vlrpl (approved):
lgtm, thanks!

PR #640, opened 2026-07-09
Core: Support Debian Kernels

utoni:
This PR adds support for Debian kernel versions i.e. `6.12.95+deb13-amd64`.

### DIFF
--- a/retis/src/core/inspect/kernel_version.rs
+++ b/retis/src/core/inspect/kernel_version.rs
@@ -38,7 +38,7 @@ impl KernelVersion {
 
     /// Parse a version string of the `$(uname -r)` form into a KernelVersion.
     pub(crate) fn parse(version: &str) -> Result<Self> {
-        let mut parts = version.split('.');
+        let mut parts = version.split(['.', '+']);
 
         let major: u32 = parts
             .next()
@@ -55,7 +55,6 @@ impl KernelVersion {
         let patch: u32 = tmp
             .next()
             .ok_or_else(|| anyhow!("Could not get kernel patch version from {version}"))?
-            .trim_end_matches('+')
             .parse()?;
 
         // Build can be in any position of the remaining string, e.g:
@@ -303,6 +302,13 @@ mod tests {
         assert_eq!(version.patch, 12);
         assert_eq!(version.build, Some(1));
         assert_eq!(version.full, "6.4.12-arch1-1");
+
+        let version = KernelVersion::parse("6.12.95+deb13-amd64").unwrap();
+        assert_eq!(version.major, 6);
+        assert_eq!(version.minor, 12);
+        assert_eq!(version.patch, 95);
+        assert_eq!(version.build, None);
+        assert_eq!(version.full, "6.12.95+deb13-amd64");
 
         assert!(KernelVersion::parse("6.2").is_err());
     }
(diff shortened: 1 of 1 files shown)


atenart:
The Python CI failures can be ignored here, those are fixed in #626.

atenart:
Thanks for the contribution! Improving the kernel version parsing makes sense as the `+` in Debian seems to be used as a separator and not just a flag. However I think we can further improve the logic.

With this patch when parsing `6.12.95+deb13-amd64` I think we end up with (in `parse()`):

```rust
let major = 6;
let minor = 12;
let tmp = ["95+deb13", "amd64"];
let minor = 95;
let build = None;  // Only iterates over ["amd64"]
```

We could instead of trimming what's after the `+` in the `minor` parsing logic consider it as a separator so that tmp looks like `["95", "deb13", "amd64"]`. This would make the `build` logic not skip over parts of the version string (here `deb13`). In practice this does not change things much here but IMO that would be better long term and would ease support for more version formats.

I think we can do the following:

```diff
         let mut tmp = parts
             .next()
             .ok_or_else(|| anyhow!("Could not get kernel patch-build version from {version}"))?
-            .split('-');
+            .split(['-', '+']);
         let patch: u32 = tmp
             .next()
             .ok_or_else(|| anyhow!("Could not get kernel patch version from {version}"))?
-            .trim_end_matches('+')
             .parse()?;
```
 This also restrict a little bit what is expected as a version string separator as `char::is_ascii_` helpers can split on `_` which could lead to `x86_64` being considered as the build version (`64`) depending on how the version string is formatted.

Or we can just relax things even more?

```diff
     /// Parse a version string of the `$(uname -r)` form into a KernelVersion.
     pub(crate) fn parse(version: &str) -> Result<Self> {
-        let mut parts = version.split('.');
+        let mut parts = version.split(['.', '-', '+']);
 
         let major: u32 = parts
             .next()
@@ -48,19 +48,14 @@ impl KernelVersion {
             .next()
             .ok_or_else(|| anyhow!("Could not get kernel minor version from {version}"))?
             .parse()?;
-        let mut tmp = parts
-            .next()
-            .ok_or_else(|| anyhow!("Could not get kernel patch-build version from {version}"))?
-            .split('-');
-        let patch: u32 = tmp
+        let patch: u32 = parts
             .next()
             .ok_or_else(|| anyhow!("Could not get kernel patch version from {version}"))?
-            .trim_end_matches('+')
             .parse()?;
 
         // Build can be in any position of the remaining string, e.g:
         // 6.2.0-20-generic or 6.4.12-arch1-1.
-        let build = tmp.find_map(|s| s.parse::<u32>().ok());
+        let build = parts.find_map(|s| s.parse::<u32>().ok());
```

utoni:
I've decided to go with second approach as it simplifies code.
I am still keeping the previous commit in the history, so one may change it again in the future if necessary.

atenart:
> I've decided to go with second approach as it simplifies code. I am still keeping the previous commit in the history, so one may change it again in the future if necessary.

Makes sense, however can you then go with the first suggestion for the first commit then (to avoid skipping part of the string and to avoid matching false positive with `char::is_ascii_` helpers)?

atenart (approved):
Technically the split on `+` should be done at the `tmp` level in the first patch, but since the end result after patch 2 is fine let's not do more ping pong. I've also tested this locally since the CI has extra issues due to the new Rust version released yesterday (this is fixed in #626).

Thanks for the contribution!

PR #644, opened 2026-07-28
collector: netns: allow operating on non-skb probes

atenart:
The netns hook was requiring an skb to be available in the function parameters, but it can also operate on `struct net *` directly. Do something similar to what the dev collector is doing, allowing to operate on both situations.

### DIFF
--- a/retis/src/collect/collector/ns/bpf/netns_hook.bpf.c
+++ b/retis/src/collect/collector/ns/bpf/netns_hook.bpf.c
@@ -39,26 +39,19 @@ static __always_inline struct net *get_net_from_skb(struct sk_buff *skb)
 	return NULL;
 }
 
-DEFINE_HOOK_RAW(
+DEFINE_HOOK(F_GROUPS(RETIS_ALL_FILTERS, RETIS_F_WINDOW_PASS),
 	/* Netns cookies are not available on older kernels. */
 	bool get_cookie = bpf_core_field_exists(struct net, net_cookie);
 	struct netns_event *e;
 	struct sk_buff *skb;
 	struct net *net;
-	u64 cookie;
+	u64 cookie = 0;
 	u32 inum;
 
 	skb = retis_get_sk_buff(ctx);
-	if (!skb || !skb_is_tracked(skb))
+	net = skb ? get_net_from_skb(skb) : get_net_from_parms(ctx);
+	if (!net)
 		return 0;
-
-	net = get_net_from_skb(skb);
-	if (!net) {
-		/* Fallback to parameters. */
-		net = get_net_from_parms(ctx);
-		if (!net)
-			return 0;
-	}
 
 	if (get_cookie)
 		cookie = BPF_CORE_READ(net, net_cookie);
--- a/Makefile
+++ b/Makefile
@@ -219,7 +219,7 @@ lint-rust: fmt-rust clippy
 
 lint-ebpf:
 	$(call out_console,CHECKPATCH,checking eBPF coding style ...)
-	ignore="SPDX_LICENSE_TAG,FILE_PATH_CHANGES,EMAIL_SUBJECT,VOLATILE,MACRO_WITH_FLOW_CONTROL,MACRO_ARG_UNUSED"; \
+	ignore="SPDX_LICENSE_TAG,FILE_PATH_CHANGES,EMAIL_SUBJECT,VOLATILE,MACRO_WITH_FLOW_CONTROL,MACRO_ARG_UNUSED,COMMIT_LOG_LONG_LINE,COMMIT_MESSAGE,NOT_UNIFIED_DIFF"; \
 	base_hash=$$(git merge-base $${BASE_COMMIT:-main} HEAD); \
 	COMMIT_RANGE=$$(git rev-list --no-merges --reverse $${base_hash}..HEAD); \
 	for commit in $$COMMIT_RANGE; do \
--- a/docs/collectors/overview.md
+++ b/docs/collectors/overview.md
@@ -124,15 +124,17 @@ The `nft` collector produces the [nft](../events/nft.md) event section.
 
 ## Network device
 
-The `dev` collector provides information about network devices, either if a
-`struct net_device` is available as part of a probe arguments or through a
-`struct sk_buff` and its reference.
+The `dev` collector provides information about network devices, using the
+reference inside `struct sk_buff` or directly operating on `struct net_device`
+if no `skb` is available in a probed function and in an `ftrace` section.
 
 The `dev` collector produces the [dev](../events/dev.md) event section.
 
 ## Namespace
 
-The `ns` collector retrieves information about namespaces, currently only
-network namespaces.
+The `ns` collector retrieves information about network namespaces, using the
+reference inside `struct sk_buff` (through `struct net_device` or `struct sock`)
+or directly operating on `struct net` if no `skb` is available in a probed
+function and in an `ftrace` section.
 
 The `ns` collector produces the [netns](../events/netns.md) event section.
--- a/retis/src/collect/collector/dev/bpf/dev_hook.bpf.c
+++ b/retis/src/collect/collector/dev/bpf/dev_hook.bpf.c
@@ -12,7 +12,7 @@ struct dev_event {
 	u32 iif;
 } __binding;
 
-DEFINE_HOOK_RAW(
+DEFINE_HOOK(F_GROUPS(RETIS_ALL_FILTERS, RETIS_F_WINDOW_PASS),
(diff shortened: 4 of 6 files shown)


retis/src/collect/collector/ns/bpf/netns_hook.bpf.c
```diff
@@ -45,21 +45,22 @@ DEFINE_HOOK_RAW(
...
-		return 0;
+	if (skb) {
+		if (!skb_is_tracked(skb))
+			return 0;
 
-	net = get_net_from_skb(skb);
-	if (!net) {
-		/* Fallback to parameters. */
+		net = get_net_from_skb(skb);
+	} else {
```
vlrpl: aren't we doing this unconditionally?
Meaning, each path (related to any packet) hitting this would generate the event?

atenart: Right and this is also true for the dev collector. I think the missing piece to get the correct behavior is to use `DEFINE_HOOK(F_GROUPS(RETIS_ALL_FILTERS, RETIS_F_WINDOW_PASS)` instead of the raw version?

vlrpl: I guess so. The thing would be then that their respective non skb probes have to be "enclosed" in `/ftrace`. Probably documenting it would be nice as this is the main tradeoff compared to the previous design.
That said, I wonder if their known_types() (non-skb) still make sense.

atenart: > The thing would be then that their respective non skb probes have to be "enclosed" in `/ftrace`. Probably documenting it would be nice as this is the main tradeoff compared to the previous design.

I'll improve the doc as it could be more explicit (and is I think currently wrong).

I actually think this is not a tradeoff and the expected behavior. Except for special rare cases, the raw hook should not be used and all events should be linked to an skb. Unless ftrace is used now. In that regard the dev collector was broken and the netns one could have moved to a normal hook (avoiding to open code the same logic).

> That said, I wonder if their known_types() (non-skb) still make sense.

Yes that's a known thing when the known types include `struct sk_buff` but for correctness we should list all types a hook can operate on.

atenart: > > That said, I wonder if their known_types() (non-skb) still make sense.
> 
> Yes that's a known thing when the known types include `struct sk_buff` but for correctness we should list all types a hook can operate on.

You might have a point though: could we entirely remove the known types logic? Maybe.

vlrpl: If we think hooks can act on non skb types only when in-window, it makes sense (of course, for the skb-only case we still have to enforce the skb presence). For non-skb outside the window (in non raw hooks) as of now we'd still allow attaching the probe without generating anything (not that this is a bad thing).

atenart: > If we think hooks can act on non skb types only when in-window, it makes sense

Before we had the ability to link non-skb probes to the skb tracking (ftrace) we were in a weird place. We had hooks able to handle non-skb probes, but the generated events were not linked to a packet. Also those hooks were not behaving the same way (`dev` generated events, `ns` not). It also comes from a time were we were not sure how to handles those probes.

IMO (unless there's a good reason not to do so) the collectors should only report an event if it can be linked to a tracked skb. Either because the probed function has an skb (normal operation), or special logic can be applied to get the info (raw hook). The ftrace feature now provides the missing link when that is not possible; eg. linking `fn(struct net_device *dev)` to a tracked skb.

This makes me wonder if we should not make all non-raw hook be called iif:

```c
if ((skb && RETIS_TRACKABLE(ctx)) || (!skb && ctx->flags & RETIS_F_WINDOW_PASS))
        hook(...);
```

To make things more consistent and always require the skb to be tracked if present, also avoiding manual filtering in `dev` and `ns`.

In addition we might want to clear `FTRACE_WINDOW` if the skb is not tracked within an ftrace window (skb changed)?

> For non-skb outside the window (in non raw hooks) as of now we'd still allow attaching the probe without generating anything (not that this is a bad thing).

Right. This is a pre-existing issue (`dev` collector could let dev-only functions to be probed without generating events as the hook was filtering on the skb tracking). I think this is worth fixing but I feel like this needs some thinking to get things right. There might be some consolidation to do also wrt the above comment. I can take a look during next dev window.

vlrpl: I totally agree. Looks something that requires some thoughts and targeted for 1.8

atenart: > In addition we might want to clear `FTRACE_WINDOW` if the skb is not tracked within an ftrace window (skb changed)?

This is already the case actually :)

atenart: > ```c
> if ((skb && RETIS_TRACKABLE(ctx)) || (!skb && ctx->flags & RETIS_F_WINDOW_PASS))
>         hook(...);
> ```

I've taken into account the comments and also included the last patch which implements the above. This should not change the existing behavior and IMO is making things more clear. Feel free to push back if you'd prefer to discuss this later.

> > > For non-skb outside the window (in non raw hooks) as of now we'd still allow attaching the probe without generating anything (not that this is a bad thing).
> > 
> > Right. This is a pre-existing issue (`dev` collector could let dev-only functions to be probed without generating events as the hook was filtering on the skb tracking). I think this is worth fixing but I feel like this needs some thinking to get things right. There might be some consolidation to do also wrt the above comment. I can take a look during next dev window.
>
> I totally agree. Looks something that requires some thoughts and targeted for 1.8

:+1:  I added an issue to track this for 1.8, https://github.com/retis-org/retis/issues/645

vlrpl: > > ```c
> > if ((skb && RETIS_TRACKABLE(ctx)) || (!skb && ctx->flags & RETIS_F_WINDOW_PASS))
> >         hook(...);
> > ```
> 
> I've taken into account the comments and also included the last patch which implements the above. This should not change the existing behavior and IMO is making things more clear. Feel free to push back if you'd prefer to discuss this later.
> 

I'm not against the current way, but I think we went intentionally the previous way (avoiding checks inside the hook). TBF, doesn't make much difference.
Maybe in the future we might want to have some flag that not every non-raw hook might consume, but I'm not a fan of keeping logic that will never be used just because there's a small chance.

That being said, I would like to think about potential advantages or pitfalls mixing the ovs use case with the in_window scenario.
E.g., assuming we extend `/ftrace` to avoid sending kretprobes  events (let's say some internal way to instrument them without user facing events), I can imagine we could rely on simple `FALL(F_WINDOW)` instead of raw hooks.
This could potentially simplify the embedded stack tracking logic in ovs hooks. But determining whether this is worth it or it is feasible at all, looks v1.8 material.
WDYT if we target the last patch for that (once the above is clear)?

> > > > For non-skb outside the window (in non raw hooks) as of now we'd still allow attaching the probe without generating anything (not that this is a bad thing).
> > > 
> > > 
> > > Right. This is a pre-existing issue (`dev` collector could let dev-only functions to be probed without generating events as the hook was filtering on the skb tracking). I think this is worth fixing but I feel like this needs some thinking to get things right. There might be some consolidation to do also wrt the above comment. I can take a look during next dev window.
> > 
> > 
> > I totally agree. Looks something that requires some thoughts and targeted for 1.8
> 
> 👍 I added an issue to track this for 1.8, #645

atenart: > I think we went intentionally the previous way

For the normal behavior yes but then we introduced raw hooks because sometimes we need more control.

> That being said, I would like to think about potential advantages or pitfalls mixing the ovs use case with the in_window scenario. E.g., assuming we extend `/ftrace` to avoid sending kretprobes events (let's say some internal way to instrument them without user facing events), I can imagine we could rely on simple `FALL(F_WINDOW)` instead of raw hooks.

If we want to support the `skb && cfg->flags & RETIS_F_WINDOW_PASS)` case IMO a raw hook is the best solution, as this is the opposite of what we generally want for hooks; meaning there is special handling in place for some good reason. (We might provide helpers for better readability).

> This could potentially simplify the embedded stack tracking logic in ovs hooks. But determining whether this is worth it or it is feasible at all, looks v1.8 material. WDYT if we target the last patch for that (once the above is clear)?

That's fine, especially as this is not changing the behavior atm. Let me remove the patch from this PR and open a dedicated once the 1.8 dev cycle begins.

vlrpl (approved):
lgtm, thanks for this!

PR #646, opened 2026-07-31
Bump deps

atenart:
Bumps deps before the release, especially libbpf-rs/cargo/sys. Nothing scary, a single adjustment needed because of a new lint and the toggling of the 'full' feature for sync which should have been the case already.


vlrpl (approved):
lgtm, thanks!

PR #649, opened 2026-08-10
Post v1.7

vlrpl:
- Added small improvement to RELEASE.md
- Kick off v1.8.0 development

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -21,6 +21,8 @@
    1. `podman run --rm --env MATURIN_PYPI_TOKEN=$(cat ~/.my_pypi_token)
        -v $(pwd):/io:z ghcr.io/pyo3/maturin publish -m retis-events/Cargo.toml
        -F python-lib`.
+       It is best practice to test your package on TestPyPI before publishing it to the
+       official index. For testing purposes, append `-r testpypi` to the options.
 
 ## After the release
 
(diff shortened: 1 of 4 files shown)


RELEASE.md
```diff
@@ -21,6 +21,8 @@
    1. `podman run --rm --env MATURIN_PYPI_TOKEN=$(cat ~/.my_pypi_token)
        -v $(pwd):/io:z ghcr.io/pyo3/maturin publish -m retis-events/Cargo.toml
        -F python-lib`.
+       It is best practice to test your package on TestPyPI before publishing it to the
+       official index. For testing purposes, use `--env MATURIN_REPOSITORY_URL="[https://test.pypi.org/legacy/](https://test.pypi.org/legacy/)"`.
```
atenart: I think we can use `-r testpypi` too, which has the advantage of keeping the exact URL hidden in case it changes one day (e.g. legacy API is not used anymore).

(Or `MATURIN_REPOSITORY` if you prefer the env variable way).

retis/Cargo.toml
```diff
@@ -13,7 +13,7 @@ homepage.workspace = true
 edition.workspace = true
 
 [package.metadata.misc]
-release_name = "olive oil"
+release_name = "Tempranillo"
```
atenart: nit: we never used an upper case in the name

vlrpl: I knew that, but didn't seem there was a strong reason to avoid it. Changed it to lowercase, though

atenart (approved):
Thanks @vlrpl! Let the 1.8 dev cycle begin!

PR #650, opened 2026-08-11
improve the hook filtering logic

atenart:
This was discussed in https://github.com/retis-org/retis/pull/644#discussion_r3673105571 and we decided to delay the discussion in the next dev window and in a dedicated PR; this is it.

___

Under normal circumstances hooks should generate events if and only if the events can be linked to a packet (at runtime or post-processing time), unless if in an ftrace section. For special cases, raw hooks are available. We should not have to know all the filtering logic when defining hooks.

This makes the above handled in the core logic: if an skb is available, it must be tracked. If no skb is available, we only let the hook run if in an ftrace window as this is the piece linking those events to a tracked skb. Hooks are still responsible for checking the arguments they are operating on are available and valid (eg. an skb).

Thanks to this, remove special tracking checks in the skb-drop, dev and ns collectors. Also fix the kprobe:ovs_execute_actions hook that was not operating on meta filtered skbs.

This should not change the existing filtering and hooks behavior but makes writing hooks more clear, easier to maintain and to extend.

### DIFF
--- a/retis/src/core/probe/kernel/bpf/include/common.h
+++ b/retis/src/core/probe/kernel/bpf/include/common.h
@@ -47,45 +47,17 @@ struct {
 } stack_map SEC(".maps");
 
 
-/* All bits in the mask. */
-#define F_ALL(fmask)	((ctx->flags & (fmask)) == (fmask))
-/* Always true. Used by raw hooks. */
-#define F_ALWAYS	1
-
-/* OR of AND-groups: each argument is a bitmask that must be fully set (AND),
- * and the groups are OR'd together. Up to 4 groups are supported.
- *
- *   F_GROUPS(A)       ->  F_ALL(A)
- *   F_GROUPS(A, B)    ->  F_ALL(A) || F_ALL(B)
- */
-#define _F_GROUPS1(a)		(F_ALL(a))
-#define _F_GROUPS2(a, b)	(F_ALL(a) || F_ALL(b))
-#define _F_GROUPS3(a, b, c)	(F_ALL(a) || F_ALL(b) || F_ALL(c))
-#define _F_GROUPS4(a, b, c, d)	(F_ALL(a) || F_ALL(b) || F_ALL(c) || F_ALL(d))
-
-#define _F_GROUPS_SEL(_1, _2, _3, _4, N, ...)	N
-#define F_GROUPS(...)						\
-	_F_GROUPS_SEL(__VA_ARGS__,				\
-		      _F_GROUPS4, _F_GROUPS3,			\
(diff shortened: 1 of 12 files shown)


retis/src/core/probe/kernel/bpf/include/common.h
```diff
@@ -47,69 +47,47 @@ struct {
...
+#define __DEFINE_NAMED_HOOK(hook_name, is_raw, statements)			\
 	SEC("ext/hook")								\
 	int hook_name(struct retis_context *ctx, struct retis_raw_event *event) \
 	{									\
 		/* Let the verifier be happy */					\
 		if (!ctx || !event)						\
 			return 0;						\
-		if (!(fexpr))							\
-			return 0;						\
+		if (!is_raw) {							\
```
vlrpl, line 75: Just a random idea.
I know that dead code elimination will get rid of that so the branch is not a runtime overhead, but still dead code elimination consumes cycles on load (probably negligible, but still).
What about:

```
#define __DEFINE_NAMED_HOOK(hook_name, is_raw, statements)
	__DEFINE_NAMED_HOOK##is_raw(hook_name, statements)

#define __DEFINE_NAMED_HOOK0(hook_name, statements)

#define __DEFINE_NAMED_HOOK1(hook_name, statements)
```

vlrpl (commented):
in general I'm ok with the change.
It's a little more permissive, but we should be fine (as long as all hooks do not assume the existence of their target type (which should not be the case).
I have a small non-mandatory suggestion, though.

retis/src/core/probe/kernel/bpf/include/common.h
atenart: Isn't dead code elimination happening at compilation time?

vlrpl: it doesn't for multiple cases, but in this particular one, yes, almost certainly (didn't really check, but still) the compiler will take care of it (I was probably low on coffee :)

Well, consider the suggestion still valid if you don't want to save a conditional, otherwise feel free to resolve the conversation

atenart: I'm not sure what would be the difference with the existing then, besides having two definitions?

vlrpl: Duplication can be avoided keeping the approach (not sure the macro would be nice anyway), but tbh, nothing other than an alternative approach remained :) that's why I was saying to feel free to resolve the conversation based on your preference.

vlrpl (approved):
lgtm, thanks!

PR #653, opened 2026-08-12
Avoid using fake Ethernet header for common cases

atenart:
Retis adds a fake Ethernet header when the L2 information is not available (eg. in early Tx paths). We can instead only capture the data starting at L3 and handle it that way for IPv4 and IPv6, which has numerous advantages (CPU cycles, no fake info, ...).

Also improve the related code. Details in the individual commits messages and code.


retis-events/src/packet.rs
```diff
@@ -1112,11 +1052,63 @@ mod tests {
             "Oh7dUvtE6h3Fhm4TCABFAgBEAABAAECE0jEKACoBCgAqAoVME8QAAAAAAAAAAAEAACTFizMuAAGgAAAK///40kCcAAwABgAFAACAAAAEwAAABA==",
             &mut buf,
         ).unwrap();
-        let raw = RawPacket(buf);
+        let packet = PacketEvent {
+            len: 0,
+            capture_len: 0,
+            data: RawPacket(buf),
```
vlrpl, line 1058: didn't look into this, but does it still make sense to expose it?

retis/src/bindings/skb_hook_uapi.rs
```diff
@@ -1,4 +1,4 @@
-/* automatically generated by rust-bindgen 0.72.1 */
+/* automatically generated by rust-bindgen 0.71.1 */
```
vlrpl, line 1: any reason to downgrade the version here?

retis-events/src/compat/compat.rs
```diff
@@ -36,6 +36,7 @@ const FIXUPS: &[&[CompatFixup]] = &[
         Add("startup/cmdline", CompatValue::String("unknown")),
         Move("ct/tcp_state", "ct/proto_state"),
         Move("ct/parent/tcp_state", "ct/parent/proto_state"),
+        Add("packet/kind", CompatValue::String("ethernet")),
```
vlrpl, line 39: this will break compat on the current version. It requires a new entry for >= 1.8.0

retis/src/process/cli/pcap.rs
```diff
@@ -149,7 +154,11 @@ impl EventParser {
             false => {
                 blocks.push(
                     InterfaceDescriptionBlock {
-                        linktype: DataLink::ETHERNET,
+                        linktype: match kind {
+                            PacketKind::Ethernet | PacketKind::FakeEthernet => DataLink::ETHERNET,
+                            PacketKind::Ipv4 => DataLink::IPV4,
```
vlrpl, line 159: I guess caching based on probe name only can become a problem now when a probe captures heterogeneous Lx packets (e.g. net:netif_receive_skb from an l2 and l3 device like a tun).
This could potentially decode part of the packets with the wrong linktype.

In the scenario above I reproduced it:
3 ethernet
3 ipv4

The generated pcap assigned all packets with linktype 129:
0  129  raw:ip:icmp:data
0  129  raw:ip // these were l2 and decoding was affected

retis-events/src/packet.rs
vlrpl, line 165: isn't safer to turn this into:
```rust
if format.print_ll && !fake {
    self.format_vlan(f, &vlan)?;
}
```

retis-events/src/packet.rs
atenart: I didn't get what "it" refers to.

retis/src/bindings/skb_hook_uapi.rs
atenart: No reason, that was just the version used on my laptop. I can remove that diff, or keep it as the following changes where generated with that version, although nothing else should change here.

retis-events/src/compat/compat.rs
atenart: Ahh thanks for noticing, I forgot to add a new entry :facepalm:

retis/src/process/cli/pcap.rs
atenart: Great point, we also need to encode the packet kind in the interface key (name).

retis-events/src/packet.rs
atenart: If there's a nested VLAN header (tunnel) and we want LL printing, we should display it. The fake Ethernet header doesn't fake any VLAN header.

retis/src/bindings/skb_hook_uapi.rs
vlrpl: yeah, it's really a minor detail, but I would drop this for consistency. gen-bindings is (intentionally) all or nothing and in the doc we require to use the latest stable version.

retis-events/src/packet.rs
vlrpl: I don't recall the details and maybe it's still mandatory for some operations.
My question was about the type "demotion" (RawPacket, but also the field direct access, if possible as there's no longer bytes, to_scapy ...) and whether this may or may not be worthy of being exposed on the python side.

If we can do things like:
raw = bytes(event.packet)

maybe the removal makes things a little cleaner.

PR #655, opened 2026-08-21
Improve runtime collect performances

atenart:
This improves performances at runtime while collecting events (`retis collect`) using a number of techniques:

- Handle events (dequeue from BPF, parsing, formatting) in parallel in userspace.
- Push back as early as possible in the pipeline when reaching the limit of what can be processed => only allow for legitimate bursts.
- Slightly optimize the eBPF side.

More details in the commits (message + comments) and some improvements alongside the reworks.

Handling events in parallel has a consequence: events might not be sorted by default in the output (while running `retis collect` and current `retis print/sort`), because they might be processed out of order. This could be fixed using a buffer before outputting the events, but this is not so simple to implement and there would always be cases were this would not work. Instead events are printed as they arrive during collection and post-processing commands were fixed to reorder events (within a defined time window).

Measuring and comparing performances is not so simple: the fundamental logic was changed in how we buffered and dropped events.

Before:
- Events were dropped in the eBPF side if the ringbuf was full.
- The channel in userspace was unbounded... but we didn't flush it at exit time (no drop reported).

Now:
- Events are dropped as early as possible in the eBPF side: the ringbuf is the main buffer and is only emptied as events are fully processed.
- There is no unbounded buffer (channel) anymore.

Because of this you might see more drops after the changes. But the real metric is how many events were processed, this increases as threads are added.

Feel free to make your own tests and experiments and to tweak some of the values.


atenart:
There are a few issues reported by the runtime CI: test_ovs_detrace_throttle is failing and there is a verifier issue on older kernel. This doesn't impact the direction of the changes and they can be tested and reviewed already, I'll have a look at the issues later.

